PR #12, opened 2020-08-25
Feature/zenoh vendor package

gbiggs:
Replaces the use of `zenoh_ament` with a vendor shim for Zenoh.

The vendor shim:
- clones the Zenoh repository
- checks out a specific commit ([2a93c83afb7b0865fb99d5a6fc4b153badc3b8c7](https://github.com/eclipse-zenoh/zenoh/commit/2a93c83afb7b0865fb99d5a6fc4b153badc3b8c7), so that we don't have to work against a moving target; we can update this as necessary)
- compiles `zenoh-ffi` only
- installs the files we want to the install space
- provides a CMake `find_package()` module to locate the library and headers

The module allows the user to specify a different location for Zenoh rather than the locally-cloned one, if they wish to, for example, use a more recent version or a different branch.

`rmw_zenoh_cpp` has been updated to use the vendor package. Similar changes will need to be made in other repositories that depend on `zenoh_ament`. Packages that do depend on `rmw_zenoh_cpp` rather than on `zenoh` directly will not need any changes made.


codebot (approved):
Thanks!!

PR #17, opened 2020-08-26
update zenoh hash to include latest upstream bugfix

codebot:
Signed-off-by: Morgan Quigley <morgan@osrfoundation.org>


methylDragon (approved):
The bug seems to be squashed!

LGTM

PR #25, opened 2020-09-02
Fix problems reported by cppcheck

gbiggs:
Fixes all problems except the lack of copyright notices, which are fixed by #23.

### DIFF
--- a/rmw_zenoh_cpp/include/rmw_zenoh_cpp/rmw_context_impl.hpp
+++ b/rmw_zenoh_cpp/include/rmw_zenoh_cpp/rmw_context_impl.hpp
@@ -12,4 +12,4 @@ struct rmw_context_impl_t
   bool is_shutdown;
 };
 
-#endif // RMW_ZENOH_CPP__RMW_CONTEXT_IMPL_HPP_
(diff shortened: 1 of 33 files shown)


methylDragon (commented):
The Google C++ style guide (which ROS follows(?)) specifies that multiline function calls should have 4 spaces instead of 2, but cppcheck doesn't seem to be checking for that.

I changed the style for multiline function calls outside of the changes here to match that (though I personally think it's a bit ugly.)

What should we standardise towards?

gbiggs:
We follow the [ROS 2 style guide](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id3). While it's based on the Google style guide, it has a significant number of changes.

For multiline function calls, it [says this](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#open-versus-cuddled-braces):

> When a function call cannot fit on one line, wrap at the open parenthesis (not in between arguments) and start them on the next line with a 2-space indent. Continue with the 2-space indent on subsequent lines for more arguments.

methylDragon (approved):
Ah, ok LGTM.

Watch out for merge conflicts!

PR #30, opened 2020-09-09
Use the rmw API for freeing guard conditions

gbiggs:
The RMW API was being used to allocate guard condition structures but a raw `delete` was being used to free them. This PR corrects that by using the RMW API for freeing them. It also adds some additional checks to avoid bad frees.


methylDragon (approved):
Seems ok.

Builds ok, runs ok.

PR #31, opened 2020-09-09
Don't delete resources allocated by Zenoh

gbiggs:
The `session` value is allocated by Zenoh internally when it creates a new session. Deleting it here creates a race condition that means Zenoh's internal threads may try to access this memory for shutting down after it has been freed. Don't delete it, let Zenoh clean it up itself.


methylDragon (approved):
Oof, ok

PR #33, opened 2020-09-10
Set constant QoS

codebot:
Builds on #32 

Hard-codes QoS to reflect the TCP behavior of Zenoh. Removes the QoS events from the waitset because they aren't relevant anymore, with "static" QoS. Adds some debug/error messages to help future developers hopefully be less-confused than I was :smiley:


methylDragon:
How strange, before it would've sent events to rmw_wait, now it doesn't. Do you know what changed and why?

Also, I really like the debug logs.

PR #38, opened 2020-10-05
Fix exports

esteve:
This PR fixes the exports so that downstream packages have access to zenoh include dirs and libraries.


methylDragon:
This looks ok, but I think we still need the FastRTPS/FastCDR exports since the typesupport package uses FastCDR to do it's serialisation.

Any missed exports with that regard may be masked by the fact that FastCDR is separately sourced in the default ROS underlay.

methylDragon:
I will have to defer to @gbiggs to confirm though, I'm unsure if there have been any major changes to this package.

gbiggs (approved):

PR #42, opened 2020-10-08
Added message repos

esteve:


methylDragon:
I am hesitant to be attaching these message packages along with rmw_zenoh. But I guess it might work because these are common packages...

Unsure. What do you think @gbiggs ?

gbiggs:
I don't think we want these permanently in the `.repos` file. They increase the build time but someone may not want to use them. We don't plan to have `rmw_zenoh` be a separate repository in the long term.

PR #56, opened 2021-09-23
Make rmw_zenoh compatible with zenoh_plugin_dds

kisg:
zenoh_plugin_dds adds a /rt prefix to every ROS2 topic in Zenoh.
This change adds the same prefix to rmw_zenoh. 

This way nodes using rmw_zenoh may communicate with nodes running a DDS based ROS2 backend with the zenoh-bridge-dds in the middle.

We tested this change with the following configuration:
- a node running on Linux with the CycloneDDS backend
- zenoh-bridge-dds
- zenoh-router 
- and an Android app using rmw_zenoh connecting to the router in client mode.

### DIFF
--- a/rmw_zenoh_common_cpp/CMakeLists.txt
+++ b/rmw_zenoh_common_cpp/CMakeLists.txt
@@ -22,6 +22,10 @@ find_package(rosidl_typesupport_zenoh_cpp REQUIRED)
 
 include_directories(include)
 
+if(DEFINED ZENOH_DDS_COMPAT)
+  add_definitions(-DZENOH_DDS_COMPAT=${ZENOH_DDS_COMPAT})
+endif()
+
 add_library(rmw_zenoh_common_cpp
   src/rmw_zenoh_common_event.cpp
   src/rmw_zenoh_common_get_node_info_and_types.cpp
--- a/rmw_zenoh_common_cpp/include/rmw_zenoh_common_cpp/rmw_zenoh_common.h
+++ b/rmw_zenoh_common_cpp/include/rmw_zenoh_common_cpp/rmw_zenoh_common.h
@@ -15,6 +15,8 @@
 #ifndef RMW_ZENOH_COMMON_CPP__RMW_ZENOH_COMMON_H_
 #define RMW_ZENOH_COMMON_CPP__RMW_ZENOH_COMMON_H_
 
+#define ZN_TOPIC_PREFIX "/rt"
+
 #ifdef __cplusplus
 extern "C"
 {
--- a/rmw_zenoh_common_cpp/src/rmw_zenoh_common_publisher.cpp
+++ b/rmw_zenoh_common_cpp/src/rmw_zenoh_common_publisher.cpp
@@ -154,7 +154,8 @@ rmw_zenoh_common_create_publisher(
   // even in the same Zenoh network, because the ID is never transmitted over the wire.
   // Conversely, the ID used in two communicating processes cannot be used to determine if they are
   // using the same topic or not.
-  publisher_data->zn_topic_id_ = zn_declare_resource(session, zn_rname(publisher->topic_name));
+  std::string zn_topic_name = std::string(ZN_TOPIC_PREFIX) + publisher->topic_name;
+  publisher_data->zn_topic_id_ = zn_declare_resource(session, zn_rname(zn_topic_name.c_str()));
 
   // Assign publisher data members
   publisher_data->zn_session_ = session;
--- a/rmw_zenoh_common_cpp/src/rmw_zenoh_common_publisher.cpp
+++ b/rmw_zenoh_common_cpp/src/rmw_zenoh_common_publisher.cpp
@@ -154,7 +154,11 @@ rmw_zenoh_common_create_publisher(
   // even in the same Zenoh network, because the ID is never transmitted over the wire.
   // Conversely, the ID used in two communicating processes cannot be used to determine if they are
   // using the same topic or not.
-  std::string zn_topic_name = std::string(ZN_TOPIC_PREFIX) + publisher->topic_name;
+  #ifdef ZENOH_DDS_COMPAT
+    std::string zn_topic_name = std::string(ZN_TOPIC_PREFIX) + publisher->topic_name; 
+  #else
+    const std::string &zn_topic_name = publisher->topic_name;
+  #endif
   publisher_data->zn_topic_id_ = zn_declare_resource(session, zn_rname(zn_topic_name.c_str()));
 
   // Assign publisher data members
--- a/rmw_zenoh_common_cpp/src/rmw_zenoh_common_subscriber.cpp
+++ b/rmw_zenoh_common_cpp/src/rmw_zenoh_common_subscriber.cpp
@@ -201,7 +201,11 @@ rmw_zenoh_common_create_subscription(
 
     // We initialise subscribers ONCE (otherwise we'll get duplicate messages)
     // The topic name will be the same for any duplicate subscribers, so it is ok
-    std::string zn_topic_name = std::string(ZN_TOPIC_PREFIX) + subscription->topic_name;
+    #ifdef ZENOH_DDS_COMPAT
+      std::string zn_topic_name = std::string(ZN_TOPIC_PREFIX) + subscription->topic_name;
+    #else
+      const std::string &zn_topic_name = subscription->topic_name;
+    #endif
     subscription_data->zn_subscriber_ = zn_declare_subscriber(
       subscription_data->zn_session_,
       zn_rname(zn_topic_name.c_str()),
--- a/rmw_zenoh_common_cpp/src/rmw_zenoh_common_subscriber.cpp
+++ b/rmw_zenoh_common_cpp/src/rmw_zenoh_common_subscriber.cpp
@@ -201,9 +201,10 @@ rmw_zenoh_common_create_subscription(
 
     // We initialise subscribers ONCE (otherwise we'll get duplicate messages)
     // The topic name will be the same for any duplicate subscribers, so it is ok
+    std::string zn_topic_name = std::string(ZN_TOPIC_PREFIX) + subscription->topic_name;
     subscription_data->zn_subscriber_ = zn_declare_subscriber(
       subscription_data->zn_session_,
-      zn_rname(subscription->topic_name),
+      zn_rname(zn_topic_name.c_str()),
       zn_subinfo_default(),  // NOTE(CH3): Default for now
       subscription_data->zn_sub_callback,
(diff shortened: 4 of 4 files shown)


JEnoch:
Actually that's not zenoh_plugin_dds that adds `/rt` prefix to ROS2 topics, that's ROS2 itself when mapping ROS topic names to DDS topic names, as specified here:
 - https://design.ros2.org/articles/topic_and_service_names.html#mapping-of-ros-2-topic-and-service-names-to-dds-concepts

See also some examples of mapping here:
 - https://github.com/eclipse-zenoh/zenoh-plugin-dds#mapping-ros2-names-to-zenoh-resources

As you can see, `/rt` is not the only prefix. If you also want to have ROS services, actions and parameters to be interoperable with a ROS2+DDS system via the zenoh_plugin_dds, the other prefixes will also have to be addressed.

But in a "pure" ROS+zenoh system (without DDS), the services, actions and parameters would benefit to use zenoh's `get`+`queryable` rather than a pub/sub mechanism based on several topics for requests/replies.

kisg:
We are actually working on another PR to change the service and parameter implementation in rmw_zenoh to work with DDS as well. We don't use actions, so that is not a high priority for us right now.

If you say that from a Zenoh perspective the get + queryable would be a better solution, wouldn't it be better to make this conversion in the zenoh_plugin_dds?

Based on our tests, rmw_zenoh provides a very nice way to integrate Android or other external mobile devices with a ROS2 based ecosystem without the added complexity of a DDS backend.

JEnoch:
> We are actually working on another PR to change the service and parameter implementation in rmw_zenoh to work with DDS as well. We don't use actions, so that is not a high priority for us right now.

That's great! Thanks!

> If you say that from a Zenoh perspective the get + queryable would be a better solution, wouldn't it be better to make this conversion in the zenoh_plugin_dds?

The main goal of this plugin is to be generic for any DDS application, not only ROS2. And DDS doesn't natively support the request/reply pattern.

But my question was more on the purpose of rmw_zenoh:
 - do we want it to be interoperable with DDS-based rmw nodes ?
 - do we want it to leverage the zenoh specificities, but loosing interop with DDS ? Or developing a specific ROS2 bridge for interop ?
 - do we want both to be possible ?

That's open questions to me. Maybe @kydos has some opinions on this.

> Based on our tests, rmw_zenoh provides a very nice way to integrate Android or other external mobile devices with a ROS2 based ecosystem without the added complexity of a DDS backend.

That's a cool use case! Moreover if you do this over the internet with one or more zenoh routers between Android and the robot!  Can't wait to see a demo 😃

gbiggs:
I don't think that it is a good idea to force `rmw_zenoh` to match what `zenoh_plugin_dds` spits out when it's being used with a ROS 2 system. That would be unnecessarily constraining `rmw_zenoh` to match how ROS 2 makes use of DDS, and effectively exposing ROS 2 internals. I think it is better to let `rmw_zenoh` be its own thing and provide what ROS 2 needs in the best way for Zenoh.

If someone wants to have a zenoh-powered ROS 2 node and a DDS-powered ROS 2 node talk to each other, then writing a bridge specifically for that would be better in my opinion.

kydos:
Here my perspective:

> But my question was more on the purpose of rmw_zenoh:
> 
> * do we want it to be interoperable with DDS-based rmw nodes ?

Ideally yes and through the **zenoh-plugin-dds**.

> * do we want it to leverage the zenoh specificities, but loosing interop with DDS ? Or developing a specific ROS2 bridge for interop ?

We can mitigate some of the mismatch via the bridge. For instance queries for DDS-based robots could be replied by the bridge... In any case short term I'd say that the more interesting exercise is to have an RMW that fully exploits zenoh to gain scalability, performance and to ease some of the use cases we know are problematic with DDS-based RMW.

> * do we want both to be possible ?

As I said above, for me the first goal should be to leverage zenoh at it maximum. Because (1) changing RMW is as simple as changing  an environment variable and (2) for those cases in which this is not possible we can leverage the bridge.

kisg:
Thank you @kydos @JEnoch and @gbiggs for sharing your perspective.

> @JEnoch , here my perspective:
> 
> > But my question was more on the purpose of rmw_zenoh:
> > 
> > * do we want it to be interoperable with DDS-based rmw nodes ?
> 
> Ideally yes and through the **zenoh-plugin-dds**.
> 
So you would rather accept patches that change how the zenoh-plugin-dds maps to Zenoh?

> > * do we want it to leverage the zenoh specificities, but loosing interop with DDS ? Or developing a specific ROS2 bridge for interop ?
> 
> We can mitigate some of the mismatch via the bridge. For instance queries for DDS-based robots could be replied by the bridge... In any case short term I'd say that the more interesting exercise is to have an RMW that fully exploits zenoh to gain scalability, performance and to ease some of the use cases we know are problematic with DDS-based RMW.
> 
> > * do we want both to be possible ?
> 
> As I said above, for me the first goal should be to leverage zenoh at it maximum. Because (1) changing RMW is as simple as changing an environment variable and (2) for those cases in which this is not possible we can leverage the bridge.

For us replacing CycloneDDS with rmw_zenoh in its current form is not a viable option:
- rmw_zenoh is still experimental (I know it is an easy fix in the README :) )
- Even if it is technically easy to do the switch, a substantial amount of testing (and most probably fixing) is required to make sure it actually works for every use case in the whole system.
- Choosing an RMW implementation is a critical decision. We spent literally months trying to fix all the problems with FastRTPs, before CycloneDDS was available, and practically solved all our issues.

On the other hand, we have a requirement to connect over a routed network to our system from mobile devices (Android, iOS). One option is to just create a custom REST API and call it a day. But for the long term, exposing a subset of the ROS2 APIs we already have is a much better solution. Using the DDS bridge + a full blown DDS implementation on mobile devices is essentially out of the question. But here, because the external connection is not critical, it is much easier to argue the case for Zenoh and rmw_zenoh. 

We already talked about this a couple months back: I think the current rmw_zenoh implementation is a good interim solution, but I think the right way forward would be reimplementing it in Rust. 

That new implementation could be very well optimized (without even going through the zenoh-c layer or using CDR for serialization) for maximum performance; at the same time it would leverage the safety features of the Rust language. Even better: that implementation could be designed in a way that supports different modes of communication: "compatibility" for using with the DDS bridge and optimized when rmw_zenoh is used exclusively.

This is the reason why I think it would be more important for the current rmw_zenoh implementation to be compatible with the DDS bridge than aim for maximum performance. 

What do you think?

kydos:
> > As I said above, for me the first goal should be to leverage zenoh at it maximum. Because (1) changing RMW is as simple as changing an environment variable and (2) for those cases in which this is not possible we can leverage the bridge.
> 
> For us replacing CycloneDDS with rmw_zenoh in its current form is not a viable option:
> 
> * rmw_zenoh is still experimental (I know it is an easy fix in the README :) )

:-)

> * Even if it is technically easy to do the switch, a substantial amount of testing (and most probably fixing) is required to make sure it actually works for every use case in the whole system.
> * Choosing an RMW implementation is a critical decision. We spent literally months trying to fix all the problems with FastRTPs, before CycloneDDS was available, and practically solved all our issues.

I am glad that Cyclone DDS solved all the issues.

> On the other hand, we have a requirement to connect over a routed network to our system from mobile devices (Android, iOS). One option is to just create a custom REST API and call it a day. But for the long term, exposing a subset of the ROS2 APIs we already have is a much better solution. Using the DDS bridge + a full blown DDS implementation on mobile devices is essentially out of the question. But here, because the external connection is not critical, it is much easier to argue the case for Zenoh and rmw_zenoh.

Makes complete sense and based in our experience this is the most common use case we see in the short term. 

> 
> We already talked about this a couple months back: I think the current rmw_zenoh implementation is a good interim solution, but I think the right way forward would be reimplementing it in Rust.

> 
> That new implementation could be very well optimized (without even going through the zenoh-c layer or using CDR for serialization) for maximum performance; at the same time it would leverage the safety features of the Rust language. Even better: that implementation could be designed in a way that supports different modes of communication: "compatibility" for using with the DDS bridge and optimized when rmw_zenoh is used exclusively.

+1
 
> This is the reason why I think it would be more important for the current rmw_zenoh implementation to be compatible with the DDS bridge than aim for maximum performance.
> 
> What do you think?

Makes sense and as mentioned above is in line with the short term use cases we are seeing.

JEnoch:
So I think we all agree 😄 
To summarize:
 
1. The main purpose of `rmw_zenoh` (whether in Rust, C++ or Python) is to leverage as much as possible the benefits of zenoh. This should be the "default mode".
2. Optionally (via a Rust feature or any compile-time option) `rmw_zenoh` could switch to a "DDS-compatible" mode where it "mimics" a ROS2 DDS-based system, but using zenoh pub/sub. This would make it interoperable with a ROS2 DDS-based system via the `zenoh-plugin-dds` in its current version.
3. The interop between a ROS2 zenoh-based system and a ROS2 DDS-based system could be achieved with a an extension of the `zenoh-plugin-dds` that would bridge the usage of zenoh by `rmw_zenoh` (default mode) with a ROS2 DDS-based system.

@kisg if you have no objection, can you update your PR to make the addition of `rt/` prefix optional with a compile-time config ?
Out of curiosity, why didn't you consider to directly develop zenoh application (without RMW layer) for Android, similarly than explained in [this blog](https://zenoh.io/blog/2021-04-28-ros2-integration/) ?

kisg:
> @kisg if you have no objection, can you update your PR to make the addition of `rt/` prefix optional with a compile-time config ?
Sure, we will update the PR, thank you.

> Out of curiosity, why didn't you consider to directly develop zenoh application (without RMW layer) for Android, similarly than explained in [this blog](https://zenoh.io/blog/2021-04-28-ros2-integration/) ?

The reason is that we already have an API implemented on top of ROS2, and our goal is to use that API on every platform regardless of the backend implementation. This API is reusing ROS2 types defined in IDL, but it is actually a bit more constrained than the ROS2 API, e.g. it requires the user to always use fixed size messages allocated by the API to work well with ROS2's LoanedMessage abstraction to support shared memory based backends. The API also abstracts other ROS2 specifics, e.g. the QoS settings and the exact topic where the messages would be sent.
We see it as a big benefit that we can replace the backend of our API implementation by switching to a different ROS2 RMW. (And that we don't have to implement these backends ourselves... :) )

That said, I am very much interested in using Zenoh as the ROS2 backend for the whole system in the future.

JEnoch (approved):
Looks good to me. Thanks for this update!

JEnoch (approved):

JEnoch:
I'm not sure why the build for this PR is failing with this message:
`Unknown CMake command "rosidl_get_typesupport_target"`

Anyway, the code changes seems unrelated, and the build in main branch succeeds... Let's merge it.

PR #61, opened 2023-11-09
Verify a zenoh router is available for the session

francocipollone:
### Summary

Verify that there is a connection between the Zenoh session being initialized at rmw_init method and a Zenoh router.
In case there is no connected router then return with error.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
+++ b/rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
@@ -1,0 +1,88 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "zenoh_router_check.hpp"
+
+#include <rcutils/env.h>
+#include <rcutils/logging_macros.h>
+
+#include <string>
+#include <sstream>
+#include <iomanip>
+
+namespace
+{
+
+// Convert a Zenoh Id to a string
+// Zenoh IDs are LSB-first 128bit unsigned and non-zero integers in hexadecimal lowercase.
+// @param pid Zenoh Id to convert
+std::string ZidToStr(z_id_t pid)
+{
+  std::stringstream ss;
+  int len = 0;
+  for (int i = 0; i < 16; i++) {
+    if (pid.id[i]) {
+      len = i + 1;
+    }
+  }
+  if (!len) {
+    ss << "";
+  } else {
+    for (int i = len - 1; i >= 0; --i) {
+      ss << std::hex << std::setfill('0') << std::setw(2) << static_cast<int>(pid.id[i]);
+    }
+  }
+  return ss.str();
+}
+
+}  // namespace
+
+rmw_ret_t zenoh_router_check(z_session_t session)
+{
+  // Initialize context for callback
+  void * context = malloc(sizeof(int));
+  *(static_cast<int *>(context)) = 0;
+
+  // Define callback
+  auto callback = [](const struct z_id_t * id, void * ctx) {
+      const std::string id_str = ZidToStr(*id);
+      RCUTILS_LOG_INFO_NAMED(
+        "ZenouRouterCheck",
+        "A Zenoh router connected to the session with id '%s'", id_str.c_str());
+      // Note: Callback is guaranteed to never be called
+      // concurrently according to z_info_routers_zid docstring
+      (*(static_cast<int *>(ctx)))++;
+    };
+
+  z_owned_closure_zid_t router_callback = z_closure(callback, nullptr /* drop */, context);
+  z_info_routers_zid(session, z_move(router_callback));
+
+  rmw_ret_t ret;
+  if (*(static_cast<int *>(context)) == 0) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "ZenouRouterCheck",
+      "No Zenoh router connected to the session");
+    ret = RMW_RET_ERROR;
+  } else {
+    RCUTILS_LOG_INFO_NAMED(
+      "ZenouRouterCheck",
+      "There are %d Zenoh routers connected to the session", *(static_cast<int *>(context)));
+
+    ret = RMW_RET_OK;
+  }
+
+  // Not using the drop function from the closure as we want to keep the context for logging.
+  free(context);
+  return ret;
+}
--- a/rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
+++ b/rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
@@ -17,9 +17,9 @@
 #include <rcutils/env.h>
 #include <rcutils/logging_macros.h>
 
-#include <string>
-#include <sstream>
 #include <iomanip>
+#include <sstream>
+#include <string>
 
 namespace
 {
@@ -27,13 +27,13 @@ namespace
 // Convert a Zenoh Id to a string
 // Zenoh IDs are LSB-first 128bit unsigned and non-zero integers in hexadecimal lowercase.
 // @param pid Zenoh Id to convert
-std::string ZidToStr(z_id_t pid)
+std::string zid_to_str(const z_id_t & pid)
 {
   std::stringstream ss;
   int len = 0;
-  for (int i = 0; i < 16; i++) {
+  for (size_t i = 0; i < sizeof(pid.id); ++i) {
     if (pid.id[i]) {
-      len = i + 1;
+      len = static_cast<int>(i) + 1;
     }
   }
   if (!len) {
@@ -51,14 +51,13 @@ std::string ZidToStr(z_id_t pid)
 rmw_ret_t zenoh_router_check(z_session_t session)
 {
   // Initialize context for callback
-  void * context = malloc(sizeof(int));
-  *(static_cast<int *>(context)) = 0;
+  int context = 0;
 
   // Define callback
   auto callback = [](const struct z_id_t * id, void * ctx) {
-      const std::string id_str = ZidToStr(*id);
+      const std::string id_str = zid_to_str(*id);
       RCUTILS_LOG_INFO_NAMED(
-        "ZenouRouterCheck",
+        "ZenohRouterCheck",
         "A Zenoh router connected to the session with id '%s'", id_str.c_str());
       // Note: Callback is guaranteed to never be called
       // concurrently according to z_info_routers_zid docstring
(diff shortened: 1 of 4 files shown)


rmw_zenoh_cpp/src/detail/zenoh_router_check.hpp
```diff
@@ -0,0 +1,27 @@
...
+#define DETAIL__ZENOH_ROUTER_CHECK_HPP_
+
+#include <zenoh.h>
+
+#include "rmw/ret_types.h"
+
+/// Check if a Zenoh router is connected to the session.
+/// @param session Zenoh session to check.
+/// @return RMW_RET_OK if a Zenoh router is connected to the session.
+rmw_ret_t zenoh_router_check(z_session_t session);
```
Yadunund, line 25: If we're adding a single function that is also not going to be reused in multiple places, do you think we need a new header file? Can we just define the helper function within `rmw_init.cpp` or even within `rmw_init()`?

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,88 @@
...
+      // Note: Callback is guaranteed to never be called
+      // concurrently according to z_info_routers_zid docstring
+      (*(static_cast<int *>(ctx)))++;
+    };
+
+  z_owned_closure_zid_t router_callback = z_closure(callback, nullptr /* drop */, context);
+  z_info_routers_zid(session, z_move(router_callback));
+
+  rmw_ret_t ret;
+  if (*(static_cast<int *>(context)) == 0) {
```
Yadunund: I'm trying to understand how the `z_info_routers_zid()` API works.
```
/**
 * Fetches the Zenoh IDs of all connected routers.
 *
 * `callback` will be called once for each ID, is guaranteed to never be called concurrently,
 * and is guaranteed to be dropped before this function exits.
 *
 * Retuns 0 on success, negative values on failure.
 */
ZENOHC_API
int8_t z_info_routers_zid(struct z_session_t session,
                          struct z_owned_closure_zid_t *callback);
 ```
 
 If it returns an `int8_t`, is it a blocking function or will the function execute callbacks asynchronously? If its the former, we could just check if the return value is 0 and if so return `RMW_RET_OK`? Unless we plan to parse the zids we receive in the callbacks to check if the zid matches that of the zenoh router we run (and not any other zenoh router)?

File: rmw_zenoh_cpp/src/detail/zenoh_router_check.hpp (hunk 1 above)
Comment (francocipollone) [reply rmw_zenoh_cpp/src/detail/zenoh_router_check.hpp:25]: Yeah it is true we are not using this more than once. The only reason to have a separate file with this method is to make the `rmw_init.cpp` file a bit more clean and if at some we want to add tests this will be useful.
(Without mention that if we want to launch the zenoh router from here probably another method might be implementeed)

Instead of having just a sole method in this header file we can merge this file with the `zenoh_config.hpp` header and create a `zenoh_utils.hpp` file.

If you still recommend having it all in the same `rmw_init.cpp` file then I am ok with moving this code into the `rmw_init.cpp`. ptal @Yadunund

File: rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp (hunk 2 above)
Comment (francocipollone) [reply rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp]: > is it a blocking function or will the function execute callbacks asynchronously? 

It is a blocking function.

> we could just check if the return value is 0 and if so return RMW_RET_OK?

It "returns 0 in success" but that doesn't convey whether there are or not available routers. It would return 0 even if the zenoh router is missing. This only states whether the execution was correct. (Just realized that we should return error in case there is a failure, adding that...)

Comment (francocipollone) [reply rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp]: > Unless we plan to parse the zids we receive in the callbacks to check if the zid matches that of the zenoh router we run (and not any other zenoh router)?

For the moment we need the callback to check that at least a router is connected to the session. I added the parsing of the id mainly for logging purposes. We could at some point verify the id to correctly assert that we are connected to the zenoh router we want. However it is a bit early for this I think, for the moment we don't have environments with more than one zenoh router.

File: rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+
+#include "zenoh_router_check.hpp"
+
+#include <rcutils/env.h>
+#include <rcutils/logging_macros.h>
+
+#include <string>
+#include <sstream>
+#include <iomanip>
```
clalancette:
```suggestion
#include <iomanip>
#include <sstream>
#include <string>
```

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+
+// Convert a Zenoh Id to a string
+// Zenoh IDs are LSB-first 128bit unsigned and non-zero integers in hexadecimal lowercase.
+// @param pid Zenoh Id to convert
+std::string ZidToStr(z_id_t pid)
+{
+  std::stringstream ss;
+  int len = 0;
+  for (int i = 0; i < 16; i++) {
```
clalancette:
```suggestion
  for (size_t i = 0; i < sizeof(pid.id); ++i) {
```

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <sstream>
+#include <iomanip>
+
+namespace
+{
+
+// Convert a Zenoh Id to a string
+// Zenoh IDs are LSB-first 128bit unsigned and non-zero integers in hexadecimal lowercase.
+// @param pid Zenoh Id to convert
+std::string ZidToStr(z_id_t pid)
```
clalancette:
```suggestion
std::string zid_to_str(const z_id_t & pid)
```

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+  return ss.str();
+}
+
+}  // namespace
+
+rmw_ret_t zenoh_router_check(z_session_t session)
+{
+  // Initialize context for callback
+  void * context = malloc(sizeof(int));
+  *(static_cast<int *>(context)) = 0;
```
clalancette: Hm, does this need to be heap-allocated memory?  Could this just be a stack-allocated `int` that we capture in the callback?

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+      // Note: Callback is guaranteed to never be called
+      // concurrently according to z_info_routers_zid docstring
+      (*(static_cast<int *>(ctx)))++;
+    };
+
+  rmw_ret_t ret;
+  z_owned_closure_zid_t router_callback = z_closure(callback, nullptr /* drop */, context);
+  if (z_info_routers_zid(session, z_move(router_callback))) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "ZenouRouterCheck",
```
clalancette:
```suggestion
      "ZenohRouterCheck",
```

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+      "Failed to evaluate if Zenoh routers are connected to the session");
+    ret = RMW_RET_ERROR;
+  } else {
+    if (*(static_cast<int *>(context)) == 0) {
+      RCUTILS_LOG_ERROR_NAMED(
+        "ZenouRouterCheck",
```
clalancette:
```suggestion
        "ZenohRouterCheck",
```

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+        "No Zenoh router connected to the session");
+      ret = RMW_RET_ERROR;
+    } else {
+      RCUTILS_LOG_INFO_NAMED(
+        "ZenouRouterCheck",
```
clalancette:
```suggestion
        "ZenohRouterCheck",
```

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -0,0 +1,92 @@
...
+        "A Zenoh router connected to the session with id '%s'", id_str.c_str());
+      // Note: Callback is guaranteed to never be called
+      // concurrently according to z_info_routers_zid docstring
+      (*(static_cast<int *>(ctx)))++;
+    };
+
+  rmw_ret_t ret;
+  z_owned_closure_zid_t router_callback = z_closure(callback, nullptr /* drop */, context);
+  if (z_info_routers_zid(session, z_move(router_callback))) {
```
clalancette, line 69: How long does this wait around for the callback to be called?

clalancette (changes requested):

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Thanks for the suggestion. Yeah it wasn't necessary.

Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
francocipollone: From what I understood from checking out the code (no rust savvy ),

https://github.com/eclipse-zenoh/zenoh-c/blob/e1f2c1a3cdfd9534a44aaeb4aa8e5cf40200a843/src/info.rs#L74-L89

 it is simply getting the router's info from the session and iterating over them while applying the "callback" which is simply like a predicate.
So I don't expect to be any wait.

clalancette: Ah, got it.  Thanks for the explanation.

rmw_zenoh_cpp/src/detail/zenoh_router_check.hpp
clalancette: It's a little overkill to have this header for just one function, but I'm fine with this for now.  If we never end up adding anything here, we can always combine this into `rmw_init` later on.

clalancette (approved):

PR #65, opened 2023-11-15
Add storage plugin config to router config

Yadunund:
Temporarily rely on storage plugin to store mock liveliness updates for late joining participants.


Yadunund:
Actually let's hold off on this since liveliness bindings are available in `zenoh-c` now.

Yadunund:
No longer required following #67

PR #66, opened 2023-11-15
Finish GraphCache implementation for Pub/Sub

Yadunund:
This PR:
- Refactors how we store and update the `GraphCache` within each session. Specifically we move away from a `YAML::Node` to an `unordered_map` (although implementing a conversion from our custom map to `YAML` may be helpful for logging or transmitting the graph as a serialized `YAML` over the network).
- Makes the switch to rely on liveliness tokens to broadcast changes to the ROS graph. The previous implementation of relying on put/del/get convey the changes is retained but commented out as it could be reused to support discovery with `zenoh-pico`.
- Implements some basic topic/node introspection functions in `rmw` (the rest will be done in a separate PR).

> Note: Without the router, nodes will not be able to discover each other since multicast discovery is disabled by default in the node's session config. Instead, nodes will receive discovery information about other peers via the router's gossip function.


rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -44,8 +45,17 @@ static std::string generate_base_token(
   const std::string & name)
 {
   std::stringstream token_ss;
-  // TODO(Yadunund): Empty namespace will contain /. Fix non-empty namespace.
-  token_ss << "@ros2_lv/" << domain_id << "/" << entity << namespace_ << name;
+  token_ss << "@ros2_lv/" << domain_id << "/" << entity << namespace_;
```
francocipollone, line 62: The use of @ is to indicate that it is for internal use? (I know that we are using the same approach to zenoh-dds plugin (according to slides) however I don't remember what's this for) Adding a comment with that would be nice for future reviews.

francocipollone: As it is being used in more than one method we could create a static const variable with that definition and replace all the ocurrences

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -44,8 +45,17 @@ static std::string generate_base_token(
...
+  // An empty namespace from rcl will contain "/"" but zenoh does not allow keys with "//".
```
francocipollone: nit: extra "

```suggestion
  // An empty namespace from rcl will contain "/" but zenoh does not allow keys with "//".
```

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -44,8 +45,17 @@ static std::string generate_base_token(
...
+  // Hence we add an "_" to denote an empty namespace such that splitting the key
+  // will always result in 5 parts.
+  if (namespace_ == "/") {
+    token_ss << "_/";
+  } else {
+    token_ss << "/";
+  }
```
francocipollone, line 70: nit: it can be repaced by one line:

```suggestion
  token_ss << (namespace_ == "/") ? "_/" : "/";
```

francocipollone: nit: I think that a docstring for the method with the expected token format will be neat

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -69,6 +79,22 @@ std::string GenerateToken::publisher(
 {
   std::string token = generate_base_token("MP", domain_id, node_namespace, node_name);
   token += topic + "/" + type + "/" + qos;
+  printf("GenerateToken::Publisher: %s\n", token.c_str());
```
francocipollone: nit: printf left behind. Perhaps it could be evolved to a debug message or just be removed

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -69,6 +79,22 @@ std::string GenerateToken::publisher(
...
+  size_t domain_id,
+  const std::string & node_namespace,
+  const std::string & node_name,
+  const std::string & topic,
+  const std::string & type,
+  const std::string & qos)
+{
+  std::string token = generate_base_token("MS", domain_id, node_namespace, node_name);
+  token += topic + "/" + type + "/" + qos;
+  printf("GenerateToken::Subscription: %s\n", token.c_str());
```
francocipollone: nit: same comment here

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -270,33 +191,185 @@ static std::vector<std::string> split_keyexpr(const std::string & keyexpr)
...
+
   // Get the entity, ie N, MP, MS, SS, SC.
-  const std::string & entity = parts[2];
+  std::string & entity = parts[2];
+  if (entity != "NN" &&
+    entity != "MP" &&
+    entity != "MS" &&
+    entity != "SS" &&
+    entity != "SC")
+  {
```
francocipollone: We should document the entities somewhere in the code:
```
NN = node?
MP: = Publisher
MS: = Subscriber
SS: = Service Server
SC: = Service Client
```

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -270,33 +191,185 @@ static std::vector<std::string> split_keyexpr(const std::string & keyexpr)
...
+    data.topic = "/" + std::move(parts[5]);
+    data.type = std::move(parts[6]);
+    data.qos = std::move(parts[7]);
+
+    if (entity == "MP") {
+      node.pubs.push_back(std::move(data));
+    } else if (entity == "MS") {
+      node.subs.push_back(std::move(data));
+    } else {
+      // TODO.
```
francocipollone:
```suggestion
      RCUTILS_LOG_ERROR_NAMED(
        "rmw_zenoh_cpp",
        "Entity type is not supported yet");
```

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -270,33 +191,185 @@ static std::vector<std::string> split_keyexpr(const std::string & keyexpr)
...
-    RCUTILS_LOG_WARN_NAMED("rmw_zenoh_cpp", "Adding node %s to the graph.", parts.back().c_str());
-    const bool has_namespace = entity.size() == 5 ? true : false;
-    graph_[parts.back()] = YAML::Node();
-    // TODO(Yadunund): Implement enclave support.
-    graph_[parts.back()]["enclave"] = "";
-    graph_[parts.back()]["namespace"] = has_namespace ? parts.at(4) : "/";
+    auto ns_it = graph_.find(node->ns);
+    if (ns_it == graph_.end()) {
+      // New namespace.
+      std::string ns = node->ns;
```
francocipollone: nit: only used below, this extra line can be avoided

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -270,33 +191,185 @@ static std::vector<std::string> split_keyexpr(const std::string & keyexpr)
...
+      std::unordered_map<std::string, GraphNodePtr> map = {
+        {node->name, node}
+      };
+      graph_.insert(std::make_pair(std::move(ns), std::move(map)));
+    } else {
+      auto insertion = ns_it->second.insert(std::make_pair(node->name, node));
+      if (!insertion.second) {
```
francocipollone, line 299: We can rely on `std::find_if` from `<algorithm>` here as we are in cpp 17

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -312,29 +385,121 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+        auto erase_it = found_node->pubs.begin();
+        for (; erase_it != found_node->pubs.end(); ++erase_it) {
+          const auto & pub = *erase_it;
+          if (pub.topic == node->pubs.at(0).topic &&
+            pub.type == node->pubs.at(0).type &&
+            pub.qos == node->pubs.at(0).qos)
+          {
+            break;
+          }
+        }
```
francocipollone, line 437: nit: use `find_if` here and there 
```suggestion
        auto erase_it = std::find_if(found_node->pubs.begin(), found_node->pubs.end(), [&node](const auto &pub) {
            return pub.topic == node->pubs.at(0).topic &&
                   pub.type == node->pubs.at(0).type &&
                   pub.qos == node->pubs.at(0).qos;
        });
```
Then proceed as you do:

```cpp
        if (erase_it != found_node->pubs.end()) {
            // Found a matching element, erase it
            found_node->pubs.erase(erase_it);
            //...
            //...
            //...
        }
```

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -370,7 +535,14 @@ rmw_ret_t GraphCache::get_node_names(
   RCUTILS_CHECK_ALLOCATOR_WITH_MSG(
     allocator, "get_node_names allocator is not valid", return RMW_RET_INVALID_ARGUMENT);
 
-  size_t nodes_number = graph_.size();
+  size_t nodes_number = 0;
+  for (auto it = graph_.begin(); it != graph_.end(); ++it) {
+    nodes_number += it->second.size();
+  }
+  // TODO(Yadunund): Delete.
```
francocipollone: Is this TODO for the logging?

rmw_zenoh_cpp/src/rmw_get_topic_names_and_types.cpp
```diff
@@ -31,63 +33,69 @@ rmw_get_topic_names_and_types(
...
-  }
-  auto cleanup_names_and_types = rcpputils::make_scope_exit(
-    [topic_names_and_types] {
-      rmw_ret_t fail_ret = rmw_names_and_types_fini(topic_names_and_types);
-      if (fail_ret != RMW_RET_OK) {
-        RMW_SAFE_FWRITE_TO_STDERR("failed to cleanup names and types during error handling");
-      }
-    });
+  // rmw_ret_t ret = rmw_names_and_types_init(topic_names_and_types, 1, allocator);
+  // if (ret != RMW_RET_OK) {
```
francocipollone: nit: Missing removing commented code

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -423,31 +595,30 @@ rmw_ret_t GraphCache::get_node_names(
...
+        return RMW_RET_BAD_ALLOC;
+      }
+      if (enclaves) {
+        enclaves->data[j] = rcutils_strdup(
+          node->enclave.c_str(), *allocator);
+        if (!enclaves->data[j]) {
+          return RMW_RET_BAD_ALLOC;
+        }
+      }
+      ++j;
```
francocipollone, line 623: General comment for this method. 
I wasn't able to retrieve the list of nodes via `ros2 node list`. I verified and we de call this method from `rmw_get_node_names`. I couldn't debug yet to see what is going on.

francocipollone (commented):
Great Job @Yadunund  :rocket: 
I left a bunch of not-game-changing comments, ptal

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Surprised to hear this. Was the router running? Without the router the node that gets spun via `ros2 node list` will not get gossip information about other nodes.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Just a custom admin space for liveliness tokens https://zenoh.io/docs/manual/abstractions/#admin-space

clalancette:
I ended up rebasing this on the latest, as well as fixing some bugs that I found.  I also addressed most of @francocipollone 's comments, with the exception of a switch to `find_if`.  I'll take a closer look at that later.

However, with all of that done, `ros2 node list` and `ros2 topic list -t` both work entirely correctly for me.  There is still an open TODO in there where a topic with multiple types on it won't be shown properly.  To fix that, we'll need to revamp the way some of the data is put together.  That's a project for next week.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
clalancette: While this could be a one-liner, I actually think it is harder to read that way.  So I left it as-is.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
clalancette: I documented this in the docblock for `generate_base_token`.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
clalancette: This ended up working for me, as long as the router is running.  @francocipollone give it another shot with the latest, see if you can make it work.

francocipollone:
> I ended up rebasing this on the latest, as well as fixing some bugs that I found. I also addressed most of @francocipollone 's comments, with the exception of a switch to `find_if`. I'll take a closer look at that later.
> 
> However, with all of that done, `ros2 node list` and `ros2 topic list -t` both work entirely correctly for me. There is still an open TODO in there where a topic with multiple types on it won't be shown properly. To fix that, we'll need to revamp the way some of the data is put together. That's a project for next week.

Confirmed, it is working correctly. It seems it was my bad: I had some issues with the `daemon`. Basically, it was already running (I was working on another branch) and then when I checkouted this branch and built I hadn't stopped the daemon.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
francocipollone: Confirmed --> https://github.com/ros2/rmw_zenoh/pull/66#issuecomment-1817112633

Yadunund:
Thanks for cleaning this up Chris! I'll work on massaging the cache data structures to support multiple types on the same topic in a separate PR to keep the diffs minimal. If both of you are okay with the code here, should we merge this in first?

clalancette:
> Thanks for cleaning this up Chris! I'll work on massaging the cache data structures to support multiple types on the same topic in a separate PR to keep the diffs minimal. If both of you are okay with the code here, should we merge this in first?

Yeah, I think we should do that.  The open issues here are 1) to switch to find_if as suggested by Franco, and 2) fix multiple types per topic.  As long as we fix both of those in a follow-up, I'm good with merging this.

clalancette (approved):

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: `std::find_if()` will have O(n) complexity while the insertion approach should have `O(log n)`. The idea is that insertion will fail if the entry exists and will give us an iterator to the element that already exists which can be updated.

Yadunund: Rust has a nicer way to do this with HashMap's`entry().or_insert()` [pattern](https://stackoverflow.com/questions/30414424/how-can-i-update-a-value-in-a-mutable-hashmap).

PR #68, opened 2023-11-20
Add docker support for Iron

CatInTheRain:
Hi this is a working Dockerfile. I tried to build this 2 dockerfiles and run them in two different machines but they cannot interact with each other with zenoh (they are connected in the same WiFi). Communication works in the same machine.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,0 +1,52 @@
+FROM osrf/ros:iron-desktop
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+
+#  install dependencies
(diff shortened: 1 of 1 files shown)


clalancette:
Thanks for this, but we aren't yet ready to add in Dockerfiles for this.  We don't know which ROS releases we are going to support yet, and we don't know exactly what our deployment profile is going to be.  So I think it is too early to put this in.  For that reason, I'm going to mark this as a "Draft" for now, and we can revisit it later when we are closer to knowing what this all needs to look like.

clalancette:
Since Iron is now EOL, I'm going to close this one.  Thanks for thinking about it, it is much appreciated.

PR #70, opened 2023-11-24
Refactor GraphCache to support multiple types for the same topic.

Yadunund:
This PR
- Updates the `GraphCache` implementation to support multiple types over the same topic name.
- Previously we were not keeping track of multiple pub/subs in the same node that publish over the same topic and let alone with different topic types. This PR addresses both these problems as well.
- It also reduces code duplication when adding/removing pub/subs by relying on lambdas that are reused for both pub/subs.
- Adds data structures and methods (see `liveliness_utils.hpp`) to manage token generation and conversion to C++ types. No longer have any magic string literal comparisons or custom key generation for unordered_maps.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
 #include <memory>
 #include <mutex>
 #include <optional>
@@ -425,16 +426,13 @@ void GraphCache::parse_del(const std::string & keyexpr)
         // Here we iterate throught the list of publishers and remove the one
         // with matching name, type and qos.
         // TODO(Yadunund): This can be more optimal than O(n) with some caching.
-        auto erase_it = found_node->pubs.begin();
-        for (; erase_it != found_node->pubs.end(); ++erase_it) {
-          const auto & pub = *erase_it;
-          if (pub.topic == node->pubs.at(0).topic &&
+        auto erase_it = std::find_if(
+          found_node->pubs.begin(), found_node->pubs.end(),
+          [&node](const auto & pub) {
+            return pub.topic == node->pubs.at(0).topic &&
             pub.type == node->pubs.at(0).type &&
-            pub.qos == node->pubs.at(0).qos)
-          {
-            break;
-          }
-        }
+            pub.qos == node->pubs.at(0).qos;
+          });
         if (erase_it != found_node->pubs.end()) {
           found_node->pubs.erase(erase_it);
           // Bookkeeping
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -71,6 +71,16 @@ void GraphCache::parse_put(const std::string & keyexpr)
         return;
       }
 
+      if (!entity.topic_info().has_value()) {
+        // This should not happen as add_topic_data() is called after validating the existence
+        // of topic_info.
+        RCUTILS_LOG_WARN_NAMED(
+          "rmw_zenoh_cpp",
+          "add_topic_data() called without valid TopicInfo. Report this.");
+        return;
+      }
+
+      const liveliness::TopicInfo topic_info = entity.topic_info().value();
       GraphNode::TopicMap & topic_map = entity.type() ==
         EntityType::Publisher ? graph_node.pubs_ : graph_node.subs_;
       const std::string entity_desc = entity.type() ==
@@ -78,13 +88,13 @@ void GraphCache::parse_put(const std::string & keyexpr)
       const std::size_t pub_count = entity.type() == EntityType::Publisher ? 1 : 0;
       const std::size_t sub_count = !pub_count;
       TopicDataPtr graph_topic_data = std::make_shared<TopicData>(
-        entity.topic_info().value(),
+        topic_info,
         TopicStats{pub_count, sub_count});
 
       GraphNode::TopicDataMap topic_data_map = {
         {graph_topic_data->info_.type_, graph_topic_data}};
       std::pair<GraphNode::TopicMap::iterator, bool> insertion =
-        topic_map.insert(std::make_pair(entity.topic_info()->name_, topic_data_map));
+        topic_map.insert(std::make_pair(topic_info.name_, topic_data_map));
       if (!insertion.second) {
         // A topic with the same name already exists in the node so we append the type.
         std::pair<GraphNode::TopicDataMap::iterator, bool> type_insertion =
@@ -102,25 +112,25 @@ void GraphCache::parse_put(const std::string & keyexpr)
       }
 
       // Bookkeeping: Update graph_topics_ which keeps track of topics across all nodes in the graph
-      GraphNode::TopicMap::iterator cache_topic_it = graph_topics_.find(entity.topic_info()->name_);
+      GraphNode::TopicMap::iterator cache_topic_it = graph_topics_.find(topic_info.name_);
       if (cache_topic_it == graph_topics_.end()) {
         // First time this topic name is added to the graph.
         auto topic_data_ptr = std::make_shared<TopicData>(
-          entity.topic_info().value(),
+          topic_info,
           TopicStats{pub_count, sub_count}
         );
-        graph_topics_[entity.topic_info()->name_] = GraphNode::TopicDataMap{
-          {entity.topic_info()->type_, topic_data_ptr}
+        graph_topics_[topic_info.name_] = GraphNode::TopicDataMap{
+          {topic_info.type_, topic_data_ptr}
         };
       } else {
         // If a TopicData entry for the same type exists in the topic map, update pub/sub counts
         // or else create an new TopicData.
         std::pair<GraphNode::TopicDataMap::iterator, bool> topic_data_insertion =
-          cache_topic_it->second.insert(std::make_pair(entity.topic_info()->type_, nullptr));
+          cache_topic_it->second.insert(std::make_pair(topic_info.type_, nullptr));
         if (topic_data_insertion.second) {
           // A TopicData for the topic_type does not exist.
           topic_data_insertion.first->second = std::make_shared<TopicData>(
-            entity.topic_info().value(),
+            topic_info,
             TopicStats{pub_count, sub_count});
         } else {
           // Update the existing counters.
@@ -133,9 +143,9 @@ void GraphCache::parse_put(const std::string & keyexpr)
         "rmw_zenoh_cpp",
         "Added %s on topic %s with type %s and qos %s to node /%s.",
         entity_desc.c_str(),
-        entity.topic_info()->name_.c_str(),
-        entity.topic_info()->type_.c_str(),
-        entity.topic_info()->qos_.c_str(),
+        topic_info.name_.c_str(),
+        topic_info.type_.c_str(),
+        topic_info.qos_.c_str(),
         graph_node.name_.c_str());
     };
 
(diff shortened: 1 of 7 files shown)


gavanderhoorn:
> This PR updates the GraphCache implementation to support multiple types over the same topic name.

I'm just watching this repository, so likely out-of-the-loop a bit, but what would be a use-case for this? Wouldn't that break the "strong typing" of ROS topics?

clalancette:
> I'm just watching this repository, so likely out-of-the-loop a bit, but what would be a use-case for this? Wouldn't that break the "strong typing" of ROS topics?

This is a supported use-case today:

Terminal 1:
```
$ ros2 topic pub -r 1 /foo std_msgs/msg/String '{data: hello}'
```

Terminal 2:
```
$ ros2 topic pub -r 1 /foo std_msgs/msg/Int32 '{data: 3}'
```

Terminal 3:
```
$ ros2 topic list -t
/foo [std_msgs/msg/Int32, std_msgs/msg/String]
/parameter_events [rcl_interfaces/msg/ParameterEvent]
/rosout [rcl_interfaces/msg/Log]
```

Is it a good idea?  No.  But since the aim is to keep most of the semantics the same with `rmw_zenoh_cpp`, we should support it.

gavanderhoorn:
> This is a supported use-case today

yes, I know ..

I still can't really come up with a good use-case for this though.

> Is it a good idea? No.

Indeed.

Yadunund:
> > I'm just watching this repository, so likely out-of-the-loop a bit, but what would be a use-case for this? Wouldn't that break the "strong typing" of ROS topics?
> 
> This is a supported use-case today:
> 
> Terminal 1:
> 
> ```
> $ ros2 topic pub -r 1 /foo std_msgs/msg/String '{data: hello}'
> ```
> 
> Terminal 2:
> 
> ```
> $ ros2 topic pub -r 1 /foo std_msgs/msg/Int32 '{data: 3}'
> ```
> 
> Terminal 3:
> 
> ```
> $ ros2 topic list -t
> /foo [std_msgs/msg/Int32, std_msgs/msg/String]
> /parameter_events [rcl_interfaces/msg/ParameterEvent]
> /rosout [rcl_interfaces/msg/Log]
> ```
> 
> Is it a good idea? No. But since the aim is to keep most of the semantics the same with `rmw_zenoh_cpp`, we should support it.

@clalancette this should work with PR now. Could you give it a spin?
Edit: `ros2 topic list -v` should also be supported now

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -0,0 +1,362 @@
...
+std::string Entity::node_enclave() const
+{
+  return this->node_info_.enclave_;
+}
+
+///=============================================================================
+const std::optional<TopicInfo> & Entity::topic_info() const
+{
+  return this->topic_info_;
+}
```
clalancette: I'm not a huge fan of APIs like this, that return a reference to something inside of the class.

There are two (related) reasons:
1.  It becomes difficult to tell what the lifetime of this returned value is.  If the caller took a reference, but then this class gets destroyed, it is now holding onto a dangling reference.
2.  If we want to add in locking around `topic_info_`, for instance, it becomes tricky; we have to expose the locking mechanism outside of the class to all callers, and then all callers have to remember to actually lock it.

You can see that I actually changed `node_info` above to three separate APIs that just return the needed data without references.  That is much safer, in my opinion, and I think we should do a similar thing for `topic_info()`, `type()`, and `keyexpr()` in this PR.

clalancette (changes requested):
Overall, this is looking pretty good.

Inline, I've left one major piece of feedback about APIs.

Besides that, I'm noticing segmentation faults and errors when I shut nodes down with this.  I suspect that we are trying to process some liveliness tokens while we are shutting down the context, but I'm not sure.  I think we should probably address that before merging this in.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Thanks for the feedback! I've pushed https://github.com/ros2/rmw_zenoh/pull/70/commits/4488b976fa205a1ec2c8f98b538b49d96934d185. I hope it's fine to let `topic_info()`still return an `std::optional<TopicInfo>`. Else if we split into 3 APIs, we will need to check all 3 params are set each time we want to parse necessary data for pub/subs. To prevent large copies of strings each time, I've updated the implementation to make one copy of `topic_info()` and access the required fields where ever possible.  

> Besides that, I'm noticing segmentation faults and errors when I shut nodes down with this. I suspect that we are trying to process some liveliness tokens while we are shutting down the context, but I'm not sure. I think we should probably address that before merging this in.

Do you have any commands to reproduce this behavior? I wasn't able to do so. But I did encounter this problem early on when I had the `ros2 daemon` running from an older commit of this branch. After `ros2 daemon stop` and rebuilding I never saw that again.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -32,491 +33,347 @@
...
 {
-  std::string token = generate_base_token("MP", domain_id, node_namespace, node_name);
-  token += topic + "/" + type + "/" + qos;
-  return token;
-}
+  std::optional<liveliness::Entity> valid_entity = liveliness::Entity::make(keyexpr);
+  if (!valid_entity.has_value()) {
+    // Error message has already been logged.
+    return;
+  }
```
francocipollone, line 62: Regarding error handling.

When the entity can't be created, the error msg is set and then it is returned. However, when the method (`parse_put`) is being called there is no way to know it failed from the caller's perspective.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -0,0 +1,362 @@
...
+  std::optional<TopicInfo> topic_info = std::nullopt;
+
+  // Populate topic_info if we have a token for an entity other than a node.
+  if (entity_type != EntityType::Node) {
+    if (parts.size() < 8) {
+      RCUTILS_LOG_ERROR_NAMED(
+        "rmw_zenoh_cpp",
+        "Received liveliness token for non-node entity without required parameters.");
+      return std::nullopt;
+    }
```
francocipollone, line 255: nit: Just for consistency, should we check here if we have strictly more than 8 parts? We shouldn't reach that part though.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -32,491 +33,347 @@
...
-        "rmw_zenoh_cpp",
-        "Received invalid liveliness token");
-      return std::nullopt;
-    }
+  // If the namespace did not exist, create it and add the node to the graph and return.
+  NamespaceMap::iterator ns_it = graph_.find(entity.node_namespace());
+  if (ns_it == graph_.end()) {
+    NodeMap node_map = {
+      {entity.node_name(), make_graph_node(entity)}};
+    graph_.insert(std::make_pair(entity.node_namespace(), std::move(node_map)));
```
francocipollone: nit: using `emplace` instead of `insert` might be useful to not rely on the compiler for optimization in the construction/copy of the object

Emplace creates the object in-place instead of moving it with insert if possible.

https://en.cppreference.com/w/cpp/container/unordered_map/emplace

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -32,491 +33,347 @@
...
-///=============================================================================
-bool PublishToken::put(
-  z_owned_session_t * session,
-  const std::string & token)
-{
-  if (!z_session_check(session)) {
-    RCUTILS_SET_ERROR_MSG("The zenoh session is invalid.");
-    return false;
-  }
+      const liveliness::TopicInfo topic_info = entity.topic_info().value();
```
francocipollone, line 83:
```suggestion
      const liveliness::TopicInfo & topic_info = entity.topic_info().value();
```

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -32,491 +33,347 @@
...
+      GraphNode::TopicMap::iterator cache_topic_it = graph_topics_.find(topic_info.name_);
+      if (cache_topic_it == graph_topics_.end()) {
+        // First time this topic name is added to the graph.
+        auto topic_data_ptr = std::make_shared<TopicData>(
+          topic_info,
+          TopicStats{pub_count, sub_count}
+        );
+        graph_topics_[topic_info.name_] = GraphNode::TopicDataMap{
+          {topic_info.type_, topic_data_ptr}
+        };
```
francocipollone: Isn't it the same as `topic_data_map`?

```suggestion
        graph_topics_[topic_info.name_] = topic_data_map;
```

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -137,20 +122,14 @@ class GraphCache final
...
-
-    // Constructor which initialized counters to 0.
-    TopicStats(std::size_t pub_count, std::size_t sub_count);
-  };
-  using TopicStatsPtr = std::unique_ptr<TopicStats>;
-  std::unordered_map<std::string, TopicStatsPtr> graph_topics_ = {};
+  NamespaceMap graph_ = {};
+
+  // Optimize topic lookups across the graph.
+  GraphNode::TopicMap graph_topics_ = {};
```
francocipollone, line 132: Just for documentation's sake. Without this optimized collection for topic lookups, does the time for the queries(for topics) go up considerably? Just wondering why solving the query on demand isn't a chance. Taking into account that keeping the double source of truth might be error-prone.

francocipollone (commented):
Great stuff!  :rocket: 

I left some minor comments 

I tested locally running demo nodes:

```
(docker) franco@:~/ros_dev_ws$ ros2 topic list -v
Published topics:
 * /chatter [std_msgs/msg/String] 1 publisher
 * /parameter_events [rcl_interfaces/msg/ParameterEvent] 3 publishers
 * /rosout [rcl_interfaces/msg/Log] 3 publishers

Subscribed topics:
 * /chatter [std_msgs/msg/String] 1 subscriber
 * /parameter_events [rcl_interfaces/msg/ParameterEvent] 2 subscribers
```

I haven't experienced any error when shutting down the nodes

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: It can be argued how to deal with errors with discovery and whether that should throw and exception or continue the process if the data transport is still working. Maybe something we can iterate on as things get clearer?

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: I'd be in favor of not having any magic numbers like `8` anywhere in the code. Maybe we can update the `Entity` API to return the number of parts for a given `EntityType`.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: We want to create a copy so that the `graph_topic_data` object created in the next line will not have a dangling reference to `TopicInfo` after the lambda goes out of scope.

rmw_zenoh_cpp/src/detail/graph_cache.hpp
Yadunund: With the cache, lookups to check if a certain topic exists in the graph have `O(1)` complexity on average. Else if we have to iterate through `graph_cache_`, it will be `O(n*n'*n'')` where we have to iterate through `n` namepaces, `n'` nodes within each namespace and potentially `n''` topics per node.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/70/commits/357f6588fab1f2f9192e7be418d9f1c3b4956da5

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/70/commits/357f6588fab1f2f9192e7be418d9f1c3b4956da5

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Opened a ticket to work on in the next iteration https://github.com/ros2/rmw_zenoh/issues/72

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Ticketed to work on in the next iteration https://github.com/ros2/rmw_zenoh/issues/71

Yadunund:
Thanks for the review @clalancette and @francocipollone! I've addressed the blocking feedback and opened tickets for the others. Will merge this in now to continue impl of introspection methods.

PR #82, opened 2023-12-04
Enables CI when pushing rolling branch.

francocipollone:
I wanted to double-check check CI is passing after last zenoh-c commit and found this minor bug.

### Summary

 - Trigger CI when commit is pushed to default branch (rolling)
 - Enables workflow_dispatch

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,7 +2,8 @@ name: build
 on:
   pull_request:
   push:
-    branches: [ main ]
+    branches: [ rolling ]
+  workflow_dispatch:
 defaults:
   run:
     shell: bash


francocipollone:
CI failing due to: https://github.com/eclipse-zenoh/zenoh-c/pull/214

francocipollone:
Retriggering CI after https://github.com/eclipse-zenoh/zenoh-c/pull/214 is merged

.github/workflows/build.yaml
```diff
@@ -2,7 +2,8 @@ name: build
 on:
   pull_request:
   push:
-    branches: [ main ]
+    branches: [ rolling ]
+  workflow_dispatch:
```
Yadunund, line 6: Do we still need the `workflow_dispatch` key?

francocipollone: As we are targeting master for the zenoh dependency I found it useful to be able to run it via `workflow_dispatch` to verify that everything works as expected when a new commit is pushed in zenoh-c .

clalancette: > As we are targeting master for the zenoh dependency I found it useful to be able to run it via `workflow_dispatch` to verify that everything works as expected when a new commit is pushed in zenoh-c .

I'm fine with it for now, though I expect very soon that we should pin the zenoh-c commit that we are using, and only update it when we need a new feature.

clalancette (approved):

PR #86, opened 2023-12-15
Adds base support for services.

francocipollone:
### Summary
 - Provides support for services
 - Graph cache update is not tackled in this PR.

### Notes on the implementation
 - Queryables are being used for the service client interaction. (instead of topics as dds implementations) 
 - Main rmw methods to implement from the server side
   - rmw_create_service
   - rmw_take_request
   - rmw_send_response
 - Main rmw methods to implement from the clilent side:
   - rmw_create_client
   - rmw_send_request
   - rmw_take_response
 
      
### Pendings
 - [x] :exclamation: `z_query_value` is failing to obtain the message when using `z_owned_query` approach, even replicated in simpler script (Notified zettascale folks) :exclamation:(blocker)  
 - [ ] Define headers (request/response) and sequence id information.
 - [x] Verify waits added to the rmw_wait method.
 - [ ] Verify keyexpression for the service 
 - [x] Fix style
 - [ ] `rmw_service_server_is_available` method needs graph information. 
 - [ ] Update graph cache

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -81,13 +81,6 @@ void sub_data_handler(
   }
 }
 
-
-//==============================================================================
-std::size_t rmw_service_data_t::get_new_uid()
-{
-  return client_count++;
-}
-
 //==============================================================================
 void service_data_handler(const z_query_t * query, void * data)
 {
@@ -115,10 +108,7 @@ void service_data_handler(const z_query_t * query, void * data)
   // Get the query parameters and payload
   {
     std::lock_guard<std::mutex> lock(service_data->query_queue_mutex);
-    const std::size_t client_id = service_data->get_new_uid();
-    service_data->id_query_map.emplace(
-      std::make_pair(client_id, z_query_clone(query)));
-    service_data->to_take.push_back(client_id);
+    service_data->query_queue.push_back(z_query_clone(query));
   }
   {
     // Since we added new data, trigger the guard condition if it is available
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -17,6 +17,7 @@
 #include <mutex>
 #include <utility>
 
+#include "rcpputils/scope_exit.hpp"
 #include "rcutils/logging_macros.h"
 
 #include "rmw_data_types.hpp"
@@ -34,6 +35,10 @@ void sub_data_handler(
   void * data)
 {
   z_owned_str_t keystr = z_keyexpr_to_string(sample->keyexpr);
+  auto drop_keystr = rcpputils::make_scope_exit(
+    [&keystr]() {
+      z_drop(z_move(keystr));
+    });
 
   auto sub_data = static_cast<rmw_subscription_data_t *>(data);
   if (sub_data == nullptr) {
@@ -74,8 +79,6 @@ void sub_data_handler(
       sub_data->condition->notify_one();
     }
   }
-
-  z_drop(z_move(keystr));
 }
 
 
@@ -93,6 +96,10 @@ void service_data_handler(const z_query_t * query, void * data)
     "[service_data_handler] triggered"
   );
   z_owned_str_t keystr = z_keyexpr_to_string(z_query_keyexpr(query));
+  auto drop_keystr = rcpputils::make_scope_exit(
+    [&keystr]() {
+      z_drop(z_move(keystr));
+    });
 
   rmw_service_data_t * service_data = static_cast<rmw_service_data_t *>(data);
   if (service_data == nullptr) {
@@ -102,7 +109,6 @@ void service_data_handler(const z_query_t * query, void * data)
       "service for %s",
       z_loan(keystr)
     );
-    z_drop(z_move(keystr));
     return;
   }
 
@@ -121,8 +127,6 @@ void service_data_handler(const z_query_t * query, void * data)
       service_data->condition->notify_one();
     }
   }
-
-  z_drop(z_move(keystr));
 }
 
 //==============================================================================
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -22,13 +22,13 @@
 #include "rmw_data_types.hpp"
 
 ///==============================================================================
-
 saved_msg_data::saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16])
 : payload(p), recv_timestamp(recv_ts)
 {
   memcpy(publisher_gid, pub_gid, 16);
 }
 
+//==============================================================================
 void sub_data_handler(
   const z_sample_t * sample,
   void * data)
@@ -79,12 +79,14 @@ void sub_data_handler(
 }
 
 
-unsigned int rmw_service_data_t::get_new_uid()
+//==============================================================================
+std::size_t rmw_service_data_t::get_new_uid()
 {
   return client_count++;
 }
 
-void service_data_handler(const z_query_t * query, void * service_data)
+//==============================================================================
+void service_data_handler(const z_query_t * query, void * data)
 {
   RCUTILS_LOG_INFO_NAMED(
     "rmw_zenoh_cpp",
@@ -92,8 +94,8 @@ void service_data_handler(const z_query_t * query, void * service_data)
   );
   z_owned_str_t keystr = z_keyexpr_to_string(z_query_keyexpr(query));
 
-  auto rmw_service_data = static_cast<rmw_service_data_t *>(service_data);
-  if (rmw_service_data == nullptr) {
+  rmw_service_data_t * service_data = static_cast<rmw_service_data_t *>(data);
+  if (service_data == nullptr) {
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_zenoh_cpp",
       "Unable to obtain rmw_service_data_t from data for "
@@ -106,43 +108,31 @@ void service_data_handler(const z_query_t * query, void * service_data)
 
   // Get the query parameters and payload
   {
-    std::lock_guard<std::mutex> lock(rmw_service_data->query_queue_mutex);
-
-    const unsigned int client_id = rmw_service_data->get_new_uid();
-    rmw_service_data->id_query_map.emplace(
-      std::make_pair(client_id, std::make_unique<z_owned_query_t>(z_query_clone(query))));
-    rmw_service_data->to_take.push_back(client_id);
-
-
+    std::lock_guard<std::mutex> lock(service_data->query_queue_mutex);
+    const std::size_t client_id = service_data->get_new_uid();
+    service_data->id_query_map.emplace(
+      std::make_pair(client_id, z_query_clone(query)));
+    service_data->to_take.push_back(client_id);
+  }
+  {
     // Since we added new data, trigger the guard condition if it is available
-    std::lock_guard<std::mutex> internal_lock(rmw_service_data->internal_mutex);
-    if (rmw_service_data->condition != nullptr) {
-      rmw_service_data->condition->notify_one();
+    std::lock_guard<std::mutex> internal_lock(service_data->internal_mutex);
+    if (service_data->condition != nullptr) {
+      service_data->condition->notify_one();
     }
   }
 
   z_drop(z_move(keystr));
 }
 
-void client_data_handler(z_owned_reply_t * reply, void * client_data)
+//==============================================================================
+void client_data_handler(z_owned_reply_t * reply, void * data)
 {
-  auto rmw_client_data = static_cast<rmw_client_data_t *>(client_data);
-  if (rmw_client_data == nullptr) {
+  auto client_data = static_cast<rmw_client_data_t *>(data);
+  if (client_data == nullptr) {
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_zenoh_cpp",
-      "Unable to obtain rmw_client_data_t "
-    );
-    return;
-  }
-  RCUTILS_LOG_INFO_NAMED(
-    "rmw_zenoh_cpp",
-    "[client_data_handler] triggered for %s",
-    rmw_client_data->service_name
-  );
-  if (!z_check(*reply)) {
-    RCUTILS_LOG_ERROR_NAMED(
-      "rmw_zenoh_cpp",
-      "z_check returned False"
+      "Unable to obtain client_data_t "
     );
     return;
   }
@@ -160,29 +150,16 @@ void client_data_handler(z_owned_reply_t * reply, void * client_data)
     );
     return;
   }
-
-  z_sample_t sample = z_reply_ok(reply);
-
-  z_owned_str_t keystr = z_keyexpr_to_string(sample.keyexpr);
-
-  RCUTILS_LOG_DEBUG_NAMED(
-    "rmw_zenoh_cpp",
-    "[client_data_handler] keyexpr of sample: %s",
-    z_loan(keystr)
-  );
-
   {
-    std::lock_guard<std::mutex> msg_lock(rmw_client_data->message_mutex);
-    rmw_client_data->message = std::make_unique<saved_msg_data>(
-      zc_sample_payload_rcinc(&sample), sample.timestamp.time, sample.timestamp.id.id);
+    std::lock_guard<std::mutex> msg_lock(client_data->message_mutex);
+    // Take ownership of the reply.
+    client_data->replies.emplace_back(*reply);
+    *reply = z_reply_null();
   }
   {
-    std::lock_guard<std::mutex> internal_lock(rmw_client_data->internal_mutex);
-    if (rmw_client_data->condition != nullptr) {
-      rmw_client_data->condition->notify_one();
+    std::lock_guard<std::mutex> internal_lock(client_data->internal_mutex);
+    if (client_data->condition != nullptr) {
+      client_data->condition->notify_one();
     }
   }
-
-  z_reply_drop(reply);
-  z_drop(z_move(keystr));
 }
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -78,15 +78,6 @@ void sub_data_handler(
   z_drop(z_move(keystr));
 }
 
-saved_queryable_data::saved_queryable_data(z_owned_query_t query)
-: query(query)
-{}
-
-saved_queryable_data::~saved_queryable_data()
-{
-  z_query_drop(&query);
-}
-
 
 unsigned int rmw_service_data_t::get_new_uid()
 {
@@ -109,6 +100,7 @@ void service_data_handler(const z_query_t * query, void * service_data)
       "service for %s",
       z_loan(keystr)
     );
+    z_drop(z_move(keystr));
     return;
   }
 
@@ -118,7 +110,7 @@ void service_data_handler(const z_query_t * query, void * service_data)
 
     const unsigned int client_id = rmw_service_data->get_new_uid();
     rmw_service_data->id_query_map.emplace(
-      std::make_pair(client_id, std::make_unique<saved_queryable_data>(z_query_clone(query))));
+      std::make_pair(client_id, std::make_unique<z_owned_query_t>(z_query_clone(query))));
     rmw_service_data->to_take.push_back(client_id);
 
 
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -139,10 +139,17 @@ void client_data_handler(z_owned_reply_t * reply, void * client_data)
     "[client_data_handler] triggered for %s",
     rmw_client_data->service_name
   );
-  if (!z_reply_check(reply)) {
+  if (!z_check(*reply)) {
     RCUTILS_LOG_ERROR_NAMED(
       "rmw_zenoh_cpp",
       "z_check returned False"
+    );
+    return;
+  }
+  if (!z_reply_check(reply)) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "z_reply_check returned False"
     );
     return;
   }
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -22,6 +22,13 @@
 #include "rmw_data_types.hpp"
 
 ///==============================================================================
+
+saved_msg_data::saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16])
+: payload(p), recv_timestamp(recv_ts)
+{
+  memcpy(publisher_gid, pub_gid, 16);
+}
+
 void sub_data_handler(
   const z_sample_t * sample,
   void * data)
@@ -69,6 +76,21 @@ void sub_data_handler(
   }
 
   z_drop(z_move(keystr));
+}
+
+saved_queryable_data::saved_queryable_data(z_owned_query_t query)
+: query(query)
+{}
+
+saved_queryable_data::~saved_queryable_data()
+{
+  z_query_drop(&query);
+}
+
+
+unsigned int rmw_service_data_t::get_new_uid()
+{
+  return client_count++;
 }
 
 void service_data_handler(const z_query_t * query, void * service_data)
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -70,3 +70,98 @@ void sub_data_handler(
 
   z_drop(z_move(keystr));
 }
+
+void service_data_handler(const z_query_t * query, void * service_data)
+{
+  RCUTILS_LOG_INFO_NAMED(
+    "rmw_zenoh_cpp",
+    "[service_data_handler] triggered"
+  );
+  z_owned_str_t keystr = z_keyexpr_to_string(z_query_keyexpr(query));
+
+  auto rmw_service_data = static_cast<rmw_service_data_t *>(service_data);
+  if (rmw_service_data == nullptr) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to obtain rmw_service_data_t from data for "
+      "service for %s",
+      z_loan(keystr)
+    );
+    return;
+  }
+
+  // Get the query parameters and payload
+  {
+    std::lock_guard<std::mutex> lock(rmw_service_data->query_queue_mutex);
+
+    const unsigned int client_id = rmw_service_data->get_new_uid();
+    rmw_service_data->id_query_map.emplace(
+      std::make_pair(client_id, std::make_unique<saved_queryable_data>(z_query_clone(query))));
+    rmw_service_data->to_take.push_back(client_id);
+
+
+    // Since we added new data, trigger the guard condition if it is available
+    std::lock_guard<std::mutex> internal_lock(rmw_service_data->internal_mutex);
+    if (rmw_service_data->condition != nullptr) {
+      rmw_service_data->condition->notify_one();
+    }
+  }
+
+  z_drop(z_move(keystr));
+}
+
+void client_data_handler(z_owned_reply_t * reply, void * client_data)
+{
+  auto rmw_client_data = static_cast<rmw_client_data_t *>(client_data);
+  if (rmw_client_data == nullptr) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to obtain rmw_client_data_t "
+    );
+    return;
+  }
+  RCUTILS_LOG_INFO_NAMED(
+    "rmw_zenoh_cpp",
+    "[client_data_handler] triggered for %s",
+    rmw_client_data->service_name
+  );
+  if (!z_reply_check(reply)) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "z_check returned False"
+    );
+    return;
+  }
+  if (!z_reply_is_ok(reply)) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "z_reply_is_ok returned False"
+    );
+    return;
+  }
+
+  z_sample_t sample = z_reply_ok(reply);
+
+  z_owned_str_t keystr = z_keyexpr_to_string(sample.keyexpr);
+
+  RCUTILS_LOG_DEBUG_NAMED(
+    "rmw_zenoh_cpp",
+    "[client_data_handler] keyexpr of sample: %s",
+    z_loan(keystr)
+  );
+
+  {
+    std::lock_guard<std::mutex> msg_lock(rmw_client_data->message_mutex);
+    rmw_client_data->message = std::make_unique<saved_msg_data>(
+      zc_sample_payload_rcinc(&sample), sample.timestamp.time, sample.timestamp.id.id);
+  }
+  {
+    std::lock_guard<std::mutex> internal_lock(rmw_client_data->internal_mutex);
+    if (rmw_client_data->condition != nullptr) {
+      rmw_client_data->condition->notify_one();
+    }
+  }
+
+  z_reply_drop(reply);
+  z_drop(z_move(keystr));
+}
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -99,11 +99,7 @@ void sub_data_handler(const z_sample_t * sample, void * sub_data);
 
 struct saved_msg_data
 {
-  explicit saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16])
-  : payload(p), recv_timestamp(recv_ts)
-  {
-    memcpy(publisher_gid, pub_gid, 16);
-  }
+  explicit saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16]);
 
   zc_owned_payload_t payload;
   uint64_t recv_timestamp;
@@ -144,22 +140,17 @@ void client_data_handler(z_owned_reply_t * reply, void * client_data);
 
 struct saved_queryable_data
 {
-  explicit saved_queryable_data(z_owned_query_t query)
-  : query(query)
-  {
-  }
+  explicit saved_queryable_data(z_owned_query_t query);
+  ~saved_queryable_data();
 
-  const z_owned_query_t query;
+  z_owned_query_t query;
 };
 
 ///==============================================================================
 
 struct rmw_service_data_t
 {
-  unsigned int get_new_uid()
-  {
-    return client_count++;
-  }
+  unsigned int get_new_uid();
 
   const char * zn_queryable_key;
   z_owned_queryable_t zn_queryable;
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -156,7 +156,7 @@ struct rmw_service_data_t
   rmw_context_t * context;
 
   // Map to store the query id and the query.
-  // The query handler is saved as it is needed to answer the query later on.
+  // The query handle is saved as it is needed to answer the query later on.
   std::unordered_map<std::size_t, z_owned_query_t> id_query_map;
   // The query id's of the queries that need to be processed.
   std::deque<std::size_t> to_take;
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -137,14 +137,13 @@ void service_data_handler(const z_query_t * query, void * service_data);
 
 void client_data_handler(z_owned_reply_t * reply, void * client_data);
 
-
 ///==============================================================================
 
 struct rmw_service_data_t
 {
-  unsigned int get_new_uid();
+  std::size_t get_new_uid();
 
-  const char * keyexpr;
+  z_owned_keyexpr_t keyexpr;
   z_owned_queryable_t qable;
 
   const void * request_type_support_impl;
@@ -157,27 +156,27 @@ struct rmw_service_data_t
 
   // Map to store the query id and the query.
   // The query handler is saved as it is needed to answer the query later on.
-  std::unordered_map<unsigned int, std::unique_ptr<z_owned_query_t>> id_query_map;
+  std::unordered_map<std::size_t, z_owned_query_t> id_query_map;
   // The query id's of the queries that need to be processed.
-  std::deque<unsigned int> to_take;
+  std::deque<std::size_t> to_take;
   std::mutex query_queue_mutex;
 
   std::mutex internal_mutex;
   std::condition_variable * condition{nullptr};
 
-  unsigned int client_count{};
+  std::size_t client_count = 0;
 };
 
 ///==============================================================================
 
 struct rmw_client_data_t
 {
-  const char * service_name;
+  z_owned_keyexpr_t keyexpr;
 
   z_owned_closure_reply_t zn_closure_reply;
 
   std::mutex message_mutex;
-  std::unique_ptr<saved_msg_data> message;
+  std::vector<z_owned_reply_t> replies;
 
   const void * request_type_support_impl;
   const void * response_type_support_impl;
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -138,22 +138,14 @@ void service_data_handler(const z_query_t * query, void * service_data);
 void client_data_handler(z_owned_reply_t * reply, void * client_data);
 
 
-struct saved_queryable_data
-{
-  explicit saved_queryable_data(z_owned_query_t query);
-  ~saved_queryable_data();
-
-  z_owned_query_t query;
-};
-
 ///==============================================================================
 
 struct rmw_service_data_t
 {
   unsigned int get_new_uid();
 
-  const char * zn_queryable_key;
-  z_owned_queryable_t zn_queryable;
+  const char * keyexpr;
+  z_owned_queryable_t qable;
 
   const void * request_type_support_impl;
   const void * response_type_support_impl;
@@ -165,7 +157,7 @@ struct rmw_service_data_t
 
   // Map to store the query id and the query.
   // The query handler is saved as it is needed to answer the query later on.
-  std::unordered_map<unsigned int, std::unique_ptr<saved_queryable_data>> id_query_map;
+  std::unordered_map<unsigned int, std::unique_ptr<z_owned_query_t>> id_query_map;
   // The query id's of the queries that need to be processed.
   std::deque<unsigned int> to_take;
   std::mutex query_queue_mutex;
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -25,6 +25,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
+#include <vector>
 
 #include "rcutils/allocator.h"
 
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -133,7 +133,6 @@ struct rmw_subscription_data_t
 
 ///==============================================================================
 
-// z_owned_closure_query_t
 void service_data_handler(const z_query_t * query, void * service_data);
 
 void client_data_handler(z_owned_reply_t * reply, void * client_data);
@@ -155,17 +154,16 @@ struct rmw_service_data_t
 
   rmw_context_t * context;
 
-  // Map to store the query id and the query.
-  // The query handle is saved as it is needed to answer the query later on.
-  std::unordered_map<std::size_t, z_owned_query_t> id_query_map;
-  // The query id's of the queries that need to be processed.
-  std::deque<std::size_t> to_take;
+  // Deque to store the queries in the order they arrive.
+  std::deque<z_owned_query_t> query_queue;
   std::mutex query_queue_mutex;
+
+  // Map to store the sequence_number -> query_id
+  std::map<int64_t, z_owned_query_t> sequence_to_query_map;
+  std::mutex sequence_to_query_map_mutex;
 
   std::mutex internal_mutex;
   std::condition_variable * condition{nullptr};
-
-  std::size_t client_count = 0;
 };
 
 ///==============================================================================
@@ -177,7 +175,7 @@ struct rmw_client_data_t
   z_owned_closure_reply_t zn_closure_reply;
 
   std::mutex message_mutex;
-  std::vector<z_owned_reply_t> replies;
+  std::deque<z_owned_reply_t> replies;
 
   const void * request_type_support_impl;
   const void * response_type_support_impl;
@@ -189,6 +187,8 @@ struct rmw_client_data_t
 
   std::mutex internal_mutex;
   std::condition_variable * condition{nullptr};
+
+  size_t sequence_number{1};
 };
 
 #endif  // DETAIL__RMW_DATA_TYPES_HPP_
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -22,6 +22,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <unordered_map>
 #include <unordered_set>
 #include <utility>
 
@@ -31,6 +32,7 @@
 
 #include "graph_cache.hpp"
 #include "message_type_support.hpp"
+#include "service_type_support.hpp"
 
 /// Structs for various type erased data fields.
 
@@ -126,6 +128,85 @@ struct rmw_subscription_data_t
 
   size_t queue_depth;
   bool reliable;
+
+  std::mutex internal_mutex;
+  std::condition_variable * condition{nullptr};
+};
+
+
+///==============================================================================
+
+// z_owned_closure_query_t
+void service_data_handler(const z_query_t * query, void * service_data);
+
+void client_data_handler(z_owned_reply_t * reply, void * client_data);
+
+
+struct saved_queryable_data
+{
+  explicit saved_queryable_data(z_owned_query_t query)
+  : query(query)
+  {
+  }
+
+  const z_owned_query_t query;
+};
+
+///==============================================================================
+
+struct rmw_service_data_t
+{
+  unsigned int get_new_uid()
+  {
+    return client_count++;
+  }
+
+  const char * zn_queryable_key;
+  z_owned_queryable_t zn_queryable;
+
+  const void * request_type_support_impl;
+  const void * response_type_support_impl;
+  const char * typesupport_identifier;
+  RequestTypeSupport * request_type_support;
+  ResponseTypeSupport * response_type_support;
+
+  rmw_context_t * context;
+
+  // Map to store the query id and the query.
+  // The query handler is saved as it is needed to answer the query later on.
+  std::unordered_map<unsigned int, std::unique_ptr<saved_queryable_data>> id_query_map;
+  // The query id's of the queries that need to be processed.
+  std::deque<unsigned int> to_take;
+  std::mutex query_queue_mutex;
+
+  std::mutex internal_mutex;
+  std::condition_variable * condition{nullptr};
+
+  unsigned int client_count{};
+};
+
+///==============================================================================
+
+struct rmw_client_data_t
+{
+  const char * service_name;
+
+  // TODO(francocipollone): Remove this. For some reason if I remove this(not being even used) it
+  //                        ends up panicking when calling the service. Something is missing.
+  z_owned_reply_channel_t zn_reply_channel;
+  z_owned_closure_reply_t zn_closure_reply;
+
+
+  std::mutex message_mutex;
+  std::unique_ptr<saved_msg_data> message;
+
+  const void * request_type_support_impl;
+  const void * response_type_support_impl;
+  const char * typesupport_identifier;
+  RequestTypeSupport * request_type_support;
+  ResponseTypeSupport * response_type_support;
+
+  rmw_context_t * context;
 
   std::mutex internal_mutex;
   std::condition_variable * condition{nullptr};
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -174,11 +174,7 @@ struct rmw_client_data_t
 {
   const char * service_name;
 
-  // TODO(francocipollone): Remove this. For some reason if I remove this(not being even used) it
-  //                        ends up panicking when calling the service. Something is missing.
-  z_owned_reply_channel_t zn_reply_channel;
   z_owned_closure_reply_t zn_closure_reply;
-
 
   std::mutex message_mutex;
   std::unique_ptr<saved_msg_data> message;
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -2080,7 +2080,8 @@ static int64_t get_sequence_num_from_attachment(const z_attachment_t * const att
     return -1;
   } else if (errno != 0) {
     // Some other error occurred, which may include overflow or underflow
-    RMW_SET_ERROR_MSG("An undefined error occurred while getting the sequence number, this may be an overflow");
+    RMW_SET_ERROR_MSG(
+      "An undefined error occurred while getting the sequence number, this may be an overflow");
     return -1;
   }
 
@@ -2505,7 +2506,9 @@ rmw_take_request(
   // Add this query to the map, so that rmw_send_response can quickly look it up later
   {
     std::lock_guard<std::mutex> lock(service_data->sequence_to_query_map_mutex);
-    if (service_data->sequence_to_query_map.find(sequence_number) != service_data->sequence_to_query_map.end()) {
+    if (service_data->sequence_to_query_map.find(sequence_number) !=
+      service_data->sequence_to_query_map.end())
+    {
       RMW_SET_ERROR_MSG("duplicate sequence number in the map");
       return RMW_RET_ERROR;
     }
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -2284,6 +2284,7 @@ rmw_create_service(
     RMW_SET_ERROR_MSG("unable to create zenoh keyexpr.");
     return nullptr;
   }
+  service_data->keyexpr = z_keyexpr_to_string(z_loan(keyexpr))._cstr;
 
   RCUTILS_LOG_INFO_NAMED(
     "rmw_zenoh_cpp", "[rmw_create_service] keyexpr: %s",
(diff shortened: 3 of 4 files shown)


francocipollone:
### Test it!
Terminal 1
`ros2 run demo_nodes_cpp add_two_ints_server` 

Terminal 2
`ros2 run demo_nodes_cpp add_two_ints_client`

The client is calling the service with 2 and 3 so the expected result is 5

### Output
```
[INFO] [1703192805.637515098] [add_two_ints_client]: Result of add_two_ints: 5
```

### Issue (FIXED)
This is more or less the logging:

```
[INFO] [1703192805.637014300] [rmw_zenoh_cpp]: [client_data_handler] triggered for /add_two_ints
[INFO] [1703192805.637093573] [rmw_zenoh_cpp]: [client_data_handler] triggered for /add_two_ints
[ERROR] [1703192805.637109494] [rmw_zenoh_cpp]: z_reply_is_ok returned False
[INFO] [1703192805.637376320] [rmw_zenoh_cpp]: [rmw_take_response]

```
I added a logging message to get when the `client_data_handler` callback is triggered and for some reason, it is being called twice, with the second one containing a garbage `reply`. `client_data_handler` is the callback that is executed for a `z_get` called in the client side. Here is some explanation of the architecture so far

image

I added some comments in the code as I still have to verify if this is the correct workflow to be used with Zenoh queryable or if the use of Zenoh reply channels would be better in this case.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +133,83 @@ struct rmw_subscription_data_t
...
+
+///==============================================================================
+
+struct rmw_client_data_t
+{
+  const char * service_name;
+
+  // TODO(francocipollone): Remove this. For some reason if I remove this(not being even used) it
+  //                        ends up panicking when calling the service. Something is missing.
+  z_owned_reply_channel_t zn_reply_channel;
```
francocipollone: Sorry for this comment, I think that at some point I am using the api wrongly otherwise I don't explain the behavior here stated.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2054,6 +2763,29 @@ rmw_wait(
     }
   }
 
+
+  if (services) {
```
francocipollone, line 2907: I wasn't able to fully grasp the idea of how waits and waits events are being used by `rcl` for the `rmw`. So the code is mere deducted from the other cases. In case you could provide more information here would be great for my understanding :smile: cc: @clalancette @Yadunund

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +133,83 @@ struct rmw_subscription_data_t
...
+
+// z_owned_closure_query_t
+void service_data_handler(const z_query_t * query, void * service_data);
+
+void client_data_handler(z_owned_reply_t * reply, void * client_data);
+
+
+struct saved_queryable_data
+{
+  explicit saved_queryable_data(z_owned_query_t query)
```
Yadunund: Move implementation to .cpp file

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +133,83 @@ struct rmw_subscription_data_t
...
+  }
+
+  const z_owned_query_t query;
+};
+
+///==============================================================================
+
+struct rmw_service_data_t
+{
+  unsigned int get_new_uid()
```
Yadunund: Move implementation to .cpp file.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +133,83 @@ struct rmw_subscription_data_t
...
+
+
+struct saved_queryable_data
+{
+  explicit saved_queryable_data(z_owned_query_t query)
+  : query(query)
+  {
+  }
+
+  const z_owned_query_t query;
```
Yadunund: I don't think this needs to be `const`. Instead you can instantiate an `save_queryable_data` object as const.
```suggestion
  z_owned_query_t query;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +133,83 @@ struct rmw_subscription_data_t
...
+};
```
Yadunund: Overwrite default destructor to `z_drop(query)`.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1628,25 +1655,235 @@ rmw_return_loaned_message_from_subscription(
...
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    client,
+    client->implementation_identifier,
+    rmw_zenoh_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+
+  rcutils_allocator_t * allocator = &node->context->options.allocator;
+
+  auto client_data = static_cast<rmw_client_data_t *>(client->data);
```
Yadunund: Check `client_data` is valid before dereferencing below.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1628,25 +1655,235 @@ rmw_return_loaned_message_from_subscription(
...
+
+  // CLEANUP ===================================================================
+  allocator->deallocate(client_data->request_type_support, allocator->state);
+  allocator->deallocate(client_data->response_type_support, allocator->state);
+  allocator->deallocate(client->data, allocator->state);
```
Yadunund, line 1907: Make sure to call `z_drop()` on `z_owned_XX` objects in `client_data` before deallocating

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1657,10 +1894,93 @@ rmw_send_request(
...
+
+  // TODO(francocipollone): Do I really need the sequency number here?
+  *sequence_id = 0;
+
+
+  // Send request
+  z_get_options_t opts = z_get_options_default();
+  opts.value.payload = z_bytes_t{data_length, reinterpret_cast<const uint8_t *>(request_bytes)};
+  opts.value.encoding = z_encoding(Z_ENCODING_PREFIX_EMPTY, NULL);
+
```
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1657,10 +1894,93 @@ rmw_send_request(
...
+  {
+    allocator->deallocate(request_bytes, allocator->state);
+    return RMW_RET_ERROR;
+  }
+
+  size_t data_length = ser.getSerializedDataLength();
+
+  // TODO(francocipollone): Do I really need the sequency number here?
+  *sequence_id = 0;
+
```
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1739,20 +2122,199 @@ rmw_create_service(
...
+  RMW_CHECK_FOR_NULL_WITH_MSG(
+    rmw_service->service_name,
+    "failed to allocate service name",
+    return nullptr);
+
+  auto free_service_name = rcpputils::make_scope_exit(
+    [rmw_service, allocator]() {
+      allocator->deallocate(const_cast<char *>(rmw_service->service_name), allocator->state);
+    });
+
```
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1739,20 +2122,199 @@ rmw_create_service(
...
+    RMW_SET_ERROR_MSG("unable to create zenoh queryable");
     return nullptr;
   }
 
-  return service;
+  auto undeclare_z_queryable = rcpputils::make_scope_exit(
+    [service_data]() {
+      z_undeclare_queryable(z_move(service_data->zn_queryable));
+    });
+
```
Yadunund, line 2390:
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1778,11 +2342,84 @@ rmw_take_request(
...
+    service,
+    service->implementation_identifier,
+    rmw_zenoh_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+
+  RMW_CHECK_FOR_NULL_WITH_MSG(
+    service->service_name, "service has no service name", RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_FOR_NULL_WITH_MSG(
+    service->data, "service implementation pointer is null", RMW_RET_INVALID_ARGUMENT);
+
```
Yadunund, line 2485:
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1766,6 +2328,8 @@ rmw_destroy_service(rmw_node_t * node, rmw_service_t * service)
     return RMW_RET_ERROR;
   }
   rmw_service_free(service);
+
```
Yadunund, line 2451: Before calling `rmf_service_free()` we should deallocate everything in `service->data`. Including calling `z_drop()` on any `z_owned_XX` objects

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1657,10 +1894,93 @@ rmw_send_request(
...
+  auto * client_data = static_cast<rmw_client_data_t *>(client->data);
+
+  rmw_context_impl_s * context_impl = static_cast<rmw_context_impl_s *>(
+    client_data->context->impl);
+
+  // Serialize data
+
+  rcutils_allocator_t * allocator = &(client_data->context->options.allocator);
+
+  size_t max_data_length = (
```
Yadunund, line 1968: The codeblock from here 1955 is repeated in many places in the codebase. Can we refactor it out into a function that can be called wherever needed?

Yadunund (changes requested):
Some initial comments before I do a deeper review!

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+
+///==============================================================================
+
+// z_owned_closure_query_t
+void service_data_handler(const z_query_t * query, void * service_data);
+
+void client_data_handler(z_owned_reply_t * reply, void * client_data);
+
+
+struct saved_queryable_data
```
Yadunund: Why do we need this struct if it only has one field? Can we not store the `z_owned_query_t` object directly?

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+  z_owned_query_t query;
+};
+
+///==============================================================================
+
+struct rmw_service_data_t
+{
+  unsigned int get_new_uid();
+
+  const char * zn_queryable_key;
```
Yadunund: Nit: Follow naming style used in other places.
```suggestion
  const char * keyexpr;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+  z_owned_queryable_t zn_queryable;
```
Yadunund: Nit: Follow naming style used in other places.
```suggestion
  z_owned_queryable_t qable;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+  std::condition_variable * condition{nullptr};
+
+  unsigned int client_count{};
+};
+
+///==============================================================================
+
+struct rmw_client_data_t
+{
+  const char * service_name;
```
Yadunund:
```suggestion
  const char * keyexpr;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -22,6 +22,13 @@
 #include "rmw_data_types.hpp"
 
 ///==============================================================================
+
+saved_msg_data::saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16])
+: payload(p), recv_timestamp(recv_ts)
+{
+  memcpy(publisher_gid, pub_gid, 16);
```
Yadunund, line 29:
```suggestion
  memcpy(publisher_gid, pub_gid, sizeof(pub_gid);
```

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -70,3 +77,113 @@ void sub_data_handler(
 
   z_drop(z_move(keystr));
 }
+
+saved_queryable_data::saved_queryable_data(z_owned_query_t query)
+: query(query)
```
Yadunund: See comment in .hpp about not requiring this stuct.
But I also wonder if this definition causes undefined behavior since the variable name in the function signature and that of the stored field as both `query`.
Assuming we really need this struct, I think the implementation should be something like
```bash
saved_queryable_data::saved_queryable_data(z_owned_query_t _query)
: query(z_move(_query))
``` 

so that the ownership is properly transferred to the struct.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+  const void * response_type_support_impl;
+  const char * typesupport_identifier;
+  RequestTypeSupport * request_type_support;
+  ResponseTypeSupport * response_type_support;
+
+  rmw_context_t * context;
+
+  // Map to store the query id and the query.
+  // The query handler is saved as it is needed to answer the query later on.
+  std::unordered_map<unsigned int, std::unique_ptr<saved_queryable_data>> id_query_map;
```
Yadunund: Is there a reason this map cannot simply be
```suggestion
  std::unordered_map<unsigned int, z_owned_queryable_t> id_query_map;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+  // The query id's of the queries that need to be processed.
+  std::deque<unsigned int> to_take;
```
Yadunund: Why do we need this separate queue? Can we not treat `id_query_map` is a queue for queries that need to be taken? Once they're taken we can erase then from the map leaving ones that have not been taken yet.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -70,3 +77,113 @@ void sub_data_handler(
...
+
+  auto rmw_service_data = static_cast<rmw_service_data_t *>(service_data);
+  if (rmw_service_data == nullptr) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to obtain rmw_service_data_t from data for "
+      "service for %s",
+      z_loan(keystr)
+    );
+    return;
```
Yadunund, line 105: Before returning we should call `z_drop(z_move(keystr))`

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -70,3 +77,113 @@ void sub_data_handler(
...
+  RCUTILS_LOG_DEBUG_NAMED(
+    "rmw_zenoh_cpp",
+    "[client_data_handler] keyexpr of sample: %s",
+    z_loan(keystr)
+  );
+
+  {
+    std::lock_guard<std::mutex> msg_lock(rmw_client_data->message_mutex);
+    rmw_client_data->message = std::make_unique<saved_msg_data>(
+      zc_sample_payload_rcinc(&sample), sample.timestamp.time, sample.timestamp.id.id);
```
Yadunund: I not very sure but it feels like relying on `zc_sample_payload_rcinc` could be problematic here since the sample is owned within the closure which will be dropped by zenoh as per https://zenoh-c.readthedocs.io/en/0.10.1-rc/api.html#c.z_get.

I'm wondering if we should be taking ownership of the replies by relying on the `z_call()` pattern instead of `z_get()` which requires a callback that keeps ownership of the reply? This would be similar to the example [here](https://github.com/eclipse-zenoh/zenoh-c/blob/67428d72935f7576c240ed5ab7170342e3affa17/examples/z_non_blocking_get.c#L50-L68).

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -97,11 +99,7 @@ void sub_data_handler(const z_sample_t * sample, void * sub_data);
 
 struct saved_msg_data
 {
-  explicit saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16])
-  : payload(p), recv_timestamp(recv_ts)
-  {
-    memcpy(publisher_gid, pub_gid, 16);
-  }
+  explicit saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16]);
```
Yadunund, line 103: From a closer look, I don't think we need this struct. We can directly store the `z_owned_reply_t` in `rmw_client_data_t`?

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,78 @@ struct rmw_subscription_data_t
...
+  const char * service_name;
+
+  // TODO(francocipollone): Remove this. For some reason if I remove this(not being even used) it
+  //                        ends up panicking when calling the service. Something is missing.
+  z_owned_reply_channel_t zn_reply_channel;
+  z_owned_closure_reply_t zn_closure_reply;
+
+
+  std::mutex message_mutex;
+  std::unique_ptr<saved_msg_data> message;
```
Yadunund: See comments in .cpp on relying on `z_call()` pattern and taking ownership of the reply instead of relying on a callback. This should also be a vector to support receiving replies from multiple services with the same endpoint?

```suggestion
  std::vector<z_owned_reply_t> message;
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -381,6 +381,33 @@ static const rosidl_message_type_support_t * find_type_support(
   return type_support;
 }
 
+//==============================================================================
+static const rosidl_service_type_support_t * find_service_type_support(
```
Yadunund: Nit: We've been relying on [anonymous namespaces](https://stackoverflow.com/questions/25724787/static-functions-outside-classes) in most places in this package instead of `static` keyword so let's adopt that here too for consistency.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1628,25 +1655,243 @@ rmw_return_loaned_message_from_subscription(
...
+    return nullptr;
+  }
+
+  if (validation_result != RMW_TOPIC_VALID && !qos_profile->avoid_ros_namespace_conventions) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("service name is malformed: %s", service_name);
+    return nullptr;
+  }
+
+  // client data
+  rmw_client_t * rmw_client = static_cast<rmw_client_t *>(allocator->allocate(
```
Yadunund: Follow pattern with `zero_allocate()` as seen in other places https://github.com/ros2/rmw_zenoh/blob/d734bc77b976e6eb482aa9072a5c1cebe7d48a28/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L1150-L1152

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1657,10 +1902,91 @@ rmw_send_request(
...
+    return RMW_RET_ERROR;
+  }
+
+  size_t data_length = ser.getSerializedDataLength();
+
+  // TODO(francocipollone): Do I really need the sequency number here?
+  *sequence_id = 0;
+
+  // Send request
+  z_get_options_t opts = z_get_options_default();
```
Yadunund, line 2007: I wonder if we should customize the options for ROS 2. Eg. the [z_query_target_t](https://zenoh-c.readthedocs.io/en/0.10.1-rc/api.html#c.z_query_target_t) should be set to `ALL`?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1657,10 +1902,91 @@ rmw_send_request(
...
+    });
+  if (!z_keyexpr_check(&keyexpr)) {
+    RMW_SET_ERROR_MSG("unable to create zenoh keyexpr.");
+    return RMW_RET_ERROR;
+  }
+
+  client_data->service_name = client->service_name;
+  client_data->zn_closure_reply = z_closure(client_data_handler, nullptr, client_data);
+
+  z_get(z_loan(context_impl->session), z_loan(keyexpr), "", &client_data->zn_closure_reply, &opts);
```
Yadunund: Again, should we follow pattern in [z_non_blocking_get](https://github.com/eclipse-zenoh/zenoh-c/blob/67428d72935f7576c240ed5ab7170342e3affa17/examples/z_non_blocking_get.c#L51-L68) example which would allow us to get ownership of the reply? We can then store the replies directly in `rmf_client_data_t->messages`

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1672,11 +1998,67 @@ rmw_take_response(
...
+  {
+    std::lock_guard<std::mutex> lock(client_data->message_mutex);
+    if (client_data->message == nullptr) {
+      // TODO(francocipollone): Verify behavior.
+      RCUTILS_LOG_ERROR_NAMED("rmw_zenoh_cpp", "[rmw_take_response] Response message is empty");
+      return RMW_RET_ERROR;
+    }
+
+    msg_data = std::move(client_data->message);
+    client_data->message.release();
```
Yadunund: While this releases the unique_ptr, does it free the sample we took ownership of using `zc_sample_payload_rcinc()`? Anyways if we switch to storing the owned replies directly, we will not have this problem.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1739,33 +2128,236 @@ rmw_create_service(
...
+    return nullptr);
+  if (!z_check(context_impl->session)) {
+    RMW_SET_ERROR_MSG("zenoh session is invalid");
+    return nullptr;
+  }
+
+  // SERVICE DATA ==============================================================
+  rcutils_allocator_t * allocator = &node->context->options.allocator;
+
+  rmw_service_t * rmw_service = static_cast<rmw_service_t *>(allocator->allocate(
```
Yadunund: See comment on `zero_allocate()` above.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1739,33 +2128,236 @@ rmw_create_service(
...
+  if (!z_keyexpr_check(&keyexpr)) {
+    RMW_SET_ERROR_MSG("unable to create zenoh keyexpr.");
+    return nullptr;
+  }
+
+  RCUTILS_LOG_INFO_NAMED(
+    "rmw_zenoh_cpp", "[rmw_create_service] keyexpr: %s",
+    z_loan(z_keyexpr_to_string(z_loan(keyexpr))));
+
+  z_owned_closure_query_t callback = z_closure(service_data_handler, nullptr, service_data);
```
Yadunund, line 2374: I wonder if we should rely on [fifo channels](https://github.com/eclipse-zenoh/zenoh-c/blob/67428d72935f7576c240ed5ab7170342e3affa17/examples/z_queryable_with_channels.c#L63-L65) to set up the queryable?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1778,11 +2370,83 @@ rmw_take_request(
...
+      deser,
+      ros_request,
+      service_data->request_type_support_impl))
+  {
+    RMW_SET_ERROR_MSG("could not deserialize ROS message");
+    return RMW_RET_ERROR;
+  }
+
+  RCUTILS_LOG_INFO_NAMED("rmw_zenoh_cpp", "[rmw_take_request] deserialized message");
+
```
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1778,11 +2370,83 @@ rmw_take_request(
...
+
+  // Fill in the request header.
+  request_header->request_id.sequence_number = query_id;
+
```
Yadunund, line 2544:
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: I think this is fine!

Yadunund (changes requested):
Hi Franco,

Here's my second round of review. Most of my suggestion are around how we parse and store the replies from zenoh. On the client side, instead of relying on a zenoh callback that retains ownership of the reply, I think we could adopt the pattern in the non-blocking z_get [example](https://github.com/eclipse-zenoh/zenoh-c/blob/67428d72935f7576c240ed5ab7170342e3affa17/examples/z_non_blocking_get.c#L48-L68) and directly store the `z_owned_reply_t`s within `rmw_client_data_t`.

francocipollone:
> I think we could adopt the pattern in the non-blocking z_get [example](https://github.com/eclipse-zenoh/zenoh-c/blob/67428d72935f7576c240ed5ab7170342e3affa17/examples/z_non_blocking_get.c#L48-L68) and directly store the z_owned_reply_ts within rmw_client_data_t.

Yes, I was actually planning to try that approach as I am still dealing with the double-response issue.
Regarding the channel vs closure approach. The recommendations were on using closures: "In general, if your system is designed such that only one response should be sent to any query, and unless you want to go for an actor model, I'd suggest sticking to closures" from [discord chat](https://discord.com/channels/914168414178779197/1187424012364685393/1187430607437967400). So I was hesitant to change the approach.

I will switch to using that and see how it goes.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Every time we receive a `z_get` command the callback is executed and a `z_owned_query` object is added to the map. 

We need this query object
 - For processing the input message (in `rmw_take_request`)
 - For sending a reply (in `rmw_send_response`)
 
 We cannot erase them from the map when it is "taken" given that we need that very same object to execute the reply. However, we do need to know that it was "processed/taken".

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
francocipollone: Right! Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Originally I used a custom struct `saved_queryable_data` due to I was expecting more data to be added but it wasn't necessary. I removed the struct.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
francocipollone: Thanks for the suggestions. I removed the struct.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
francocipollone: As pub_gid is an array coming from a function's argument, it decays into a pointer to the first element, at which point knowledge of its size is lost. So we cannot get the size of `pub_gid` unless we pass that as parameter.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: This is the service name in the ros world. Using `keyexpr` might be ambiguous.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
francocipollone: Done.

francocipollone:
Hey @Yadunund . Regarding the use of the fifo channels and replicate the behavior in the [example](https://github.com/eclipse-zenoh/zenoh-c/blob/67428d72935f7576c240ed5ab7170342e3affa17/examples/z_non_blocking_get.c#L50-L68)

I've been taking a look at this pattern and finding a way to bring it in the implementation. My concern here is that this pattern creates a necessity to continuously checking whether a request has been made (on the service side) and the reply is made (on the client side):

service side:
```cpp
for (z_call(channel.recv, &oquery); z_check(oquery); z_call(channel.recv, &oquery)) {
```
(from [z_queryable_with_channels.c](https://github.com/eclipse-zenoh/zenoh-c/blob/a2209b729cdab99e9d1e2c7722693bc7907d1d59/examples/z_queryable_with_channels.c))
client side:
```cpp
for (bool call_success = z_call(channel.recv, &reply); !call_success || z_check(reply);
```
(from [z_non_blocking_get.c](https://github.com/eclipse-zenoh/zenoh-c/blob/3d43cceb375d04e40626841b430d578389cfce3f/examples/z_non_blocking_get.c))

So for example, after creating the service, I'd need to kick off a thread that only has that check in a loop. So once I get the result I can unblock the condition variable that is holding on to that rmw wait-set event and then `rmw_take_request` will be triggered.

That's why the callback pattern fits better as it is analogous to the use of publisher and subscriber for the imeplentation of the service as it is for other implementations.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
Yadunund: Fair enough.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
Yadunund: I see thanks for explaining that. In that case, let's pass that as a parameter? Just to minimize magic numbers.

Yadunund:
I see thanks for explaining those differences. I misunderstood that we are not able to take ownership of the reply with the callback approach. It makes more sense to stick to this approach in that case.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -131,4 +129,66 @@ struct rmw_subscription_data_t
...
+  const void * response_type_support_impl;
+  const char * typesupport_identifier;
+  RequestTypeSupport * request_type_support;
+  ResponseTypeSupport * response_type_support;
+
+  rmw_context_t * context;
+
+  // Map to store the query id and the query.
+  // The query handler is saved as it is needed to answer the query later on.
+  std::unordered_map<unsigned int, std::unique_ptr<z_owned_query_t>> id_query_map;
```
Yadunund: Another suggestion to consider: It might be simpler to store the owned query directly. Once we're done "taking" the reply, we can just drop it without needing to reset the unique ptr AND dropping the value it was managing. 

```suggestion
  std::unordered_map<unsigned int, z_owned_query_t> id_query_map;
```

Yadunund:
I ran into a couple issues when running some examples. Sepcifically,
The client node panicked when taking a response
```bash
[INFO] [1703683133.857969473] [rmw_zenoh_cpp]: [client_data_handler] triggered for /add_two_ints
thread 'async-std/runtime' panicked at library/core/src/panicking.rs:136:5:
panic in a function that cannot unwind
stack backtrace:
   0:     0x7f1fdf3ed740 - std::backtrace_rs::backtrace::libunwind::trace::h67a838aed1f4d6ec
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x7f1fdf3ed740 - std::backtrace_rs::backtrace::trace_unsynchronized::h1d1786bb1962baf8
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f1fdf3ed740 - std::sys_common::backtrace::_print_fmt::h5a0b1f807a002d23
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x7f1fdf3ed740 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hf84ab6ad0b91784c
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7f1fdf1c632c - core::fmt::rt::Argument::fmt::h28f463bd1fdabed5
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/fmt/rt.rs:138:9
   5:     0x7f1fdf1c632c - core::fmt::write::ha37c23b175e921b3
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/fmt/mod.rs:1114:21
   6:     0x7f1fdf3bd2dd - std::io::Write::write_fmt::haa1b000741bcbbe1
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/io/mod.rs:1763:15
   7:     0x7f1fdf3eee4e - std::sys_common::backtrace::_print::h1ff1030b04dfb157
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/sys_common/backtrace.rs:47:5
   8:     0x7f1fdf3eee4e - std::sys_common::backtrace::print::hb982056c6f29541c
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/sys_common/backtrace.rs:34:9
   9:     0x7f1fdf3eea30 - std::panicking::default_hook::{{closure}}::h11f92f82c62fbd68
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:272:22
  10:     0x7f1fdf3efa1a - std::panicking::default_hook::hb8810fe276772c66
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:292:9
  11:     0x7f1fdf3efa1a - std::panicking::rust_panic_with_hook::hd2f0efd2fec86cb0
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:731:13
  12:     0x7f1fdf3ef4f8 - std::panicking::begin_panic_handler::{{closure}}::h3651b7fc4f61d784
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:601:13
  13:     0x7f1fdf3ef486 - std::sys_common::backtrace::__rust_end_short_backtrace::hbc468e4b98c7ae04
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/sys_common/backtrace.rs:170:18
  14:     0x7f1fdf3ef471 - rust_begin_unwind
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:597:5
  15:     0x7f1fdf08b077 - core::panicking::panic_nounwind_fmt::h802148bc84c1e34d
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:106:14
  16:     0x7f1fdf08b0e8 - core::panicking::panic_nounwind::h13bc2cb514e17671
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:136:5
  17:     0x7f1fdf08b092 - core::panicking::panic_cannot_unwind::hf0e1d75c7a4ef91a
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:213:5
  18:     0x7f1fdf3f0784 - std::sys::unix::thread::Thread::new::thread_start::hd28b46dbf5673d17
                               at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/sys/unix/thread.rs:102:9
  19:     0x7f1fe0a94ac3 - start_thread
                               at ./nptl/pthread_create.c:442:8
  20:     0x7f1fe0b26660 - clone3
                               at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
  21:                0x0 - <unknown>
thread caused non-unwinding panic. aborting.
```

Also seeing undefined symbols when terminating the server node
```bash
^C[INFO] [1703683208.856754851] [rclcpp]: signal_handler(signum=2)
/opt/ros/iron/lib/demo_nodes_cpp/add_two_ints_server: symbol lookup error: /home/yadunund/ws_rmw/install/rmw_zenoh_cpp/lib/librmw_zenoh_cpp.so: undefined symbol: _Z6z_dropISt10unique_ptrI15z_owned_query_tSt14default_deleteIS1_EEEN15zenoh_drop_typeIT_E4typeEPS6_
[ros2run]: Process exited with failure 127
```

I've opened #88 to fix these problems.

francocipollone:
> I ran into a couple issues when running some examples. Sepcifically, The client node panicked when taking a response

Just in case, have you deleted `zenohc_vendor` builds? and build again? Throughout the last weeks, some fixes to the zenoh-c api were pushed.

> I've opened https://github.com/ros2/rmw_zenoh/pull/88 to fix these problems.

Great I will take a look

Yadunund:
> Just in case, have you deleted zenohc_vendor builds? and build again? Throughout the last weeks, some fixes to the zenoh-c api were pushed.

Yep this is on a fresh build from a few hours ago.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Done.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: I created a function with firm:
```cpp
rmw_ret_t serialize_ros_message(rcutils_allocator_t * allocator, TypeSupport * type_support, const void * type_support_impl, const void * ros_message, char * msg_bytes, size_t * data_length)
```
Click on details to see the implementation.
<details>

```cpp
// Serialize a ROS message into a buffer.
// Typically used by rmw_send_request (client side) and rmw_send_response (server side) to
// serialize the message before sending it over the wire.
rmw_ret_t serialize_ros_message(rcutils_allocator_t * allocator, TypeSupport * type_support, const void * type_support_impl, const void * ros_message, char * msg_bytes, size_t * data_length)
{
  // Assertions
  RMW_CHECK_FOR_NULL_WITH_MSG(
    allocator, "allocator is null",
    return RMW_RET_INVALID_ARGUMENT);
  RMW_CHECK_FOR_NULL_WITH_MSG(
    type_support, "type support handle is null",
    return RMW_RET_INVALID_ARGUMENT);
  RMW_CHECK_FOR_NULL_WITH_MSG(
    type_support_impl, "type support impl is null",
    return RMW_RET_INVALID_ARGUMENT);
  RMW_CHECK_FOR_NULL_WITH_MSG(
    ros_message, "ros message is null",
    return RMW_RET_INVALID_ARGUMENT);
  RMW_CHECK_FOR_NULL_WITH_MSG(
    data_length, "data length is null",
    return RMW_RET_INVALID_ARGUMENT);

  // Serialize data.
  size_t max_data_length = type_support->get_estimated_serialized_size(
    ros_message,
    type_support_impl);

  msg_bytes = static_cast<char *>(allocator->allocate(
      max_data_length,
      allocator->state));
  if (!msg_bytes) {
    RMW_SET_ERROR_MSG("failed allocate request message bytes");
    allocator->deallocate(msg_bytes, allocator->state);
    return RMW_RET_ERROR;
  }

  // Object that manages the raw buffer
  eprosima::fastcdr::FastBuffer fastbuffer(msg_bytes, max_data_length);
  // Object that serializes the data
  eprosima::fastcdr::Cdr ser(
    fastbuffer,
    eprosima::fastcdr::Cdr::DEFAULT_ENDIAN,
    eprosima::fastcdr::Cdr::DDS_CDR);
  if (!type_support->serialize_ros_message(
      ros_message,
      ser,
      type_support_impl))
  {
    allocator->deallocate(msg_bytes, allocator->state);
    return RMW_RET_ERROR;
  }
  *data_length = ser.getSerializedDataLength();

  return RMW_RET_OK;
}
```
</details>

And I used it in the `rmw_send_request` and `rmw_send_response` functions.

I got an odd behaviour with the `msg_bytes` pointer. **After the method is returned the `msg_bytes` pointer becomes null** even when the ownership isn't from that function. 

According to the FastBuffer docstring:
```cpp
    /*!
     * @brief This constructor assigns the user's stream of bytes to the eprosima::fastcdr::FastBuffers object.
     * The user's stream will be used to serialize.
     *
     * @param buffer The user's buffer that will be used. This buffer is not deallocated in the object's destruction. Cannot be NULL.
     * @param bufferSize The length of user's buffer.
     */
    FastBuffer(
            char* const buffer,
            const size_t bufferSize);
```

So the buffer pointer shouldn't be destructed after going off scope.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1630,145 +1662,695 @@ rmw_return_loaned_message_from_subscription(
...
+      client_data->request_type_support_impl))
+  {
+    allocator->deallocate(request_bytes, allocator->state);
+    return RMW_RET_ERROR;
+  }
+
+  size_t data_length = ser.getSerializedDataLength();
+
+  // TODO(francocipollone): Do I really need the sequency number here?
+  *sequence_id = 0;
```
francocipollone: Hey @clalancette @Yadunund  Do you have more insight about how the sequence id is being used? I imagine this comes from pub/sub implementation for the services which might not be necessary in our case.

clalancette: So I did some research about sequence_id in ROS 2.

First of all, it is defined in https://github.com/ros2/rmw/blob/9b3d9d0e3021b7a6e75d8886e3e061a53c36c789/rmw/include/rmw/rmw.h#L2039-L2045 .

Looking through the implementations, it ends up being used mostly by the RMW layer.  But there, it is always transmitted as part of the service call, so that the service server knows the sequence number hte client chose.  It is then also exposed to the higher layers via `rmw_service_info_t` and `rmw_client_info_t`.  Via that mechanism, it is used in `rcl` when doing service introspection.  I can see this being useful, even in the rmw_zenoh context, to be able to match clients and services together.

Unfortunately, how to generate and store the sequence number is under-specified.  rmw_cyclonedds, for instance, just has a global variable that gets incremented every time a service request is made.  I'm going to suggest that we do something similar here to start with; eventually maybe we want to consider tying this to the node or the session instead of having it global, but we can do it later.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -70,3 +77,89 @@ void sub_data_handler(
...
+
+//==============================================================================
+void service_data_handler(const z_query_t * query, void * data)
+{
+  RCUTILS_LOG_INFO_NAMED(
+    "rmw_zenoh_cpp",
+    "[service_data_handler] triggered"
+  );
+  z_owned_str_t keystr = z_keyexpr_to_string(z_query_keyexpr(query));
+
```
clalancette, line 96: I'm going to suggest that we use an` rcpputils::scope_exit` to drop the `keystr`.  That way, we can remove the explicit `z_drop` from all code paths and just rely on the scope_exit (which is how we've generally done it elsewhere in `rmw_zenoh`).

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -48,6 +48,102 @@
...
+    &topic_name[start_offset], end_offset - start_offset, *allocator);
+  z_owned_keyexpr_t keyexpr = z_keyexpr_join(
+    z_keyexpr(domain_ss.str().c_str()), z_keyexpr(stripped_topic_name));
+  allocator->deallocate(stripped_topic_name, allocator->state);
+
+  return keyexpr;
+}
+
+//==============================================================================
+const rosidl_message_type_support_t * find_type_support(
```
clalancette: I'm going to suggest we rename this to `find_msg_type_support`, just to be clear about what the function does.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -48,6 +48,102 @@
...
+    if (topic_name[end_offset - 1] == '/') {
+      // Strip the trailing '/'
+      end_offset -= 1;
+    }
+  }
+
+  std::stringstream domain_ss;
+  domain_ss << domain_id;
+  char * stripped_topic_name = rcutils_strndup(
+    &topic_name[start_offset], end_offset - start_offset, *allocator);
```
clalancette, line 85: I know you didn't add this, but we should probably at least put a TODO comment in here for error checking on the `rcutils_strndup`.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1630,145 +1662,695 @@ rmw_return_loaned_message_from_subscription(
...
+
+  rmw_client_data_t * client_data = static_cast<rmw_client_data_t *>(client->data);
+  RMW_CHECK_FOR_NULL_WITH_MSG(
+    client->data, "Unable to retrieve client_data from client.", RMW_RET_INVALID_ARGUMENT);
+
+  z_owned_reply_t * latest_reply = nullptr;
+
+  std::lock_guard<std::mutex> lock(client_data->message_mutex);
+  if (client_data->replies.empty()) {
+    // TODO(francocipollone): Verify behavior.
```
clalancette: Can you explain a little bit more what you mean by "Verify behavior"?  Which behavior in particular do we still need to look at?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1630,145 +1662,695 @@ rmw_return_loaned_message_from_subscription(
...
+    RCUTILS_LOG_ERROR_NAMED("rmw_zenoh_cpp", "[rmw_take_response] Response message is empty");
+    return RMW_RET_ERROR;
+  }
+  latest_reply = &client_data->replies.back();
```
clalancette: This is always responding to the latest_reply.  Are we sure that that is always going to be for *this* client?  That is, if we had multiple clients all sending requests at the same time, how do we know that we've taken the correct one?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1630,145 +1662,695 @@ rmw_return_loaned_message_from_subscription(
...
-  rmw_service_free(service);
+
+  *taken = true;
+
+  for (z_owned_reply_t & reply : client_data->replies) {
+    z_reply_drop(&reply);
+  }
+  client_data->replies.clear();
+
+  // TODO(francocipollone): Verify request_header information.
```
clalancette: As in my other response, we should probably send the sequence_number of the wire and then fill it out here.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1630,145 +1662,695 @@ rmw_return_loaned_message_from_subscription(
...
+  // Get the RMW type support.
+  const rosidl_service_type_support_t * type_support = find_service_type_support(type_supports);
+  if (type_support == nullptr) {
+    // error was already set by find_type_support
+    return nullptr;
+  }
+
+  // TODO(francocipollone): Verify if this is the right way to get the
+  //                        type support as the architecture for the service here
+  //                        is different to the one used in DDS (with fastcdr).
```
clalancette: Given that service seem to be working, is there any further work to do here?  Or can we remove this TODO?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1796,10 +2438,85 @@ rmw_send_response(
...
+  if (query_it == service_data->id_query_map.end()) {
+    RMW_SET_ERROR_MSG("Unable to find taken request. Report this bug.");
+    return RMW_RET_ERROR;
+  }
+  const z_query_t z_loaned_query = z_query_loan(&query_it->second);
+  z_query_reply_options_t options = z_query_reply_options_default();
+  options.encoding = z_encoding(Z_ENCODING_PREFIX_EMPTY, NULL);
+  z_query_reply(
+    &z_loaned_query, z_loan(service_data->keyexpr), reinterpret_cast<const uint8_t *>(
+      response_bytes), data_length + meta_length, &options);
```
clalancette: Do we need to free `response_bytes` after we're done?  Or is that handled by the destructor for `FastBuffer`?

clalancette (changes requested):
Overall, this is a fantastic implementation.  I've left a few thoughts and suggestions inline.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
JEnoch: If I may, I have 2 suggestions then:
1. make this sequence number optional in the request payload (e.g. generating a local one in rmw_zenoh on server side if not present). Thus a zenoh-pico client or server on a microcontroller is not required to deal with such sequence number.
2. add this sequence number as an attachment to the Zenoh query and reply, rather than piggyback it with the request and reply payloads. This will simplify the serialization and probably avoid a buffer copy.
  You can see an example of attachments usage with pub/sub in [zenoh-c/examples](https://github.com/eclipse-zenoh/zenoh-c/tree/0.10.1-rc/examples). The usage with query/reply is similar (an `attachment` in options struct for query and reply)

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
clalancette: I ended up fixing this in [177451a](https://github.com/ros2/rmw_zenoh/pull/86/commits/177451a0a200e3327afe9029b5cc405731f527f5).

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: I've now done this in [9d9f6f6](https://github.com/ros2/rmw_zenoh/pull/86/commits/9d9f6f6597bf676cc7aaedeeb02fe23f63f43da4).

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: I added error checking in [ce6af2b](https://github.com/ros2/rmw_zenoh/pull/86/commits/ce6af2b52d67679705dbbb21b8216f2ae77b5b76).

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: It turns out that we do need to always free response_bytes.  I've added a scope_exit for this in [7773aee](https://github.com/ros2/rmw_zenoh/pull/86/commits/7773aee1b9dbbad38c897527c25af04041c181b0)

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Yes we can remove it! :smiley:

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: > if we had multiple clients all sending requests at the same time, how do we know that we've taken the correct one?

This is solved at the server side, as we are using queryables, from the server side we store the query and answer to it using the same channel.

This latest reply and the vector replies are here to support the situation when you have multiple servers with the same keyexpr. In this part you might get multiple replies (one from each server). We are not tackling this situation in this PR, we simply return the latest reply.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
francocipollone: Yeah thanks for asking,
Because of the way that `rmw_send_request`, `rmw_take_request`, `rmw_send_response` and `rmw_take_response` are used code shouldn't reach this location, correct? I guess we should log an error or throw if we get into this if-clause

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: Fixed in [8054227](https://github.com/ros2/rmw_zenoh/pull/86/commits/805422742249132eb990e660fe1a234e85f05d95) .

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: OK, my latest pushes implement this using attachments, and it seems to work.  There are a couple of other pieces of metadata that we should fill in as well, but we can address those separately.  I'll open an issue about it.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: Yeah, I think so.  I'm just going to remove this TODO, as I think what we have there is OK for now.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: This was done using attachments.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1781,11 +2462,91 @@ rmw_take_request(
...
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+
+  RMW_CHECK_FOR_NULL_WITH_MSG(
+    service->service_name, "service has no service name", RMW_RET_INVALID_ARGUMENT);
+
+  rmw_service_data_t * service_data = static_cast<rmw_service_data_t *>(service->data);
+  RMW_CHECK_FOR_NULL_WITH_MSG(
+    service->data, "Unable to retrieve service_data from service", RMW_RET_INVALID_ARGUMENT);
+
+  z_owned_query_t query;
```
Yadunund, line 2486: It looks like we want to "move" the `z_owned_query_t` from `query_queue.front()` to `sequence_to_query_map` when some conditions are valid. I'm a bit concerned about creating this `z_owend_query_t object` and whether we need to drop it if the function returns before emplacing into the map?
Wondering if we should instantiate this as `z_owned_query_t * query` so that we won't have to drop anything and we can just `std::move(*query)` when emplacing into the map?

Also would it be better to emplace into the map AFTER serializing the payload? Wondering if the loaned query created below this will have a dangling reference after moving the `z_owned_query_t` from queue to map.

Yadunund (approved):
Thanks for getting the sequence numbers to be generated and managed correctly! I have one clarification comment but it's not a blocker so will go ahead and merge this in.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Alternatively, the map and queue can store `unique_ptr<z_owned_query_t>` to make sure we're transferring ownership?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2055,13 +2075,35 @@ rmw_take_response(
 
   *taken = true;
 
-  for (z_owned_reply_t & reply : client_data->replies) {
```
Yadunund: @clalancette looking back at this diff, I wonder if we should keep the original. If there is more than one server present, replies from other servers will still be stored in `client_data->replies`. We need to drop other replies to let the zenoh queryable that we're done with the reply?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2341,8 +2383,8 @@ rmw_destroy_service(rmw_node_t * node, rmw_service_t * service)
   // CLEANUP ================================================================
   z_drop(z_move(service_data->keyexpr));
   z_drop(z_move(service_data->qable));
-  for (auto & id_query : service_data->id_query_map) {
-    z_drop(z_move(id_query.second));
+  for (z_owned_query_t & query : service_data->query_queue) {
```
Yadunund, line 2439: @clalancette since we move ownership of taken requests from `query_queue` to `sequence_id_map`, should we also iterate over the map and properly drop the `z_owned_query_t`?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: > I'm a bit concerned about creating this `z_owend_query_t object` and whether we need to drop it if the function returns before emplacing into the map?

So I don't think this is actually a worry.  That's because we only `pop_front` the request if we totally succeeded in this function.  Thus, if we were to fail somewhere else in here, we would just leave it at the front of the replies.  That means we would reprocess it next time; that may or may not be what we want to do, but it is what we do now.

All of that said, we can make this a little clearer by using unique_ptr here, so I've gone ahead and done that (I'll push some fixes to your other PR).  Also I realized that the locking in here was completely bogus, so I'm going to fix that up as well.

Issue #84, opened 2023-12-13
Graph introspection incomplete when nodes with duplicate names present


Yadunund:
The current [GraphCache implementation](https://github.com/ros2/rmw_zenoh/blob/d734bc77b976e6eb482aa9072a5c1cebe7d48a28/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L134-L137) assumes nodes within a namespace will have unique names. But this is not necessarily true for ROS 2 nodes. Hence, with two `talker` nodes running, commands like `ros2 node list` will return only reveal a single `/talker` node.

> Note: With DDS, GIDs are available for entities which are globally unique. With zenoh, each session will have a unique  z_session_id.  

Potential solutions:

1. Update liveliness token keyexprs to also include the `z_session_id` of the entity. Then update [NodeMap](https://github.com/ros2/rmw_zenoh/blob/d734bc77b976e6eb482aa9072a5c1cebe7d48a28/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L134) to be an [std::multimap](https://en.cppreference.com/w/cpp/container/multimap) with a custom comparator that sorts based on the `z_session_id`.

Yadunund:
Closed by https://github.com/ros2/rmw_zenoh/pull/87

PR #87, opened 2023-12-21
Handle duplicate node names by including zenoh session id in liveliness tokens

Yadunund:
This PR addresses #84.

It so does so by including the zenoh session id in the liveliness token.
The token is updated to:
`<ADMIN_SPACE>/<domainid>/<id>/<entity>/<namespace>/<nodename>` where `<id>` is meant to be a unique ID to identify this entity. However, until the zenoh API exposes the globally unique IDs for each participant/entity, this id is currently set to the zenoh session's id with elements concatenated into a string using '.' as separator.

The `NodeMap` data structure is updated to an `std::multimap` to allow storing multiple nodes with the same name in the same namespace. Codebase is updated accoridngly.

Now after running two instances of `/talker`, running the command `ros2 node list` should correctly list two separate `/talker` nodes.
Note: Make sure to kill all ros2 nodes and daemons before trying this.

```bash
yadunund@ubuntu-22-04:~/ws_rmw$ ros2 node list
/talker
/talker
```

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -207,13 +207,12 @@ void GraphCache::parse_put(const std::string & keyexpr)
     NodeMap::iterator insertion_it =
       ns_it->second.insert(std::make_pair(entity.node_name(), make_graph_node(entity)));
     if (insertion_it != ns_it->second.end()) {
-      RCUTILS_LOG_WARN_NAMED(
+      RCUTILS_LOG_INFO_NAMED(
         "rmw_zenoh_cpp",
         "Added a new node /%s with id %s to an existing namespace %s in the graph.",
         entity.node_name().c_str(),
         entity.id().c_str(),
         entity.node_namespace().c_str());
-      return;
     } else {
       RCUTILS_LOG_ERROR_NAMED(
         "rmw_zenoh_cpp",
@@ -221,12 +220,11 @@ void GraphCache::parse_put(const std::string & keyexpr)
         entity.node_name().c_str(),
         entity.id().c_str(),
         entity.node_namespace().c_str());
-      return;
     }
-  } else {
-    // The entity represents a node that already exists in the graph.
-    // Update topic info if required below.
+    return;
   }
+  // Otherwise, the entity represents a node that already exists in the graph.
+  // Update topic info if required below.
 
   // Handles additions to an existing node in the graph.
   if (entity.type() == EntityType::Node) {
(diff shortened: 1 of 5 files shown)


rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -187,29 +189,55 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+        "Unable to add a new node /%s with id %s an existing namespace %s in the graph. Report this bug.",
+        entity.node_name().c_str(),
+        entity.id().c_str(),
+        entity.node_namespace().c_str());
+      return;
+    }
+  } else {
+    // The entity represents a node that already exists in the graph.
+    // Update topic info if required below.
   }
```
clalancette: I think we can simplify this whole block a bit:

```suggestion
    if (insertion_it != ns_it->second.end()) {
      RCUTILS_LOG_INFO_NAMED(
        "rmw_zenoh_cpp",
        "Added a new node /%s with id %s to an existing namespace %s in the graph.",
        entity.node_name().c_str(),
        entity.id().c_str(),
        entity.node_namespace().c_str());
    } else {
      RCUTILS_LOG_ERROR_NAMED(
        "rmw_zenoh_cpp",
        "Unable to add a new node /%s with id %s an existing namespace %s in the graph. Report this bug.",
        entity.node_name().c_str(),
        entity.id().c_str(),
        entity.node_namespace().c_str());
    }
    return;
  }
  // Otherwise, the entity represents a node that already exists in the graph.
  // Update topic info if required below.
```

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -83,6 +83,18 @@ static const std::unordered_map<std::string, EntityType> str_to_entity = {
   {CLI_STR, EntityType::Client}
 };
 
+std::string convert(z_id_t id)
```
clalancette: How about:

```suggestion
std::string zid_to_str(z_id_t id)
```

clalancette (changes requested):
I've left two minor things to fix, this otherwise looks good to me.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/87/commits/6d3acd38637deec614692d8c1f0c9815b7aca6be

clalancette (approved):

PR #88, opened 2023-12-27
Cleanup services implementation

Yadunund:
This PR fixes
* zenoh panicing when client callback executes
* symbol error when destroying server
* memory allocation and leaks


francocipollone:
Thanks! :rocket:

PR #90, opened 2023-12-28
Update GraphCache for services & clients along with remaining graph introspection methods

Yadunund:
~This PR builds on top of #86  but with changes from https://github.com/ros2/rmw_zenoh/pull/87. Reviewing might be easier if we merge #87 in and rebase #86.~

Changes introduced:
* Liveliness tokens are put/deleted for service and client entities and GraphCache is updated with this information.
* Implement all remaining graph introspection functions.
* Fixes liveliness tokens when node or topic names have `/` in the middle. Eg `/add_two_ints_server/get_type_description`. Previously such names would get appended to the liveliness token without any substitution which could cause the parsing and reconstruction of `liveliness::Entity` to have incorrect values for types and qos. Now a `SLASH_REPLACEMENT` char is introduced and such names and mangled and demangled to correctly transmit and reconstruct graph information.

To Test:

```bash
# terminal 1
  ros2 run demo_nodes_cpp add_two_ints_server
```

```bash
# terminal 2
ros2 service list

# result
/add_two_ints_server/set_parameters
/add_two_ints_server/set_parameters_atomically
/add_two_ints
/add_two_ints_server/get_type_description
/add_two_ints_server/get_parameter_types
/add_two_ints_server/list_parameters
/add_two_ints_server/get_parameters
/add_two_ints_server/describe_parameters

```

```bash
ros2 node info /add_two_ints

# result 
/add_two_ints_server
  Subscribers:
    /parameter_events: rcl_interfaces/msg/ParameterEvent
  Publishers:
    /rosout: rcl_interfaces/msg/Log
    /parameter_events: rcl_interfaces/msg/ParameterEvent
  Service Servers:
    /add_two_ints_server/get_type_description: type_description_interfaces/srv/GetTypeDescription
    /add_two_ints: example_interfaces/srv/AddTwoInts
    /add_two_ints_server/get_parameter_types: rcl_interfaces/srv/GetParameterTypes
    /add_two_ints_server/list_parameters: rcl_interfaces/srv/ListParameters
    /add_two_ints_server/get_parameters: rcl_interfaces/srv/GetParameters
    /add_two_ints_server/describe_parameters: rcl_interfaces/srv/DescribeParameters
    /add_two_ints_server/set_parameters_atomically: rcl_interfaces/srv/SetParametersAtomically
    /add_two_ints_server/set_parameters: rcl_interfaces/srv/SetParameters
  Service Clients:

  Action Servers:

  Action Clients:
  ```
  
  ## Known issue (currently being investigated)
  ~Starting 3 or more nodes causes the liveliness get query at node initialization to hang indefinitely.~
  * Fixed by https://github.com/ros2/rmw_zenoh/pull/90/commits/689d10231665e6e13277b0ab6b189cfb98158997

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -963,7 +963,6 @@ rmw_ret_t GraphCache::get_entities_info_by_topic(
   return RMW_RET_OK;
 }
 
-
 ///=============================================================================
 rmw_ret_t GraphCache::service_server_is_available(
   const char * service_name,
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -963,3 +963,23 @@ rmw_ret_t GraphCache::get_entities_info_by_topic(
   cleanup_endpoints_info.cancel();
   return RMW_RET_OK;
 }
+
+
+///=============================================================================
+rmw_ret_t GraphCache::service_server_is_available(
+  const char * service_name,
+  const char * service_type,
+  bool * is_available)
+{
+  *is_available = false;
+  std::lock_guard<std::mutex> lock(graph_mutex_);
+  GraphNode::TopicMap::iterator service_it = graph_services_.find(service_name);
+  if (service_it != graph_services_.end()){
+    GraphNode::TopicDataMap::iterator type_it = service_it->second.find(service_type);
+    if (type_it != service_it->second.end()){
+      *is_available = true;
+    }
+  }
+
+  return RMW_RET_OK;
+}
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -755,6 +755,40 @@ rmw_ret_t GraphCache::count_subscriptions(
 }
 
 ///=============================================================================
+rmw_ret_t GraphCache::count_services(
+  const char * service_name,
+  size_t * count) const
+{
+  *count = 0;
+  std::lock_guard<std::mutex> lock(graph_mutex_);
+  if (graph_services_.count(service_name) != 0) {
+    for (const std::pair<const std::string, TopicDataPtr> & it : graph_services_.at(service_name)) {
+      // Iterate through all the types and increment count.
+      *count += it.second->stats_.sub_count_;
+    }
+  }
+
+  return RMW_RET_OK;
+}
+
+///=============================================================================
+rmw_ret_t GraphCache::count_clients(
+  const char * service_name,
+  size_t * count) const
+{
+  *count = 0;
+  std::lock_guard<std::mutex> lock(graph_mutex_);
+  if (graph_services_.count(service_name) != 0) {
+    for (const std::pair<const std::string, TopicDataPtr> & it : graph_services_.at(service_name)) {
+      // Iterate through all the types and increment count.
+      *count += it.second->stats_.pub_count_;
+    }
+  }
+
+  return RMW_RET_OK;
+}
+
+///=============================================================================
 rmw_ret_t GraphCache::get_entity_names_and_types_by_node(
   liveliness::EntityType entity_type,
   rcutils_allocator_t * allocator,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: rmw_zenoh_cpp/src/detail/graph_cache.cpp [hunk 1]
```diff
@@ -84,12 +91,31 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+            entity_desc = "service";
+            return graph_node.services_;
+
+          } else {
+            entity_desc = "client";
+            return graph_node.clients_;
+          }
+        }();
+      // For the sake of reusing data structures and lookup functions, we treat publishers and clients are equivalent.
+      // Similarly, subscriptions and services are equivalent.
```
clalancette, line 115: I'm somewhat suspicious of this assertion, as they aren't the same thing.  I'm worried that this detail is going to leak out through the RMW APIs, though I haven't tested it directly yet to tell for certain.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -84,12 +91,31 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
 
       const liveliness::TopicInfo topic_info = entity.topic_info().value();
-      GraphNode::TopicMap & topic_map = entity.type() ==
-        EntityType::Publisher ? graph_node.pubs_ : graph_node.subs_;
-      const std::string entity_desc = entity.type() ==
-        EntityType::Publisher ? "publisher" : "subscription";
-      const std::size_t pub_count = entity.type() == EntityType::Publisher ? 1 : 0;
+      std::string entity_desc = "";
+      GraphNode::TopicMap & topic_map =
+        [&]() -> GraphNode::TopicMap &
```
clalancette: We really only need to capture `entity_desc`, right?

```suggestion
        [&entity_desc]() -> GraphNode::TopicMap &
```

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -257,15 +288,19 @@ void GraphCache::parse_del(const std::string & keyexpr)
 
   // Helper lambda to update graph_topics_.
   auto update_graph_topics =
-    [&](const liveliness::TopicInfo topic_info, std::size_t pub_count,
+    [&](const liveliness::TopicInfo topic_info, const EntityType entity_type, std::size_t pub_count,
```
clalancette: I know this was here before, but can we be more explicit about what needs to be captured here?

In point of fact, since this is a method we are calling ourselves, we could possibly get away with no captures and just pass everything we need in as parameters.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -133,6 +133,12 @@ class PublishToken
     const std::string & token);
 };
 
+/// Replace "/" instances with "§".
```
clalancette:
```suggestion
/// Replace "/" instances with "%".
```

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -133,6 +133,12 @@ class PublishToken
...
+std::string mangle_name(const std::string & input);
+
+/// Replace "§" instances with "/".
```
clalancette:
```suggestion
/// Replace "%" instances with "/".
```

clalancette (changes requested):
I've left a few things to think about, but overall I think this is pretty good.  I have a whole bunch of fixes to go on top of this, so hopefully we can get this cleaned up and merged soon.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/90/commits/aae76cdfd1a9d5501302116846a6f5db0634359b

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/90/commits/aae76cdfd1a9d5501302116846a6f5db0634359b

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: The alternative is to have separate data structures  and functions for clients and services. It will create unnecessary code duplications but improves readability and maintenance.

clalancette: My main concern is that this is going to leak through to things like "publisher count" and "client count".  But let's stick with what you have for now, and we can update it later if that is a problem.

clalancette (approved):

PR #91, opened 2024-01-11
Support basic QoS settings

Yadunund:
~# TODO~
- [x] Check whether PubCache can always be created or only if TRANSIENT_LOCAL qos
- [x] Set Zenoh congestion appropriately. (Default Drop except with RELIABLE and KEEP_ALL).
~- [ ] Call ze_querying_subscriber_get() when a TRANSIENT_LOCAL publisher is discovered with matching endpoints.~
- [x] Update QoS compatibility checking methods.

This PR provides support for the following QoS settings:
- Depth
- History (`KEEP_ALL` and`KEEP_LAST`)
- Durability (`VOLATILE` and `TRANSIENT_LOCAL`)
- Reliability (`RELIABLE` and `BEST_EFFORT`)

It also updates the liveliness tokens to forward serialized qos information. Graph introspection methods are also updated.
Before testing, remember to restart any running zenoh routers and kill existing ros2 daemons.

> Note: PublicationCache only works when time-stamping is enabled for peers in the zenoh configs. 

Test below where publisher has durability `VOLATILE` buyt subsciption has durability `TRANSIENT_LOCAL`. The QoS properties including History, Durability, and Reliability should be correctly populated when 

```bash
# terminal 1
ros2 topic pub /chatter std_msgs/msg/String '{data: 1}' --qos-reliability reliable --qos-durability transient_local --qos-history keep_last --qos-depth 7 --rate 0.001
```

```bash
# terminal 2
ros2 topic echo /chatter --qos-durability transient_local --qos-reliability reliable --qos-history keep_last --qos-depth 7 --no-lost-messages

# you should see the single message that was published printed in the terminal
```

**Note: The `--no-lost-messages` flag is needed or else `ros2topic` strategy will try to create a [subscription with event callback](https://github.com/ros2/ros2cli/blob/426d6e9c6a3b6d3ec7dac5f1b222e389be5d4318/ros2topic/ros2topic/verb/echo.py#L238-L257) which will fail as the methods are not implemented. The `catch` block will create another subscription without the event but the message is not received in this case.**

```bash
# terminal 3

ros2 topic info -v /chatter

Type: std_msgs/msg/String

Publisher count: 1

Node name: _ros2cli_716307
Node namespace: /
Topic type: std_msgs/msg/String
Topic type hash: INVALID
Endpoint type: PUBLISHER
GID: 00.00.00.00.00.00.00.00.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  History (Depth): KEEP_LAST (7)
  Durability: TRANSIENT_LOCAL
  Lifespan: 0 nanoseconds
  Deadline: 0 nanoseconds
  Liveliness: AUTOMATIC
  Liveliness lease duration: 0 nanoseconds

Subscription count: 1

Node name: _ros2cli_716424
Node namespace: /
Topic type: std_msgs/msg/String
Topic type hash: INVALID
Endpoint type: SUBSCRIPTION
GID: 00.00.00.00.00.00.00.00.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  History (Depth): KEEP_LAST (7)
  Durability: TRANSIENT_LOCAL
  Lifespan: 0 nanoseconds
  Deadline: 0 nanoseconds
  Liveliness: AUTOMATIC
  Liveliness lease duration: 0 nanoseconds



```


Yadunund:
With the qos methods implemented, `actions` also work**

```bash
# terminal 1
ros2 run action_tutorials_cpp fibonacci_action_server 
```

```bash
# terminal 2
ros2 run action_tutorials_cpp fibonacci_action_client 


[INFO] [1705560511.794785988] [fibonacci_action_client]: Sending goal
[INFO] [1705560511.794917049] [rmw_zenoh_cpp]: [rmw_send_request]
[INFO] [1705560511.796353630] [rmw_zenoh_cpp]: [rmw_take_response]
[INFO] [1705560511.796443000] [fibonacci_action_client]: Goal accepted by server, waiting for result
[INFO] [1705560511.796472320] [rmw_zenoh_cpp]: [rmw_send_request]
[INFO] [1705560511.796863030] [fibonacci_action_client]: Next number in sequence received: 0 1 1 
[INFO] [1705560512.797255312] [fibonacci_action_client]: Next number in sequence received: 0 1 1 2 
[INFO] [1705560513.797292703] [fibonacci_action_client]: Next number in sequence received: 0 1 1 2 3 
[INFO] [1705560514.797283484] [fibonacci_action_client]: Next number in sequence received: 0 1 1 2 3 5 
[INFO] [1705560515.797236973] [fibonacci_action_client]: Next number in sequence received: 0 1 1 2 3 5 8 
^C[INFO] [1705560516.232405866] [rclcpp]: signal_handler(signum=2)

```

** Killing the client process does not cancel the goal and the server will continue to publish feedback messages. This should be resolved with @clalancette's fixes to service/clients.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -84,18 +86,128 @@ static const std::unordered_map<std::string, EntityType> str_to_entity = {
...
+ *
+ * <ReliabilityKind>:<DurabilityKind>:<HistoryKind>,<HistoryDepth>"
+ * Where:
+ *  <ReliabilityKind> - enum value from rmw_qos_reliability_policy_e.
+ *  <DurabilityKind> - enum value from rmw_qos_durability_policy_e.
+ *  <HistoryKind> - enum value from rmw_qos_history_policy_e.
+ *  <HistoryDepth> - The depth number.
+ */
+// TODO(Yadunund): Rely on maps to retrieve strings.
+std::string qos_to_keyexpr(rmw_qos_profile_t qos)
```
clalancette, line 158: Can you give an idea of what the keyexpression for QoS profiles looks like currently?  It kind of seems to me that it would be filled with enum numbers, but I could be wrong.

rmw_zenoh_cpp/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ find_package(rcutils REQUIRED)
 find_package(rosidl_typesupport_fastrtps_c REQUIRED)
 find_package(rosidl_typesupport_fastrtps_cpp REQUIRED)
 find_package(rmw REQUIRED)
+find_package(rmw_dds_common REQUIRED)
```
clalancette, line 24: I'm not a huge fan of linking to `rmw_dds_common`, mostly because it could confuse people about what `rmw_zenoh_cpp` is.  At the same time, there is quite a bit of code in https://github.com/ros2/rmw_dds_common/blob/rolling/rmw_dds_common/src/qos.cpp that would be somewhat unfortunate to duplicate.

I guess I'll ask; how compatible are the current QoS settings between Zenoh and DDS?  What are the things that are always going to remain different between them?

rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
```diff
@@ -69,7 +69,7 @@
   timestamping: {
     /// Whether data messages should be timestamped if not already.
     /// Accepts a single boolean value or different values for router, peer and client.
-    enabled: { router: true, peer: false, client: false },
+    enabled: { router: true, peer: true, client: false },
```
clalancette, line 74: Can you explain a little more about what this change does, and why we had it set to `false` before?

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -84,18 +86,128 @@ static const std::unordered_map<std::string, EntityType> str_to_entity = {
...
+  }
+
+  try {
+    qos.history = str_to_qos_history.at(history_parts[0]);
+    qos.reliability = str_to_qos_reliability.at(parts[0]);
+    qos.durability = str_to_qos_durability.at(parts[1]);
+  } catch (const std::exception & e) {
+    return std::nullopt;
+  }
+  sscanf(history_parts[1].c_str(), "%zu", &qos.depth);
```
clalancette: I don't think we should use `sscanf` to turn a string into an integer here.  It is too easy to misuse it, it doesn't report errors that well, and it is probably slow.

Instead I think we should probably use `strtoul`, with code similar to https://github.com/ros2/rmw_zenoh/blob/d49f9763253417da68062cadabfe30a2e0450f2b/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2168-L2188

clalancette (changes requested):
Besides what I left inline, I have two overall comments:

1.  When I'm running this code with `ros2 run demo_nodes_cpp talker`, I'm seeing a segfault on exit.  That doesn't seem to be the case on the `rolling` branch, so I think it bears looking into.
2.  I did some testing of TRANSIENT_LOCAL vs. VOLATILE on DDS and on Zenoh.  The one place where they differ is if the publisher is volatile, and the subscription is transient_local.  With Fast-DDS, in that case no data is transmitted and we get an Incompatible QoS warning.  With this PR in place, we don't get the incompatible QoS warning and data *is* transmitted.  If possible, I think it would be best if we could keep the semantics the same (at least for now).

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: I tried to explain in the docstring above this function defn. Copied below for reference.
```
/**
 * <ReliabilityKind>:<DurabilityKind>:<HistoryKind>,<HistoryDepth>"
 * Where:
 *  <ReliabilityKind> - enum value from rmw_qos_reliability_policy_e.
 *  <DurabilityKind> - enum value from rmw_qos_durability_policy_e.
 *  <HistoryKind> - enum value from rmw_qos_history_policy_e.
 *  <HistoryDepth> - The depth number.
 */
 ```
 
 Would it be clearer to include an example?

File: rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5 (hunk 3 above)
Comment (Yadunund) [reply rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5:74]: I included a note in the PR description
```
> Note: PublicationCache only works when time-stamping is enabled for peers in the zenoh configs.
```
I'll add some documentation inside the config

**Comment by Yadunund** [comment]
> When I'm running this code with ros2 run demo_nodes_cpp talker, I'm seeing a segfault on exit. That doesn't seem to be the case on the rolling branch, so I think it bears looking into.

I was able to reproduce this with a clean build and uncovered why we're seeing this after merging the latest `rolling` into this branch:
1. Turns out we've had [shm enabled in the session config](https://github.com/ros2/rmw_zenoh/blob/d49f9763253417da68062cadabfe30a2e0450f2b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5#L237) while it is [disbaled in the router config](https://github.com/ros2/rmw_zenoh/blob/d49f9763253417da68062cadabfe30a2e0450f2b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5#L237). But it should have been disabled in the session config as we observed memory leaks in shm and decided to not have it enabled by default.
2. As a result, we've been initializing [shm_manager in all sessions](https://github.com/ros2/rmw_zenoh/blob/a9b424505c55411dd37db402eae91663673980e8/rmw_zenoh_cpp/src/rmw_init.cpp#L181-L193) and actually getting memory for the published messages [from the shm_manager](https://github.com/ros2/rmw_zenoh/blob/a9b424505c55411dd37db402eae91663673980e8/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L879-L881).
3. I tracked the cause of the segfault to be [these lines that drop the shm_manager](https://github.com/ros2/rmw_zenoh/blob/a9b424505c55411dd37db402eae91663673980e8/rmw_zenoh_cpp/src/rmw_init.cpp#L348-L350) when shutting down the context. 
4. In https://github.com/ros2/rmw_zenoh/pull/95, we disabled all other features including `zenoh/shared-memory` and I think that's why we're seeing the sefault.
5.  I've pushed https://github.com/ros2/rmw_zenoh/pull/91/commits/a9b424505c55411dd37db402eae91663673980e8 which disables shm in the session config which resolves the issue. I also tried enabling build with `zenoh/shm-manager` feature but the segfault still happens if the config enables shm. We need more insights from zettascale on what other features we need to enable, eg `unstable`. For now I think disabling shm in the session config is the best way forward.

**Comment by Yadunund** [comment]
> I did some testing of TRANSIENT_LOCAL vs. VOLATILE on DDS and on Zenoh. The one place where they differ is if the publisher is volatile, and the subscription is transient_local. With Fast-DDS, in that case no data is transmitted and we get an Incompatible QoS warning. With this PR in place, we don't get the incompatible QoS warning and data is transmitted. If possible, I think it would be best if we could keep the semantics the same (at least for now).

Thanks for doing that test. I'm not sure if I agree with that behavior. Imo the subscription should still continue receiving new messages from the publisher BUT with a warning printed that older messages will not be received. But having said that, I'm surprised the implementation here does not throw the same error since we fully reuse [rmw_dds_common::qos_profile_check_compatibility](https://github.com/ros2/rmw_dds_common/blob/ae7df87714cccf4d7ad54dcd00201f4931f2060f/rmw_dds_common/src/qos.cpp#L87-L139) and hence my comment here https://github.com/ros2/rmw_zenoh/pull/91#discussion_r1464295727. 🤔 
Looking into it.

File: rmw_zenoh_cpp/CMakeLists.txt (hunk 2 above)
Comment (Yadunund) [reply rmw_zenoh_cpp/CMakeLists.txt:24]: imo the biggest difference is the lack of QoS settings, specifically `reliability` for publishers. Zenoh separates control over message resending, memory usage (buffering) and message dropping. 
The reason I chose to reuse some of the methods from `rmw_dds_common` is to ensure qos compatibility checks and generation of adapted qos profiles will remain the same for users coming from DDS based rmws.

File: rmw_zenoh_cpp/src/detail/liveliness_utils.cpp (hunk 4 above)
Comment (Yadunund) [reply rmw_zenoh_cpp/src/detail/liveliness_utils.cpp]: Done https://github.com/ros2/rmw_zenoh/pull/91/commits/52d16cb5027bb58f9ebb379b36428f149dea09f2

File: rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5 (hunk 3 above)
Comment (Yadunund) [reply rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5:74]: Done https://github.com/ros2/rmw_zenoh/pull/91/commits/34f2623fe781ad68fcf0abf7107cd99dd341de89

**Comment by Yadunund** [comment]
> > I did some testing of TRANSIENT_LOCAL vs. VOLATILE on DDS and on Zenoh. The one place where they differ is if the publisher is volatile, and the subscription is transient_local. With Fast-DDS, in that case no data is transmitted and we get an Incompatible QoS warning. With this PR in place, we don't get the incompatible QoS warning and data is transmitted. If possible, I think it would be best if we could keep the semantics the same (at least for now).
> 
> Thanks for doing that test. I'm not sure if I agree with that behavior. Imo the subscription should still continue receiving new messages from the publisher BUT with a warning printed that older messages will not be received. But having said that, I'm surprised the implementation here does not throw the same error since we fully reuse [rmw_dds_common::qos_profile_check_compatibility](https://github.com/ros2/rmw_dds_common/blob/ae7df87714cccf4d7ad54dcd00201f4931f2060f/rmw_dds_common/src/qos.cpp#L87-L139) and hence my comment here [#91 (comment)](https://github.com/ros2/rmw_zenoh/pull/91#discussion_r1464295727). 🤔 Looking into it.

After some digging, with `Fast-DDS`, the warning message is thrown because of qos incompatibility from the [default_incompatibile_qos_callback](https://github.com/ros2/rclcpp/blob/a5221336f64e9754377e5391663dc6c7f3da3975/rclcpp/src/rclcpp/subscription_base.cpp#L133C15-L133C48). Since we don't support events yet, I think this isn't being flagged. But also since the zenoh transport can handle such qos mismatches, I don't think we will need to flag it? It's one of the reasons I left this [todo](https://github.com/ros2/rmw_zenoh/blob/52d16cb5027bb58f9ebb379b36428f149dea09f2/rmw_zenoh_cpp/src/detail/graph_cache.cpp#L688) in the impl here. I'm not sure if we should report incompatibility to match DDS behavior since zenoh transport can handle the differences.

**Comment by Yadunund** [comment]
Went back and forth on whether to rely on a blocking vs non-blocking channel for making the liveliness query at startup. 

~~I think we should stick to blocking for now (https://github.com/ros2/rmw_zenoh/pull/91/commits/78ab5a18f9c3deda2490246ad5f4115b10d42bea). Rationale is fewer threads needed (i think) and perhaps more accurate results when using CLI tools like `ros2 topic list` as the graph cache will be fully updated by the end of `rmw_init`.~~
~~It's better to rely on a non-blocking channel to avoid CLI tools waiting indefinitely when there are no other nodes in the graph.~~

After clarification from zettascale, we should stick to a `blocking` channel https://github.com/ros2/rmw_zenoh/commit/2ad493d518c06d827df7f8a8d7da1acdfea9bd4a.

**Comment by Yadunund** [comment]
I've set the `SystemDefault` depth to 1 via https://github.com/ros2/rmw_zenoh/pull/91/commits/1298abd1ea3124b4a1bca54c4eec6795206766d8.

The default depth is 1 in [rmw_cyclonedds](https://github.com/ros2/rmw_cyclonedds/blob/f3d802e3c7ae3331e88868cda010bf9c17dd0655/rmw_cyclonedds_cpp/src/rmw_node.cpp#L2196-L2197).

File: rmw_zenoh_cpp/CMakeLists.txt (hunk 2 above)
Comment (clalancette) [reply rmw_zenoh_cpp/CMakeLists.txt:24]: In order to move forward with this PR, I say we take the dependency for now.  But I'm also going to suggest that we open up a new issue to revisit this, and possibly drop the dependency in the future.

File: rmw_zenoh_cpp/src/detail/liveliness_utils.cpp [hunk 5]
```diff
@@ -84,18 +88,147 @@ static const std::unordered_map<std::string, EntityType> str_to_entity = {
...
+
 std::string zid_to_str(z_id_t id)
 {
   std::stringstream ss;
   ss << std::hex;
   size_t i = 0;
-  for (; i < (sizeof(id.id) - 1); i++) {
-    ss << static_cast<int>(id.id[i]) << ".";
+  for (; i < (sizeof(id.id)); i++) {
+    ss << static_cast<int>(id.id[i]);
```
clalancette, line 122: This is changing the string representation of the zid from XX.YY.ZZ to XXYYZZ.  Can you explain the reasoning behind this?

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
clalancette: Yeah, I think an example would help, thanks.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -84,18 +88,147 @@ static const std::unordered_map<std::string, EntityType> str_to_entity = {
...
+      result.push_back(keyexpr.substr(prev_idx + 1, idx - prev_idx - 1));
+    }
+  } catch (const std::exception & e) {
+    printf("%s\n", e.what());
+    return {};
+  }
+  // Finally add the last substr.
+  result.push_back(keyexpr.substr(delim_idx.back() + 1));
+  return result;
+}
```
clalancette: Looking at the implementation of this function, it actually seems like this is unnecessarily iterating over the string twice: once to find the delimiters, and then again to copy the substrings into the vector.

But we should be able to make this more efficient by just iterating once.  Basically we iterate, looking for the delimiter, and copying into the result as we go along.  Something like this:

```
std::vector<std::string> split_keyexpr(
  const std::string & keyexpr,
  const char delim)
{
  std::vector<std::string> result = {};
  size_t start = 0;
  size_t end = keyexpr.find(delim);
  while (end != std::string::npos) {
    result.push_back(keyexpr.substr(start, end - start));
    start = end + 1;
    end = keyexpr.find(delim, start);
  }
  result.push_back(keyexpr.substr(start));
  return result;
}
```

In some basic benchmarking locally, this is about twice as fast as the current implementation.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -55,6 +55,10 @@
 #include "rmw/validate_namespace.h"
 #include "rmw/validate_node_name.h"
 
+#include "rmw_dds_common/qos.hpp"
+
+#define RMW_ZENOH_DEFAULT_HISTORY_DEPTH 1;
```
clalancette: I think we agreed to change this to 10, 100, or possibly 256 (the default Zenoh queue size).

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1231,14 +1280,19 @@ rmw_create_subscription(
...
-    sub_data->reliable = true;
+  // Adapt any 'best available' QoS options
+  sub_data->adapted_qos_profile = *qos_profile;
+  rmw_ret_t ret = rmw_dds_common::qos_profile_get_best_available_for_topic_subscription(
+    node, topic_name, &sub_data->adapted_qos_profile, rmw_get_publishers_info_by_topic);
+  if (RMW_RET_OK != ret) {
+    RMW_SET_ERROR_MSG("Failed to obtain adapted_qos_profile.");
+    return nullptr;
   }
+  // If a depth of 0 was provided, the RMW implementation can set the depth to a default.
```
```suggestion
  // If a depth of 0 was provided, the RMW implementation should choose a suitable default.
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1306,19 +1359,71 @@ rmw_create_subscription(
...
+      z_move(callback),
+      &sub_options
+    );
+    if (!z_check(std::get<ze_owned_querying_subscriber_t>(sub_data->sub))) {
+      RMW_SET_ERROR_MSG("unable to create zenoh subscription");
+      return nullptr;
+    }
   }
+  // Create a regular subscriber for all other durability settings.
+  else {
```
```suggestion
  } else {
    // Create a regular subscriber for all other durability settings.
```

clalancette (changes requested):
I did another review pass.  This is looking pretty good in general, just a few things to cleanup and followup on.

rmw_zenoh_cpp/CMakeLists.txt
Yadunund: Done https://github.com/ros2/rmw_zenoh/issues/104

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: This was a minor "optimization" I did to reduce the number of bytes in the keyexpression and reduce bandwidth.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/91/commits/c807b34855ffbb454c0056c2fa02d1759cf64116

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Doh I don't know why the original was so complicated. Updated to your more efficient approach https://github.com/ros2/rmw_zenoh/pull/91/commits/009837101b2f70e3b6cae2db88a45cbbd742a4cb

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/91/commits/e85d45d6d3796b08dc7983562aa025d5ed4a8a56

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -84,18 +88,136 @@ static const std::unordered_map<std::string, EntityType> str_to_entity = {
...
+    qos.reliability = str_to_qos_reliability.at(parts[0]);
+    qos.durability = str_to_qos_durability.at(parts[1]);
+  } catch (const std::exception & e) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("Error setting QoS values from strings: %s", e.what());
+    return std::nullopt;
+  }
+  // Get the history depth.
+  errno = 0;
+  char * endptr;
+  int64_t num = strtol(history_parts[1].c_str(), &endptr, 10);
```
clalancette: Given that `qos.depth` is a [size_t](https://github.com/ros2/rmw/blob/9b3d9d0e3021b7a6e75d8886e3e061a53c36c789/rmw/include/rmw/types.h#L574), I think we should use `strtoul` here:

```suggestion
  unsigned long num = strtoul(history_parts[1].c_str(), &endptr, 10);
```

clalancette (changes requested):
I've left one more relatively minor change.  Once that is done, I think this is good to go in.

clalancette (approved):
I just pushed some minor cpplint fixes.  With that, I'm happy with this, so once CI comes back green I'm going to go ahead and merge this.

PR #92, opened 2024-01-18
Improve service and subscription reliability

clalancette:
This rather large PR is the culmination of about a week's worth of work.  The ultimate goal of this work was to improve service reliability, but along the way numerous fixes to publishers, subscriptions, clients, services, and wait sets were made.

It's probably helpful to read the description of each of the commits, but roughly:

* We switch to unique_ptrs for dealing with the server incoming requests and the client incoming response (this was suggested in the review for #86, and implemented now).
* We switch to RAII for the query replies, responses, and subscription data, which means we clean up in more paths.
* We remove a ton of debugging statements.
* We add in source_timestamps and writer_guid to clients and services.
* We clean up a bunch of memory leaks.
* We revamp how rmw_wait works, ensuring that we get all of the data that is incoming.

With this PR in place, services now work reliably for me in local testing.  It definitely needs more testing on larger systems, but these cleanups should get us a lot closer to having that work.


rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp
```diff
@@ -56,15 +56,12 @@ rmw_ret_t zenoh_router_check(z_session_t session)
   // Define callback
   auto callback = [](const struct z_id_t * id, void * ctx) {
       const std::string id_str = zid_to_str(*id);
-      RCUTILS_LOG_INFO_NAMED(
```
Yadunund, line 59: Just realized that there is a `zid_to_str()` method defined here and also within `liveliness_utils.cpp`. I'll ticket it to be fixed later.

Yadunund (approved):
Thanks a lot for hunting down these fixes! The added documentation in `rmw_wait` is very insightful!

PR #95, opened 2024-01-20
Limit threads per zenoh session

### LINKED ISSUE #94: Thread creation error when running many nodes that spawn a lot of threads
Starting applications like [rmf_demos](https://github.com/open-rmf/rmf_demos) that launches over 30 ROS nodes and spins numerous threads, leads to runtime error, `Thread creation error (Resource temporarily unavailable)`, which indicate that the system is unable to create any new threads due to resource constraints. 
Certain zenoh transport rely on `async-std` where the global runtime does NOT manage a thread pool and hence leads to a lot of threads spawned by the zenoh session.

Yadunund:
Fix #94 by passing `"--no-default-features;--features=zenoh/transport_tcp"` to the cargo command invoked when building `zenoh-c`. The default session config is also updated to limit the threads to 1.

Special thanks to @cottsay for advising on how to deal with the `;`s in cmake 

cc: @Mallets

### DIFF
--- a/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
+++ b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
@@ -187,6 +187,13 @@
         },
         // Number of threads dedicated to transmission
         // By default, the number of threads is calculated as follows: 1 + ((#cores - 1) / 4)
+        // We limit the number of threads that the zenoh session can spin to 1.
+        // Without this limit, applications with multiple zenoh sessions can
+        // encounter system resource errors when trying to create new threads.
+        // Once zenoh migrates to relying on tokio for its async runtime,
+        // see https://github.com/eclipse-zenoh/zenoh/pull/566, we can consider
+        // removing these flags since with tokio, zenoh can better manage the threads it spins
+        // with the help of thread pools.
         threads: 1,
       },
       /// Configure the zenoh RX parameters of a link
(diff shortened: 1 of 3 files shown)


Mallets:
The Zenoh configuration looks good to me.

For tracking purposing, I link here a forthcoming PR on Zenoh (https://github.com/eclipse-zenoh/zenoh/pull/566) which will impact the configuration of the Zenoh runtime. Note the the PR will impact only the configuration file, not CMake.

rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
```diff
@@ -185,6 +185,9 @@
           /// Higher values lead to a more aggressive batching but it will introduce additional latency.
           backoff: 100,
         },
+        // Number of threads dedicated to transmission
+        // By default, the number of threads is calculated as follows: 1 + ((#cores - 1) / 4)
+        threads: 1,
```
clalancette: In the CMakeLists.txt, we have a nice comment explaining why we want to limit the threads.  Can we have a similar comment here?  It might also be a good idea to link to the PR that Luca mentioned.

clalancette (commented):
Besides what I said inline, do we need a similar configuration for the router build/config file?  I don't know the answer, so just asking here.

rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
Yadunund: Good call. I've spoken to @Mallets and it's better to limit it there too although the router is only used to forward discovery data. Also added the note in the configs https://github.com/ros2/rmw_zenoh/pull/95/commits/ec36e0f3304e50a48af81bb45b96d9eff4ae7a52

clalancette (approved):

PR #96, opened 2024-01-22
Remove the router config for threads.

clalancette:
It doesn't seem to work; the router throws an error:

JSON error: unknown field `threads`, expected `size` or `backoff` at commons/zenoh-config/src/lib.rs:601.', zenohd/src/main.rs:176:42

This partially undoes #95.  @Yadunund FYI, I'm just going to go ahead and merge this one since the router doesn't work at all right now.


Yadunund:
Ah my bad I found an old `threads` key commented out in the router config and just uncommented it without checking if it's in the right place 🤦🏼‍♂️

PR #97, opened 2024-01-23
Add threads limit to router config

Yadunund:
Follow up to #96


clalancette (approved):
I tested this out, and it works fine.  Thanks!

PR #98, opened 2024-01-23
More overall improvements to the code

clalancette:
This small series does another pass through the subscription, clients, services, and guard condition code, making improvements.  In particular:

1.  It fixes a bug where guard conditions would be ready over and over again.  This would cause very high CPU load.  Now once we get the status of a guard condition, we also reset the trigger.
2.  It encapsulates subscriptions, services, and client code into their own classes.  This allows us to make more of the class fields private, and do proper locking around them.  It also has a side benefit of moving some code out of rmw_zenoh.cpp, which has grown quite large.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
@@ -228,16 +228,14 @@ private:
   z_owned_reply_t reply_;
 };
 
-struct rmw_client_data_t
+class rmw_client_data_t final
 {
+public:
   z_owned_keyexpr_t keyexpr;
   z_owned_closure_reply_t zn_closure_reply;
 
   // Liveliness token for the client.
   zc_owned_liveliness_token_t token;
-
-  std::mutex replies_mutex;
-  std::deque<std::unique_ptr<ZenohReply>> replies;
 
   const void * request_type_support_impl;
   const void * response_type_support_impl;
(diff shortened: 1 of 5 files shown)


rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -121,14 +122,27 @@ struct rmw_subscription_data_t
...
 
-  std::deque<std::unique_ptr<saved_msg_data>> message_queue;
-  std::mutex message_queue_mutex;
-
   size_t queue_depth;
   bool reliable;
 
-  std::mutex internal_mutex;
-  std::condition_variable * condition{nullptr};
+  void attach_condition(std::condition_variable * condition_variable);
```
Yadunund, line 128: How do you feel about defining a separate concrete class with the `condition_` and `condition_mutex_` as protected members, and with the `notify()`, `attach_condition()` and `detach_condition()` methods defined? Then each of `rmf_X_data_t` classes will inherit this class. It will reduce code duplication a bit.

clalancette: While I totally understand the desire to do so, in this case I don't think it is a good idea.  As it stands, these are just "fancy structs" that have some methods attached to them.  As soon as you add in inheritance, then you add in complications with polymorphism, and also probably slower code due to the addition of vtables.  The latter can probably be mitigated with some clever definition of destructors, but the amount of duplicated code here is minimal, so I'd rather just duplicate it.

Yadunund: Sounds good, thanks for clarifying.

Yadunund (approved):
LGTM! Thanks for cleaning up the implementation.

Issue #78, opened 2023-12-02
Incorrect error handling when zenoh router node is terminated 


Yadunund:
Killing the router node (Crtl+C) started by `ros2 run rmw_zenoh_cpp init_rmw_zenoh_router` will return this error 
```bash
^CFailed to run zenoh router: zenohd -c ../ws_rmw/install/rmw_zenoh_cpp/share/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
```

We should try to handle sigint and sigterm cases in this return https://github.com/ros2/rmw_zenoh/blob/beb9fd880227375c1f81b9ca8be06978a826fe04/rmw_zenoh_cpp/apps/init_rmw_zenoh_router.cpp#L36

clalancette:
#99 somewhat addresses this, so closing.

PR #99, opened 2024-01-23
Fix handling of SIGINT by router executable

Yadunund:
Aim to address https://github.com/ros2/rmw_zenoh/issues/78

According to the [manual page()](https://linux.die.net/man/3/system), `system` will return `-1` on error.

However right now, killing the process via SIGINT (CTRL+C) returns with error code `2`. Investigating...
```bash
[2024-01-23T14:48:43Z INFO  zenohd] Starting required plugin "rest"
[2024-01-23T14:48:43Z INFO  zenohd] Successfully started plugin rest from "/usr/lib/libzenoh_plugin_rest.so"
[2024-01-23T14:48:43Z INFO  zenohd] Finished loading plugins
^C
Thread 1 "zenohd" received signal SIGINT, Interrupt.
syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
38	../sysdeps/unix/sysv/linux/x86_64/syscall.S: No such file or directory.
(gdb) bt
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00005555557a6762 in parking::Inner::park ()
#2  0x0000555555696c43 in zenohd::main ()
#3  0x000055555561eaf3 in std::sys_common::backtrace::__rust_begin_short_backtrace ()
#4  0x00005555556b23a7 in main ()
(gdb) print $eax
$1 = -512

```

### DIFF
--- a/rmw_zenoh_cpp/apps/init_rmw_zenoh_router.cpp
+++ b/rmw_zenoh_cpp/apps/init_rmw_zenoh_router.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <stdlib.h>
+#include <sys/wait.h>
 
 #include <iostream>
 #include <string>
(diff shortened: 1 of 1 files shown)


clalancette (approved):
When running this on `rolling`, I get:

```
^CFailed to run zenoh router: zenohd -c /home/clalancette/zenoh_ws/install/rmw_zenoh_cpp/share/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
[ros2run]: Process exited with failure 2
```

With this PR in place, I get:
```
^CZenoh router exited abnormally with error code [2]
```

So it is overall an improvement.

I'll just point out that almost certainly we are going to have to rethink this completely when we start testing this on Windows.

PR #103, opened 2024-01-31
Support Events

Yadunund:
The PR adds support for two independent event mechanisms
- [x] Registering and triggering of user defined callbacks when messages, requests and responses arrive -> Used by `EventsExecutor`.
- [x] Registering and triggering of user defined callbacks when the ROS Graph changes -> QoS events such as `MatchedEvent`. #118 

## How events function in rmw
1. Event callbacks are set by the client library and to be triggered by the `rmw` layer when a new message/request/response arrives. Then, the client library can call `rmw_event_take()` to get the event instead of waiting on condition variables in `rmw_wait()`.
2. Event callbacks for QoS events may be set. Presently only publisher/subscription events statuses are defined in `rmf/events.h`.  Here `rmw_publisher_event_init` and `rmw_subscription_event_init` methods will instantiate the event based on the `event_type` supplied. The RMW implementation has full discretion to initialize/reject (by returning a `nullptr`) an event initialization request based on what events the implementation supports. The [rmw_events_type_t definition](https://github.com/ros2/rmw/blob/9b3d9d0e3021b7a6e75d8886e3e061a53c36c789/rmw/include/rmw/event.h#L33-L52) enumerates the types of events that can be supported while an internal `events_map` data structure defines which events are supported by `rmw_zenoh`. Once an event is initialized, `rcl` will call `rmw_event_set_callback()` to register a callback for the event initialized previously. Then in `rmw_wait`, the RMW implementation should attach a condition variable to the event. The RMW implementation's job is to then track changes to the ROS Graph that are relevant to the pub/sub entities created in the current session and for which the callbacks have been registered. When a change is detected, an "Event instance" is to be added to an "event queue" and the attached condition variable is used to notify `rmw_wait` that an event has occurred. Finally, upon notification, `rcl` will invoke `rmw_take` to take ownership of the event.

## Changes

~This PR makes a big change to the internal data structures used to store publisher, subscription, service, and client data. All of these classes now inherit an `EventsBase` class. While this approach was [discussed earlier](https://github.com/ros2/rmw_zenoh/pull/98#discussion_r1463439158), it was apparent during the implementation that casting these different classes to the same `EventsBase` class is necessary as the [rmw_events_type_t definition](https://github.com/ros2/rmw/blob/9b3d9d0e3021b7a6e75d8886e3e061a53c36c789/rmw/include/rmw/event.h#L33-L52) bundles events for publishers and subscriptions. It also minimizes the code duplication when implementing `rmw_take` and while setting/triggering callbacks.~

This PR adds several new features and fixes bugs while making some big changes:
- **GraphCache**
    - Allow event callbacks to be registered within `GraphCache` for local pub/subs.
    - Data structure changes to track and store all details of any entity added to the graph, ie store `liveliness::Entity` entries
    - Added mechanisms to keep track of event changes and trigger event callbacks.
    - `struct TopicStats` is removed to simplify nested data structures.
    - Fixed a bug where services and clients were not removed from the GraphcCache when their node is deleted before entities (this behavior is a bug in `rclcpp` but atleast the consequence is now properly handled).
    - No more lambdas
    - No more hardcoded indices in `liveliness.cpp`. Enums are used to keep track of segments in the liveliness tokens.
 - **QoS events**
     - `RMW_EVENT_REQUESTED_QOS_INCOMPATIBLE`: We support it but this will never be triggered since there are no QoS incompatibilities in zenoh.
     - `RMW_EVENT_SUBSCRIPTION_MATCHED`: A match is reported when a sub is added to the graph and there are other pubs with the same topic and type. QoS does not need to be checked for same reasons as above.
     - `RMW_EVENT_OFFERED_QOS_INCOMPATIBLE`: Supported but never triggered as stated above.
     - `RMW_EVENT_PUBLICATION_MATCHED`: A match is reported when a pub is added to the graph and there are other subs with the same topic and type. QoS does not need to be checked for same reasons as above.
     - QoS compatibility checking was also updated to produce a warning for `TRANSIENT_LOCAL` pub and `VOLATILE` sub durabilities. All other combinations are `COMPATIBLE` to reflect zenoh protocol's behavior.
-  **Liveliness tokens** published now embed `zenoh session id`, `unique id of the entity within the session` and `the unique id of the node`. If the entity is a node, this id is appended twice to keep the number of segments consistent. This allows
      - Uniquely track any entity in the system with a combination of `zenoh session id`, `node id` and `entity id`.
      - Mapping changes to the graph cache to events.
      - Uniquely track two nodes with the same name from the same session (ie load two `talker` nodes into the same `rclcpp_components::ComponentManger`
 - **rmw_X_data_type_t**
     - No longer inherit `EventsBase` 🎉 
     - Pub & Sub data types now store an `EventsManager` class. (There are no events defined for services and clients in `rmw/events.h`.
     - Sub, Client & Service data types store a `DataCallbackManager` class. (Publishers do not need to trigger any callbacks).
      
### DataCallback tests: This functionality is utilized by the `experimental::EventsEcecutor` in `rclcpp`.
`EventsExecutor` support can be tested by running any of the demo executables from `demo_nodes_cpp`. This [PR](https://github.com/ros2/demos/pull/666) is helpful for building the executables to run the `EventsExecutor`.

### Matched events
Follow instructions to run `matched_event_detect` from `demo_nodes_py`: https://github.com/ros2/demos/tree/rolling/demo_nodes_cpp#matched-event-detect
>Note: I'm seeing some flaky outputs from `demo_nodes_cpp` executable and I think its coming from `PublisherBase` in `rclcpp`. 

### Nodes with same names from the same zenoh session
Save this text into a python file, source ROS 2, and execute it
```python
import rclpy
from rclpy.node import Node
from rclpy.executors import SingleThreadedExecutor


class myNode(Node):
  def __init__(self):
    super().__init__("my_node")

def main(args=None):
  rclpy.init(args=args)
  node1 = myNode()
  node2 = myNode()
  executor = SingleThreadedExecutor()
  executor.add_node(node1)
  executor.add_node(node2)
  executor.spin()

if __name__ == "__main__":
  main()
```

In a second terminal run `ros2 node list`. You should be able to see two `/my_node` nodes. Without the changes here, this would not be the case as the `GraphCache` was only capable of handling a singe node name from the same zenoh session.
```bash
$ ros2 node list
/my_node
/my_node
```

### Incompatible QoS
The nodes in `ros2/demos` are not relevant here since Zenoh does not have the same QoS behaviors as DDS and hence the logic in the executables does not apply.

## Future work
- Implement incompatible type events (I started on this but realized the PR is getting too big so will save that for later)
- Implement message lost events.

## Some other takeaways:
- The `rmw` API should really provide two separate enums, `rmw_publisher_events_type_t` and `rmw_subscription_events_type_t` as the current API does not enforce the fact that events registered for publishers should only be of the "publisher types" and vice versa for subscriptions.
- With this PR, RViz is finally supported. Although imo Rviz should not expect events support in every RMW implementation for it to function.


rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include "rcutils/logging_macros.h"
+
+#include "rmw/error_handling.h"
+
+
+///=============================================================================
+void EventsBase::set_user_callback(
+  const void * user_data, rmw_event_callback_t callback)
+{
+  std::lock_guard<std::recursive_mutex> lock_mutex(event_mutex_);
```
clalancette: Does this need to be a `recursive_mutex`?  If we know we need it, then we can leave it, but recursive mutexes tend to be much more expensive than regular ones.

rmw_zenoh_cpp/src/rmw_event.cpp
```diff
@@ -100,20 +131,57 @@ rmw_take_event(
...
+  std::unique_ptr<rmw_zenoh_event_status_t> st = event_data->pop_next_event(
+    zenoh_event_it->second);
+  if (st == nullptr) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "rmw_take_event called when event queue for event type [%d] is empty",
+      event_handle->event_type);
+    return RMW_RET_ERROR;
+  }
+
+  // Now depending on the event, populate the rwm event status.
```
clalancette:
```suggestion
  // Now depending on the event, populate the rmw event status.
```

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -70,9 +71,10 @@ struct rmw_node_data_t
   zc_owned_liveliness_token_t token;
 };
 
-///==============================================================================
-struct rmw_publisher_data_t
+///=============================================================================
+class rmw_publisher_data_t : public EventsBase
```
clalancette: I have to say that making all of publishers, subscriptions, services, and clients derive from `EventsBase` seems weird.  That is, inheritance is an "is-a" relationship, and I don't think of publishers as events, for instance.

As I understand it, the basic problem here is that we want to be able to call `trigger_event`, `event_set_callback`, `pop_next_event`, `event_queue_is_empty`, `event_attach_condition`, and `event_detach_condition` on all of these.    One way I can think to do this is to make another class called `RMWEntity`, or something like that, and then `RMWEntity` would contain an EventBase class.  That way it would be a "has-a" relationship.  There may also be some other ways of restructuring this class hierarchy to make it clearer what we are trying to do; I think it is worth thinking about some more and exploring.

clalancette (changes requested):
I did a first pass review.  I didn't review the changes to the GraphCache in any detail; I'll do that next.

My biggest concern has to do with the class hierarchy we are introducing here.  I've left a more detailed comment inline.

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
Yadunund: Thanks for the suggestion. Happy to switch to the "has-a" pattern. How about a restructuring like this:
* I will split the `EntitiesBase` into `UserCallbackManager` and `QoSEventManager` classes. The former handles user specified callbacks while the later manages QoS events that we support.
* The `rmw_subscription/client/service_data_t` classes will have an object of type `CallbackManager` while only `rmw_publisher/subscription_data_t` classes will additionally contain `QoSEventsManger` objects since QoS events are not are not something for clients & services.

```bash
class UserCallbackManager
{
public:
  /// @brief Set the user defined callback that should be called when
  /// a new message/response/request is received.
  /// @param user_data the data that should be passed to the callback.
  /// @param callback the callback to be set.
  void set_user_callback(const void * user_data, rmw_event_callback_t callback);

  /// Trigger the user callback.
  void trigger_user_callback();

  
private:
  std::mutex event_mutex_;
  /// User callback that can be set via set_user_callback().
  rmw_event_callback_t callback_ {nullptr};
  /// User data that should be passed to the user callback.
  const void * user_data_ {nullptr};
  /// number of trigger requests made before the callback was set.
  size_t unread_count_ {0};
};

class QoSEventsManager
{
public:
  /// @brief  Set the callback to be triggered when the relevant event is triggered.
  /// @param event_id the id of the event
  /// @param callback the callback to trigger for this event.
  /// @param user_data the data to be passed to the callback.
  void event_set_callback(
    rmw_zenoh_event_type_t event_id,
    rmw_event_callback_t callback,
    const void * user_data);

  /// @brief  Returns true if the event queue is empty.
  /// @param event_id the event id whose event queue should be checked.
  bool event_queue_is_empty(rmw_zenoh_event_type_t event_id) const;

  /// Pop the next event in the queue.
  /// @param event_id the event id whose queue should be popped.
  std::unique_ptr<rmw_zenoh_event_status_t> pop_next_event(
    rmw_zenoh_event_type_t event_id);

  /// Add an event status for an event.
  /// @param event_id the event id queue to which the status should be added.
  void add_new_event(
    rmw_zenoh_event_type_t event_id,
    std::unique_ptr<rmw_zenoh_event_status_t> event);

  /// @brief Attach the condition variable provided by rmw_wait.
  /// @param condition_variable to attach.
  void attach_event_condition(
    rmw_zenoh_event_type_t event_id,
    std::condition_variable * condition_variable);

  /// @brief Detach the condition variable provided by rmw_wait.
  void detach_event_condition(rmw_zenoh_event_type_t event_id);

private:
  /// @brief Trigger the callback for an event.
  /// @param event_id the event id whose callback should be triggered.
  void trigger_event_callback(rmw_zenoh_event_type_t event_id);

  /// Notify once event is added to an event queue.
  void notify_event(rmw_zenoh_event_type_t event_id);

  /// Mutex to lock when read/writing members.
  // The mutex is recursive as add_new_event() invokes `trigger_event_callback()`.
  mutable std::recursive_mutex event_mutex_;
  /// Mutex to lock for event_condition.
  mutable std::mutex event_condition_mutex_;
  /// Condition variable to attach for event notifications.
  std::condition_variable * event_conditions_[ZENOH_EVENT_ID_MAX + 1]{nullptr};
  rmw_event_callback_t event_callback_[ZENOH_EVENT_ID_MAX + 1] {nullptr};
  const void * event_data_[ZENOH_EVENT_ID_MAX + 1] {nullptr};
  size_t event_unread_count_[ZENOH_EVENT_ID_MAX + 1] {0};
  // A dequeue of events for each type of event this RMW supports.
  std::deque<std::unique_ptr<rmw_zenoh_event_status_t>> event_queues_[ZENOH_EVENT_ID_MAX + 1] {};
  const std::size_t event_queue_depth_ = 10;
};
```

Yadunund: No more inheritance :) https://github.com/ros2/rmw_zenoh/pull/103/commits/29f3a79f039ef679fdf9a75e6cd233d6445ae7aa

Yadunund:
@clalancette i've overhauled this PR and updated the description as well. Keeping track of messages lost is the only major event yet to be supported but I think this PR is getting way too big. I will work on that once this is reviewed and merged as it may also require changes from your GID PRs.
On that note, one thing I did in this PR was to track the `hash` of the liveliness token's keyexpr as a `GUID` for the entity. I intentionally did not call it GID to mark the difference. One thing we could could do is replace the `std::size` hash value to an `rmw_gid_t` type instead but still generate the GID based on the unique hash? Let me know!

rmw_zenoh_cpp/src/detail/event.cpp
Yadunund: No longer recursive.

rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include "rcutils/logging_macros.h"
+
+#include "rmw/error_handling.h"
+
+
+///=============================================================================
+void DataCallbackManager::set_callback(
+  const void * user_data, rmw_event_callback_t callback)
+{
+  std::lock_guard<std::mutex> lock_mutex(event_mutex_);
```
clalancette, line 28: `#include <mutex>` at the top for this.

rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,229 @@
...
+    return true;
+  }
+
+  std::lock_guard<std::recursive_mutex> lock(event_mutex_);
+
+  return event_queues_[event_id].empty();
+}
+
+///=============================================================================
+std::unique_ptr<rmw_zenoh_event_status_t> EventsManager::pop_next_event(
```
clalancette, line 122: `#include <memory>` at the top for this.

rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+  if (event_id > ZENOH_EVENT_ID_MAX) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "RMW Zenoh is not correctly configured to handle rmw_zenoh_event_type_t [%d]. "
+      "Report this bug.",
+      event_id);
+    return;
+  }
+
+  std::lock_guard<std::recursive_mutex> lock(event_mutex_);
```
clalancette: I'm not totally opposed, by why a `recursive_mutex` here?

rmw_zenoh_cpp/src/detail/event.hpp
```diff
@@ -0,0 +1,175 @@
...
+  // publisher events
+  // RMW_EVENT_LIVELINESS_LOST,
+  // RMW_EVENT_OFFERED_DEADLINE_MISSED,
+  ZENOH_EVENT_OFFERED_QOS_INCOMPATIBLE,
+  ZENOH_EVENT_PUBLISHER_INCOMPATIBLE_TYPE,
+  ZENOH_EVENT_PUBLICATION_MATCHED,
+};
+
+/// Helper value to indicate the maximum index of events supported.
+#define ZENOH_EVENT_ID_MAX rmw_zenoh_event_type_t::ZENOH_EVENT_PUBLICATION_MATCHED
```
clalancette, line 51: This is pretty minor, but our typical style here is to add a "SENTINEL" value at the end of the enum, and check against that instead.  So this would change to:

```suggestion
  ZENOH_EVENT_INVALID
};
```

And then the checks in `event.cpp` would change to `>= ZENOH_EVENT_INVALID`.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
-TopicStats::TopicStats(std::size_t pub_count, std::size_t sub_count)
-: pub_count_(pub_count),
-  sub_count_(sub_count)
+TopicData::TopicData(
+  liveliness::TopicInfo info,
+  std::unordered_set<liveliness::Entity> pubs,
+  std::unordered_set<liveliness::Entity> subs)
+: info_(std::move(info)),
+  pubs_(std::move(pubs)),
+  subs_(std::move(subs))
```
clalancette: This is probably me just not know C++, but the ownership semantics are pretty unclear to me here.  Is `TopicData` taking ownership of these objects?  If so, why are they being copied in to begin with?  If `TopicData` is not taking ownership, what is the lifetime of these objects?  Would it be clearer to use a `std::unique_ptr` here?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
 }
 
 ///=============================================================================
-TopicData::TopicData(
-  liveliness::TopicInfo info,
-  TopicStats stats)
-: info_(std::move(info)),
-  stats_(std::move(stats))
-{}
+std::shared_ptr<GraphNode> GraphCache::make_graph_node(const Entity & entity) const
```
clalancette, line 76: Should we just turn `GraphNode` into a class, and have a proper constructor for it?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+  std::unordered_set<liveliness::Entity> subs = {};
+  if (is_pub) {
+    pubs.insert(entity);
+  } else {
+    subs.insert(entity);
+  }
+  TopicDataPtr graph_topic_data = std::make_shared<TopicData>(
+    topic_info,
+    std::move(pubs),
+    std::move(subs));
```
clalancette: Actually, this relates to my previous comment.  It might make more sense to pass the `entity` into the `TopicData` constructor, and then just have `TopicData` own the `pubs` and `subs` unordered_sets.  That would resolve the ownership question.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+            for (const liveliness::Entity & sub_entity : topic_data_ptr->subs_) {
+              update_event_counters(
+                topic_info.name_,
+                ZENOH_EVENT_SUBSCRIPTION_MATCHED,
+                static_cast<int32_t>(1));
+              if (is_entity_local(sub_entity)) {
+                local_entities_with_events[sub_entity].insert(ZENOH_EVENT_SUBSCRIPTION_MATCHED);
+                // printf(
+                //   "Updating matched count by 1 for local sub: %s\n",
+                //   sub_entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+      topic_info.type_);
+    if (topic_type_map_it == topic_map_it->second.end()) {
+      // First time this topic type is added.
+      // TODO(Yadunund) Check for and report an *_INCOMPATIBLE_TYPE events.
+
+      topic_map_it->second.insert(
+        std::make_pair(
+          topic_info.type_,
+          std::move(topic_qos_map)));
+    } else {
```
clalancette: This entire function is pretty large and deeply nested.  Is there a way to refactor this to avoid that somehow?  Split this into functions, maybe?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+            // Update event counters for the new entity.
+            update_event_counters(
+              topic_info.name_,
+              ZENOH_EVENT_PUBLICATION_MATCHED,
+              match_count_for_entity);
+            if (is_entity_local(entity) && match_count_for_entity > 0) {
+              local_entities_with_events[entity].insert(ZENOH_EVENT_PUBLICATION_MATCHED);
+              // printf(
+              //   "Updating matched count by %ld for local pub: %s\n",
+              //   match_count_for_entity, entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+            for (const liveliness::Entity & pub_entity : topic_data_ptr->pubs_) {
+              update_event_counters(
+                topic_info.name_,
+                ZENOH_EVENT_PUBLICATION_MATCHED,
+                static_cast<int32_t>(1));
+              if (is_entity_local(pub_entity)) {
+                local_entities_with_events[pub_entity].insert(ZENOH_EVENT_PUBLICATION_MATCHED);
+                // printf(
+                //   "Updating matched count by 1 for local pub: %s\n",
+                //   pub_entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+            // Update event counters for the new entity.
+            update_event_counters(
+              topic_info.name_,
+              ZENOH_EVENT_SUBSCRIPTION_MATCHED,
+              match_count_for_entity);
+            if (is_entity_local(entity) && match_count_for_entity > 0) {
+              local_entities_with_events[entity].insert(ZENOH_EVENT_SUBSCRIPTION_MATCHED);
+              // printf(
+              //   "Updating matched count by %ld for local sub: %s\n",
+              //   match_count_for_entity, entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
-        // Token was for a node.
-        return graph_node;
+///=============================================================================
+void GraphCache::take_local_entities_with_events(
+  std::unordered_map<liveliness::Entity, std::unordered_set<rmw_zenoh_event_type_t>> &
+  local_entities_with_events)
+{
+  if (local_entities_with_events.empty()) {
+    return;
+  }
```
clalancette: This doesn't seem to be strictly necessary; if there are no events, then we just won't enter into the loop below.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -41,170 +41,284 @@ using Entity = liveliness::Entity;
...
+}
 
-      return graph_node;
-    };
+///=============================================================================
+void GraphCache::parse_put(
+  const std::string & keyexpr,
+  bool ignore_from_current_session)
+{
+  // printf("[parse_put %s] %s\n", zid_str_.c_str(), keyexpr.c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -220,180 +334,235 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+      cache_topic_it->second.find(topic_info.type_);
+    if (cache_topic_type_it == cache_topic_it->second.end()) {
+      // This should not happen.
+      RCUTILS_LOG_ERROR_NAMED(
+        "rmw_zenoh_cpp", "topic type %s not found in for topic %s. Report this.",
+        topic_info.type_.c_str(), topic_info.name_.c_str());
+      return;
+    } else {
+      const std::string qos_str = liveliness::qos_to_keyexpr(topic_info.qos_);
+      GraphNode::TopicQoSMap::iterator cache_topic_qos_it = cache_topic_type_it->second.find(
```
clalancette: I have the same basic thought about this function; it is long and pretty deeply nested.  Can we refactor it somehow?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -220,180 +334,235 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+            for (const liveliness::Entity & sub_entity : topic_data_ptr->subs_) {
+              update_event_counters(
+                topic_info.name_,
+                ZENOH_EVENT_SUBSCRIPTION_MATCHED,
+                static_cast<int32_t>(-1));
+              if (is_entity_local(sub_entity)) {
+                local_entities_with_events[sub_entity].insert(ZENOH_EVENT_SUBSCRIPTION_MATCHED);
+                // printf(
+                //   "Updating matched count by -1 for local sub: %s\n",
+                //   sub_entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -220,180 +334,235 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+            for (const liveliness::Entity & pub_entity : topic_data_ptr->pubs_) {
+              update_event_counters(
+                topic_info.name_,
+                ZENOH_EVENT_PUBLICATION_MATCHED,
+                static_cast<int32_t>(-1));
+              if (is_entity_local(pub_entity)) {
+                local_entities_with_events[pub_entity].insert(ZENOH_EVENT_PUBLICATION_MATCHED);
+                // printf(
+                //   "Updating matched count by -1 for pub sub: %s\n",
+                //   pub_entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -220,180 +334,235 @@ void GraphCache::parse_put(const std::string & keyexpr)
...
+    }
+  }
+}
 
+///=============================================================================
+void GraphCache::parse_del(
+  const std::string & keyexpr,
+  bool ignore_from_current_session)
+{
+  // printf("[parse_del %s] %s\n", zid_str_.c_str(), keyexpr.c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -996,11 +1195,114 @@ rmw_ret_t GraphCache::service_server_is_available(
...
+  if (event_cb_it == event_callbacks_.end()) {
+    // First time a callback is being set for this entity.
+    event_callbacks_[entity] = {};
+    event_callbacks_[entity].insert(std::make_pair(event_type, std::move(callback)));
+    return;
+  }
+  event_cb_it->second[event_type] = std::move(callback);
+  // printf(
+  //   "[graph_cache] Set callback for rmw_zenoh_event_type_t %s for entity %s\n",
+  //   std::to_string(event_type).c_str(), entity.keyexpr().c_str());
```
clalancette: Remove debugging code.

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -36,43 +38,42 @@
...
-  std::size_t sub_count_;
-
-  // Constructor which initializes counters to 0.
-  TopicStats(std::size_t pub_count, std::size_t sub_count);
-};
-
-///=============================================================================
+// Consider changing this to an array of unordered_set where the index of the
+// array corresponds to the EntityType enum. This way we don't need to mix
+// pub/sub with client/service.
```
clalancette: Maybe change this to a `TODO` comment?  Because I also agree that refactoring it in this way would be clearer.

clalancette (changes requested):

rmw_zenoh_cpp/src/detail/event.cpp
Yadunund: The header is included in `event.hpp`

rmw_zenoh_cpp/src/detail/event.cpp
Yadunund: The header is already included in `event.hpp`.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/103/commits/d723fb6332bcb04046bb80657b588141ad132037

rmw_zenoh_cpp/src/detail/event.cpp
Yadunund: Good point. After the latest split of `EventsBase` into `DataCallbackManger` and `EventsManager`, we can avoid deadlocks with a regular `mutex`. Updated https://github.com/ros2/rmw_zenoh/pull/103/commits/e51e475760347bbd6dcb4d79b7a349e9264a0af4

rmw_zenoh_cpp/src/detail/event.hpp
Yadunund: I see. If we adopt this approach, the array lengths of these [members](https://github.com/ros2/rmw_zenoh/blob/e51e475760347bbd6dcb4d79b7a349e9264a0af4/rmw_zenoh_cpp/src/detail/event.hpp#L159-L170) would now be `ZENOH_EVENT_INVALID`, eg `std::condition_variable * event_conditions_[ZENOH_EVENT_INVALID]{nullptr}` which presents a readability problem imo. If you think it's valid, I'm happy to make the change as suggested.

rmw_zenoh_cpp/src/detail/graph_cache.hpp
Yadunund: https://github.com/ros2/rmw_zenoh/pull/103/commits/f768e32ef42e0c1e9c3ad7769f1c33b98c9344be

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: `TopicData` is not taking ownership of existing objects. Instead, it is being initialized with copies. The initializer list uses `std::move` to avoid a second copy but maybe the compiler is smart enough to perform this optimization?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Hoping this makes things a bit cleaner https://github.com/ros2/rmw_zenoh/pull/103/commits/00e8e1c2929849e33913161b6d442e55050457d7

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Agreed. I've simplified things here https://github.com/ros2/rmw_zenoh/pull/103/commits/47973be62df1ea92f5387f9a7c5d5d5caec1e227

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/103/commits/47973be62df1ea92f5387f9a7c5d5d5caec1e227

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/103/commits/47973be62df1ea92f5387f9a7c5d5d5caec1e227

Yadunund:
@clalancette this is ready for another review.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Could we save it for another ticket? I've opened https://github.com/ros2/rmw_zenoh/issues/132 to keep track.

clalancette (approved):

PR #106, opened 2024-02-08
Fix a number of small issues in rmw_zenoh_cpp

clalancette:
This PR is a smorgasbord of small fixes around the tree.  It is probably easiest to look at the individual patches, but in short:

* Add a small optimization where we keep a running total of the number of nodes in the graph.  This speeds up `get_nodes_names` for large graphs, since we don't have to iterate over the namespaces anymore.
* While doing the above, I noticed that we never removed namespaces once they were added.  I added a patch so that if there are no nodes left in the namespace after this one is removed, we also remove the namespace.
* I removed a few old TODO comments.
* I removed the PublishToken implementation.  While it is somewhat nice to have, it is a bit messy and I think we are going to have to do something more substantial when we try to support zenoh-pico.  And this code is still in the git history.
* Remove some hand-coded `rcutils_strdup` implementations.  The original reason for that hand-coding is gone, so we can just use `rcutils_strdup` now.
* Speed up `ros_topic_name_to_zenoh_key`.  In particular, `std::to_string` is much faster than `std::stringstream`, so use that.
* Use RAII for saved_msg_data.  That way we don't have to remember to `z_drop` the sample.
* Implement `rmw_compare_gids_equal`, which just gets us one step closer to having a complete implementation.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -459,6 +459,9 @@ void GraphCache::parse_del(const std::string & keyexpr)
     }
     ns_it->second.erase(node_it);
     total_nodes_in_graph_ -= 1;
+    if (ns_it->second.size() == 0) {
+      graph_.erase(entity.node_namespace());
+    }
     return;
   }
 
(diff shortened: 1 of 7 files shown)


Yadunund (approved):
Thanks for the cleanup!

PR #108, opened 2024-02-09
Update the design document

clalancette:
This PR does a large rewrite of the design document, adding in a lot more detail about how rmw_zenoh_cpp conforms to the RMW interface.

I'll note that this document isn't complete; in particular, the section on how Events works is incomplete (until we get #103 in), and the section on Security is completely empty.  Nevertheless, I think it expand quite a bit on what we currently have and will give users some idea on the choices we made and why.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -9,7 +9,7 @@ The end result is that users can use ROS 2 to send and receive data over Zenoh, 
 
 There is more detail on each item below, but a brief overview on how this is accomplished is the following:
 
-* It is assumed that Zenoh a router is running on the local system.  This router will be used for discovery and host-to-host communication.  However it is *not* used for intra-host comms; that is done via direct peer-to-peer connections.
+* It is assumed that a Zenoh router is running on the local system.  This router will be used for discovery and host-to-host communication.  However it is *not* used for intra-host comms (i.e., as a message broker); that is done via direct peer-to-peer connections.
 * Each "context" in ROS 2 is mapped to a single Zenoh "session".  That means that there may be many publishers, subscriptions, services, and clients sharing the same session.
 * Every "context" has a local "graph cache" that keeps track of the details of the network graph of ROS 2 entities.
 * Zenoh publishers, subscriptions, services, and clients are created or destroyed when the corresponding RMW APIs are called.
@@ -71,7 +71,7 @@ However, local multicast has some limitations, both intrinsic and specific to Ze
 * Multicast discovery can cause a lot of discovery traffic while discovering all other entities in the graph.
 * Multicast discovery has a limited TTL (time-to-live), which means it can usually only discover peers on the local network segment.
 
-For the reasons cited above, `rmw_zenoh_cpp` requires a Zenoh router to be running.
+To alleviate issues with multicast discovery, `rmw_zenoh_cpp` relies on a Zenoh router to discover peers and forward this discovery information to other peers via Zenoh's `gossip scouting` functionality. Hence `rmw_zenoh_cpp` requires the Zenoh router to be running.
 
 It should be noted that when building upstream Zenoh from source, a `zenohd` binary is created which is the router.
 `rmw_zenoh_cpp` actually has its own simplified version of the router that nonetheless uses most of the same code.
@@ -106,7 +106,7 @@ Zenoh attempts to do the absolute minimum of discovery (for performance reasons)
 To deal with this discrepancy, each context in `rmw_zenoh_cpp` keeps a cache of all entities discovered in the graph so far.
 An "entity" is a node, publisher, subscription, service server, or service client.
 Each entity sends a unique liveliness token as it comes online, and removes that liveliness token when it is destroyed.
-These tokens contain information about the entity and it's relationship to the other entities in the system (for instance, a publisher is always attached to a node).
+The key expression of these liveliness tokens encode information about the entity and it's relationship to the other entities in the system (for instance, a publisher is always attached to a node within a certain namespace).
 
 ### Related RMW APIs
 
@@ -206,6 +206,7 @@ When a new publisher is created, a liveliness token of type `MP` is sent out.
 
 * zc_liveliness_declare_token
 * zc_publish_put_owned
+* ze_declare_publication_cache
 * z_declare_publisher
 * z_undeclare_publisher
 * z_publisher_put
@@ -241,6 +242,7 @@ When a new subscription is created, a liveliness token of type `MS` is sent out.
 
 * zc_liveliness_declare_token
 * zc_sample_payload_rcinc
+* ze_declare_querying_subscriber
 * z_declare_subscriber
 * z_undeclare_subscriber
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -71,13 +71,18 @@ However, local multicast has some limitations, both intrinsic and specific to Ze
 * Multicast discovery can cause a lot of discovery traffic while discovering all other entities in the graph.
 * Multicast discovery has a limited TTL (time-to-live), which means it can usually only discover peers on the local network segment.
 
-To alleviate issues with multicast discovery, `rmw_zenoh_cpp` relies on a Zenoh router to discover peers and forward this discovery information to other peers via Zenoh's `gossip scouting` functionality. Hence `rmw_zenoh_cpp` requires the Zenoh router to be running.
+To alleviate issues with multicast discovery, `rmw_zenoh_cpp` relies on a Zenoh router to discover peers and forward this discovery information to other peers via Zenoh's `gossip scouting` functionality.
+Hence `rmw_zenoh_cpp` requires the Zenoh router to be running.
 
 It should be noted that when building upstream Zenoh from source, a `zenohd` binary is created which is the router.
 `rmw_zenoh_cpp` actually has its own simplified version of the router that nonetheless uses most of the same code.
 This was done so that Zenoh didn't have to be vendored twice (once for zenoh-c and once for zenohd), and so that the router could be more easily integrated into the ROS 2 package format.
 
 As of 2024-02-09, the user is expected to launch the router by hand.
+In order to integrate more seamlessly into ROS 2, the Zenoh router can be launched by running `ros2 run rmw_zenoh_cpp rmw_zenohd`.
+The default configuration that the router uses is located in [DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5](../rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5).
+The user can use a custom configuration by setting the `ZENOH_ROUTER_CONFIG_URI` environment variable to point to a valid [Zenoh configuration file](https://github.com/eclipse-zenoh/zenoh/blob/main/DEFAULT_CONFIG.json5).
+
 In the future, additional work may be done to automatically launch a Zenoh router if one isn't already running.
 
 ## Serialization/Deserialization
@@ -140,6 +145,10 @@ The key expression of these liveliness tokens encode information about the entit
 
 A ROS 2 context describes a certain middleware configuration, which can contain 0 or more ROS 2 nodes.
 In `rmw_zenoh_cpp`, a context maps to a Zenoh session, along with a liveliness token for the graph cache and some additional metadata.
+
+Zenoh allows sessions to be custom configured through a configuration file.
+If otherwise unconfigured, `rmw_zenoh_cpp` uses a [default configuration file](../rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5).
+The user can use a custom configuration by setting the `ZENOH_SESSION_CONFIG_URI` environment variable to point to a valid [Zenoh configuration file](https://github.com/eclipse-zenoh/zenoh/blob/main/DEFAULT_CONFIG.json5).
 
 ### Related RMW APIs
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,6 +1,19 @@
 # Design
 
-## Network Topology
+## Introduction
+
+`rmw_zenoh_cpp` maps the ROS 2 [RMW API](https://github.com/ros2/rmw/tree/rolling/rmw/include/rmw) as of late 2023 onto Zenoh APIs, using [zenoh-c](https://github.com/eclipse-zenoh/zenoh-c).
+The end result is that users can use ROS 2 to send and receive data over Zenoh, using the APIs that they are already familiar with.
+
+## Brief overview
+
+There is more detail on each item below, but a brief overview on how this is accomplished is the following:
+
+* It is assumed that Zenoh a router is running on the local system.  This router will be used for discovery and host-to-host communication.  However it is *not* used for intra-host comms; that is done via direct peer-to-peer connections.
+* Each "context" in ROS 2 is mapped to a single Zenoh "session".  That means that there may be many publishers, subscriptions, services, and clients sharing the same session.
+* Every "context" has a local "graph cache" that keeps track of the details of the network graph of ROS 2 entities.
+* Zenoh publishers, subscriptions, services, and clients are created or destroyed when the corresponding RMW APIs are called.
+* Data is sent and received through the appropriate zenoh-c API when the corresponding RMW APIs are called.
 
 The following diagram shows the default network topology of a subsystem composed of 3 nodes:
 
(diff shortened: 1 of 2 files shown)


docs/design.md
```diff
@@ -1,6 +1,19 @@
...
+## Introduction
+
+`rmw_zenoh_cpp` maps the ROS 2 [RMW API](https://github.com/ros2/rmw/tree/rolling/rmw/include/rmw) as of late 2023 onto Zenoh APIs, using [zenoh-c](https://github.com/eclipse-zenoh/zenoh-c).
+The end result is that users can use ROS 2 to send and receive data over Zenoh, using the APIs that they are already familiar with.
+
+## Brief overview
+
+There is more detail on each item below, but a brief overview on how this is accomplished is the following:
+
+* It is assumed that Zenoh a router is running on the local system.  This router will be used for discovery and host-to-host communication.  However it is *not* used for intra-host comms; that is done via direct peer-to-peer connections.
```
Yadunund:
```suggestion
* It is assumed that a Zenoh router is running on the local system.  This router will be used for discovery and host-to-host communication.  However it is *not* used for intra-host comms (i.e., as a message broker); that is done via direct peer-to-peer connections.
```

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
 
+## Router
+
+Zenoh has the ability to do discovery using local multicast announcements.
+However, local multicast has some limitations, both intrinsic and specific to Zenoh:
+
+* Multicast discovery can cause a lot of discovery traffic while discovering all other entities in the graph.
+* Multicast discovery has a limited TTL (time-to-live), which means it can usually only discover peers on the local network segment.
+
+For the reasons cited above, `rmw_zenoh_cpp` requires a Zenoh router to be running.
```
Yadunund:
```suggestion
To alleviate issues with multicast discovery, `rmw_zenoh_cpp` relies on a Zenoh router to discover peers and forward this discovery information to other peers via Zenoh's `gossip scouting` functionality. Hence `rmw_zenoh_cpp` requires the Zenoh router to be running.
```

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+N/A
+
+## Graph Cache
+
+One of the biggest impedance mismatches between Zenoh and ROS 2 has to do with graph introspection.
+Zenoh attempts to do the absolute minimum of discovery (for performance reasons), while ROS 2 generally assumes that the entire graph is available from all entities in the system (for debugging/introspection reasons).
+To deal with this discrepancy, each context in `rmw_zenoh_cpp` keeps a cache of all entities discovered in the graph so far.
+An "entity" is a node, publisher, subscription, service server, or service client.
+Each entity sends a unique liveliness token as it comes online, and removes that liveliness token when it is destroyed.
+These tokens contain information about the entity and it's relationship to the other entities in the system (for instance, a publisher is always attached to a node).
```
Yadunund:
```suggestion
The key expression of these liveliness tokens encode information about the entity and it's relationship to the other entities in the system (for instance, a publisher is always attached to a node within a certain namespace).
```

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+
+For the reasons cited above, `rmw_zenoh_cpp` requires a Zenoh router to be running.
+
+It should be noted that when building upstream Zenoh from source, a `zenohd` binary is created which is the router.
+`rmw_zenoh_cpp` actually has its own simplified version of the router that nonetheless uses most of the same code.
+This was done so that Zenoh didn't have to be vendored twice (once for zenoh-c and once for zenohd), and so that the router could be more easily integrated into the ROS 2 package format.
+
+As of 2024-02-09, the user is expected to launch the router by hand.
+In the future, additional work may be done to automatically launch a Zenoh router if one isn't already running.
+
```
Yadunund, line 87: Should we mentioned a couple lines about the default router config and how it is configured to disable multicast scouting and enables gossip scouting instead?

Yadunund: As a follow up, should we include a section on the session config?

docs/design.md
Yadunund: Should we copy over documentation from https://github.com/ros2/rmw_zenoh/blob/ab2877d3b22f4f8f52ae2a682caaa8f3dab1a3ba/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L229-L257 and https://github.com/ros2/rmw_zenoh/blob/ab2877d3b22f4f8f52ae2a682caaa8f3dab1a3ba/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp#L128-L140 here? Particularly to give more context to the `NN` mentioned in the `Nodes` section below.

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+### Related Zenoh-c APIs
+
+N/A
+
+## Graph Cache
+
+One of the biggest impedance mismatches between Zenoh and ROS 2 has to do with graph introspection.
+Zenoh attempts to do the absolute minimum of discovery (for performance reasons), while ROS 2 generally assumes that the entire graph is available from all entities in the system (for debugging/introspection reasons).
+To deal with this discrepancy, each context in `rmw_zenoh_cpp` keeps a cache of all entities discovered in the graph so far.
+An "entity" is a node, publisher, subscription, service server, or service client.
```
Yadunund, line 112: Should we highlight the `zc_liveliness_get()` call we make within `rmw_init()` to querry for all liveliness token from existing entities when a new session starts?

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* z_undeclare_subscriber
+* z_call
+* z_session_check
+
+## Nodes
+
+A ROS 2 node can be though of as the "unit of computation" in a ROS 2 graph; usually one node performs one particular task.
+Nodes may contain publishers, subscriptions, service servers, service clients, action servers, action clients, parameters, and anything else needed to do some computation.
+Zenoh has no conceptual equivalent to the ROS 2 node, so `rmw_zenoh_cpp` creates no Zenoh entities when nodes are created.
+When a new node is created through the RMW API, a liveliness token of type `NN` is sent out.
```
Yadunund, line 209: Should we highlight namespaces here?

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+### Related Zenoh-c APIs
+
+* zc_liveliness_declare_token
+
+## Publishers
+
+A ROS 2 publisher sends data to 0 or more connected subscriptions.
+A Zenoh publisher does exactly the same thing, so ROS 2 publishers are mapped onto Zenoh publishers in `rmw_zenoh_cpp`.
+When a new publisher is created, a liveliness token of type `MP` is sent out.
+
```
Yadunund, line 229: Should we add a TODO for attachments that we aim to use for GUID, seq number, time stamp, etc? And include attachment related APIs below. Same for other relevant entities.

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* z_publisher_put
+
+## Subscriptions
+
+A ROS 2 subscription receives data from 1 or more connected publishers.
+A Zenoh subscriber does exactly the same thing, so ROS 2 subscriptions are mapped onto Zenoh subscribers in `rmw_zenoh_cpp`.
+When new data arrives, a callback within `rmw_zenoh_cpp` is executed, which takes ownership of the data and signals that there is data available.
+Then rmw_wait can find out that there is data available, and the data can be delivered via rmw_take.
+When a new subscription is created, a liveliness token of type `MS` is sent out.
+
```
Yadunund, line 261: Should we highlight whether we are making any copies?

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* zc_liveliness_declare_token
+* zc_sample_payload_rcinc
+* z_declare_subscriber
+* z_undeclare_subscriber
+
+## Service Clients
+
+In ROS 2, services are meant to be used for remote procedure calls that will return fairly quickly.
+`rmw_zenoh_cpp` uses Zenoh queryables to implement ROS 2 services.
```
Yadunund, line 292: Should we highlight how we only query "Complete" keys?

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+When a client wants to make a request, it uses the rmw API `rmw_send_request`.
+`rmw_zenoh_cpp` then calls the Zenoh `z_get` function to send a query out to the network.
+Assuming there is a service server listening to that queryable, it will receive the request, perform a computation, and return the result.
+The result will then be made available to the client via `rmw_take_response`.
+When a new service client is created, a liveliness token of type `SC` is sent out.
```
Yadunund, line 301: Highlight how attachment is used for tracking sequence numbers?

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* rmw_take_loaned_message_with_info
+* rmw_return_loaned_message_from_subscription
+* rmw_subscription_get_network_flow_endpoints
+* rmw_subscription_event_init
+
+### Related Zenoh-c APIs
+
+* zc_liveliness_declare_token
+* zc_sample_payload_rcinc
+* z_declare_subscriber
```
Yadunund, line 286:
```suggestion
* ze_declare_querying_subscriber
* z_declare_subscriber
```

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* rmw_return_loaned_message
+* rmw_publisher_wait_for_all_acked
+* rmw_publisher_get_network_flow_endpoints
+* rmw_publisher_event_init
+
+### Related Zenoh-c APIs
+
+* zc_liveliness_declare_token
+* zc_publish_put_owned
+* z_declare_publisher
```
Yadunund, line 247:
```suggestion
* ze_declare_publication_cache
* z_declare_publisher
```

docs/design.md
Yadunund: Also how we instantiate a `ze_owned_publication_cache_t` only when QoS durability is transient_local

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* z_undeclare_publisher
+* z_publisher_put
+
+## Subscriptions
+
+A ROS 2 subscription receives data from 1 or more connected publishers.
+A Zenoh subscriber does exactly the same thing, so ROS 2 subscriptions are mapped onto Zenoh subscribers in `rmw_zenoh_cpp`.
```
Yadunund, line 254: Highlight mapping to `z_owned_subscriber_t` vs `ze_owned_querying_subscriber_t` depending on QoS durability

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* DEPTH - The maximum number of samples to keep; only comes into play when KEEP_LAST history is used.
+* DURABILITY
+    * VOLATILE - Samples will only be delivered to subscriptions that are active at the time of publishing.  In `rmw_zenoh_cpp`, this is implemented via `z_declare_subscriber` on the subscription side and `z_declare_publisher` on the publisher side.
+    * TRANSIENT_LOCAL - "Late-joining" subscriptions will receive historical data, along with any new data.  In `rmw_zenoh_cpp`, this is implemented via `ze_declare_querying_subscriber` on the subscription side and `ze_declare_publication_cache` on the publisher side.
+* LIVELINESS
+    * AUTOMATIC - The "liveliness" of an entity of the system is managed by the RMW layer.  This is what `rmw_zenoh_cpp` uses.
+    * MANUAL_BY_TOPIC - It is up to the application to periodically publish to a particular topic to assert liveliness.
+* DEADLINE - The period at which messages are expected to be sent/received.  Currently unimplemented in `rmw_zenoh_cpp`.
+* LIFESPAN - The age at which messages are expired and no longer valid.  Currently unimplemented in `rmw_zenoh_cpp`.
+
```
Yadunund, line 377: Should we add a small section for QoS Compatibility? For now we ascertain compatibility akin to DDS although Zenoh transport can handle certain DDS incompatibilities.

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+* z_undeclare_queryable
+* z_query_value
+* z_query_attachment
+
+## Quality of Service (QoS)
+
+The ROS 2 RMW layer defines quite a few Quality of Service settings that are largely derived from DDS.
+Here is an incomplete list of some of the settings and the values that they can take:
+
+* RELIABILITY
```
Yadunund, line 362: We should highlight that reliability is something we can only set for subs in Zenoh. If Reliability is RELIABLE and History is KEEP_ALL for pubs, we set Congestion to BLOCK. Else it is DROP.

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+    * RELIABLE - Data delivery is retried until it is successfully delivered.  Because Zenoh is TCP-based (by default), this is the default in `rmw_zenoh_cpp`.
```
Yadunund: I think we need to be a bit careful here been reliability from TCP transport vs reliability QoS settings. By default, Zenoh subscriptions has [BestEffort](https://github.com/eclipse-zenoh/zenoh/blob/7ebdb3cab481609989f1e9d3124e75d2851458e5/commons/zenoh-protocol/src/core/mod.rs#L351-L352) Reliability.
```suggestion
    * RELIABLE - Applicable only for subscriptions. Data delivery is retried until it is successfully delivered.  The `SYSTEM_DEFAULT` reliability in `rmw_zenoh_cpp` is `BestEffort`.
```

docs/design.md
```diff
@@ -50,6 +63,328 @@ Default Configuration for Zenoh Sessions:
...
+### Related Zenoh-c APIs
+
+* zc_liveliness_declare_token
+* z_attachment_get
+* z_declare_queryable
+* z_undeclare_queryable
+* z_query_value
+* z_query_attachment
+
+## Quality of Service (QoS)
```
Yadunund: TODO: highlight what the SYSTEM_DEFAULT behavior is for each QoS setting.

Yadunund (changes requested):
Thanks for taking the time to document all this valuable information! Left some feedback comments (happy to push changes directly if you're ok too). In addition, I think we should delete [this section](https://github.com/ros2/rmw_zenoh#config) from the readme since it is expanded upon here.

docs/design.md
clalancette: I'm going to go ahead and add a section for both of these.

While looking at this, I realized that our environment variables are named `ZENOH_ROUTER_CONFIG_URI` and `ZENOH_SESSION_CONFIG_URI`.  But we don't treat them as URIs, we treat them as filenames.  Should we change that somehow?

docs/design.md
clalancette: > Should we add a TODO for attachments that we aim to use for GUID, seq number, time stamp, etc? And include attachment related APIs below. Same for other relevant entities.

I don't think that we need to encode missing features into the design.  When we add that stuff in, we can update the design document to add it as well.

> Also how we instantiate a `ze_owned_publication_cache_t` only when QoS durability is transient_local

Good idea, I'll add some language around this.

docs/design.md
clalancette: > Should we highlight namespaces here?

Yes, good idea.  I'll actually add a separate section discussing namespaces and how they relate between ROS 2 and Zenoh.

docs/design.md
clalancette: > Should we highlight whether we are making any copies?

It is already implicitly there by saying `takes ownership of the data`.

We could add some more explicit language here, but I'm inclined not to do it.  I don't want to get too bogged down in zero-copy discussions, because we clearly aren't.

docs/design.md
clalancette: I still only have a cursory knowledge of what that means.  Could you suggest some language that I could add here?

docs/design.md
clalancette: Good idea.

docs/design.md
clalancette: Done in 051c630

docs/design.md
clalancette: Cleaned up in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645), though I split it up a bit differently.

docs/design.md
clalancette: I think this should be resolved in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645), but please take a look.

docs/design.md
clalancette: I added a note saying that Zenoh never fails QoS matching, which I think matches our current understanding.

docs/design.md
clalancette: Done in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645)

docs/design.md
clalancette: Done in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645)

docs/design.md
clalancette: Done in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645)

docs/design.md
clalancette: Done in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645)

docs/design.md
clalancette: Good idea.  Done in [051c630](https://github.com/ros2/rmw_zenoh/pull/108/commits/051c63036d22261b2dd998624f27f37fb39db645)

clalancette:
@Yadunund I've responded to most of your feedback now.  There are a couple of pieces of information that I don't have, so I didn't update those.  If you can take another look, that would be appreciated.

docs/design.md
Yadunund: The envars should be set to the absolute path of the resource and not just the filename as documented [here](https://github.com/ros2/rmw_zenoh/blob/191a56b5799a34a778dacf6883cc9ebb70d0e4f9/rmw_zenoh_cpp/src/detail/zenoh_config.cpp#L28-L29) and implemented [here](https://github.com/ros2/rmw_zenoh/blob/191a56b5799a34a778dacf6883cc9ebb70d0e4f9/rmw_zenoh_cpp/src/detail/zenoh_config.cpp#L55). So imo URI is valid but happy to change it something more appropriate if necessary.

Yadunund (approved):
This looks awesome!

PR #109, opened 2024-02-09
More cleanups around the code.

clalancette:
As I was creating the design document, I noticed a few things around the code that could be cleaned up and made faster.  This PR:

1.  Switches to using a keyexpression slice for creating keyexpressions.  That way, we don't have to copy a string just to remove the final '/'.
2.  Makes sure that graph_sub_data_handler drops the keyexpression on all paths.  In particular, on error paths we would forget to drop the keyexpression, so just use a `make_scope_exit`.
3.  Remove another unnecessary comment that has to do with porting to zenoh-pico.
4.  Remove an unnecessary use of malloc/memcpy/free when creating a map to use as an attachment for service calls.  Since `z_bytes_map_insert_by_copy` will do a copy anyway, just pass in the array.
5.  When doing `rmw_destroy_service`, call `z_undeclare_queryable` to drop the queryable.  I believe that this is functionally equivalent to `z_drop`, but I also think it is more self-documenting to use `z_undeclare_queryable` here.

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_init.cpp
+++ b/rmw_zenoh_cpp/src/rmw_init.cpp
@@ -269,12 +269,6 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
   zc_liveliness_get(
     z_loan(context->impl->session), z_keyexpr(liveliness_str.c_str()),
     z_move(channel.send), NULL);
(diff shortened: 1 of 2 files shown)


Yadunund (approved):
Thanks for the additional cleanup!

Issue #102, opened 2024-01-25
Refactor zenoh_config.hpp so we can reuse it in the router


clalancette:
We have code in https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/zenoh_config.hpp and https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/zenoh_config.cpp that is currently used to allow the user to configure the *session* configuration when using `rmw_zenoh_cpp`.

We should also use that code to be able to configure *router* configuration as well.  That way, users will be able to launch the router locally with either the default config, or one they provide.  The code probably needs some refactoring to be able to use it for that purpose.

This is a follow-up task from #101 .

Yadunund:
Closed via https://github.com/ros2/rmw_zenoh/pull/119

PR #112, opened 2024-02-16
Allow connections between routers

Yadunund:
Fix the default router config to listen on all connections established to the `7447` port. 

Will add documentation on how to connect two routers, as part of the fix for https://github.com/ros2/rmw_zenoh/issues/102


clalancette (commented):
I'm hesitant about this one, from a security perspective.  That is, with the current configuration, we are sure that nothing outside the current machine can possibly connect.  If we enable this by default, then by default anyone can connect and gain access to the box.

Now, this situation isn't much different from DDS.  But it still feels a little scary to do this by default.  @codebot do you have any additional thoughts here?

codebot (approved):
LGTM; this causes the default behavior to align with ROS 1 and most other `rmw_*` in ROS 2.

PR #113, opened 2024-02-18
Implement __rmw_take_serialized.

clalancette:
This allows us to implement rmw_take_serialized_message and rmw_take_serialize_message_with_info, both of which are required for rosbag2.

Note that this implementation copies some parts of __rmw_take. Given the different types involved (void * ros_message vs. rmw_serialize_message_t *), I didn't see a readable way to reduce this duplication.

Tested by running `ros2 bag record -a`.  Before this PR, that would throw many errors, while after this PR it works as expected and can play back data.


Yadunund (approved):
LGTM!

PR #119, opened 2024-02-27
Make zenoh router configurable using envar

Yadunund:
This PR
- Allows users to specify a custom the `router` configuration file via the `ZENOH_ROUTER_CONFIG_URI` envar.
- The envar to specify the `session` config file is now `ZENOH_SESSION_CONFIG_URI`
- The `router` executable is now `zenohd` to align with that of upstream Zenoh.
- Readme updated to inform users how to configure these files.

Addresses https://github.com/ros2/rmw_zenoh/issues/102


clalancette (commented):
I didn't do a full review here, but I did want to discuss this point:

```
The router executable is now zenohd to align with that of upstream Zenoh.
```

I'm not so sure that is a great idea.  The fact is that while this is similar to the upstream `zenohd`, it is not exactly the same and doesn't support all of the same things (in particular, plugins).  So I do think we should make it clear that this is the `rmw_zenoh_cpp` version of this.  I know that `ros2 run rmw_zenoh_cpp zenohd` does that, but:

1.  You don't have to run it that way; you can also do /path/to/zenohd, and then it looks just like the upstream one.
2.  In the process listing, it will show up as `zenohd`.  Yes, the full executable path will be to an rmw_zenoh-specific path, but it is an easy detail to overlook.

My suggestion is that we do indeed rename it, but to something else.  Something like `rmw_zenohd` or something like?

Yadunund:
> I didn't do a full review here, but I did want to discuss this point:
> 
> ```
> The router executable is now zenohd to align with that of upstream Zenoh.
> ```
> 
> I'm not so sure that is a great idea. The fact is that while this is similar to the upstream `zenohd`, it is not exactly the same and doesn't support all of the same things (in particular, plugins). So I do think we should make it clear that this is the `rmw_zenoh_cpp` version of this. I know that `ros2 run rmw_zenoh_cpp zenohd` does that, but:
> 
> 1. You don't have to run it that way; you can also do /path/to/zenohd, and then it looks just like the upstream one.
> 2. In the process listing, it will show up as `zenohd`.  Yes, the full executable path will be to an rmw_zenoh-specific path, but it is an easy detail to overlook.
> 
> My suggestion is that we do indeed rename it, but to something else. Something like `rmw_zenohd` or something like?

That makes a lot of sense to me. I've renamed the executable to `rmw_zenohd` 👍🏼

README.md
```diff
@@ -45,7 +45,7 @@ source install/setup.bash
 > Note: Manually launching zenoh router won't be necessary in the future.
 ```bash
 # terminal 1
-ros2 run rmw_zenoh_cpp init_rmw_zenoh_router
+ros2 run rmw_zenoh_cpp zenohd
```
clalancette:
```suggestion
ros2 run rmw_zenoh_cpp rmw_zenohd
```

README.md
```diff
@@ -68,26 +68,31 @@ ros2 run demo_nodes_cpp listener
 The listener node should start receiving messages over the `/chatter` topic.
 > Note: Ignore all the warning printouts.
 
-### Graph introspection
-Presently we only support limited `ros2cli` commands to introspect the ROS graph such as `ros2 node list` and `ros2 topic list`.
+## Configuration
+`rmw_zenoh` relies on separate configurations files to configure the Zenoh `router` and `session` respectively. For more information on the topology of Zenoh adopted in `rmw_zenoh`, please see [Design](#design). Default configuration files are used by `rmw_zenoh` however certain environment variables may be set to provide absolute paths to custom configuration files. The table below summarizes the default files and the environment variables for the `router` and `session`. For a complete list of configurable parameters, see [zenoh/DEFAULT_CONFIG.json5](https://github.com/eclipse-zenoh/zenoh/blob/main/DEFAULT_CONFIG.json5).
```
clalancette:
```suggestion
`rmw_zenoh` relies on separate configurations files to configure the Zenoh `router` and `session` respectively.
For more information on the topology of Zenoh adopted in `rmw_zenoh`, please see [Design](#design).
Default configuration files are used by `rmw_zenoh` however certain environment variables may be set to provide absolute paths to custom configuration files.
The table below summarizes the default files and the environment variables for the `router` and `session`.
For a complete list of configurable parameters, see [zenoh/DEFAULT_CONFIG.json5](https://github.com/eclipse-zenoh/zenoh/blob/main/DEFAULT_CONFIG.json5).
```

clalancette: It might also be worthwhile to call out what a `router` and `session` are, for those who aren't familiar with the Zenoh terminology and how rmw_zenoh use them.

README.md
```diff
@@ -68,26 +68,31 @@ ros2 run demo_nodes_cpp listener
...
-
-This assumes that there is a `zenohd` running in the system at port 7447.
-A custom configuration may be provided by setting the `RMW_ZENOH_CONFIG_FILE` environment variable to point to a custom zenoh configuration file.
+For example, to set the path to a custom `router` configuration file,
+```bash
+export ZENOH_ROUTER_CONFIG_URI=$HOME/MY_ZENOH_ROUTER_CONFIG.json5
+```
 
+### Connecting multiple hosts
+By default, all discovery traffic is local per host where host is the PC running a Zenoh `router`.
```
clalancette:
```suggestion
By default, all discovery traffic is local per host, where the host is the PC running a Zenoh `router`.
```

rmw_zenoh_cpp/src/detail/zenoh_config.hpp
```diff
@@ -17,18 +17,40 @@
...
+///==============================================================================
+/// Map the configurable entity to a pair of environment variable name that
+/// stores the absolute path to the Zenoh config and the default config filename.
+/// Note: The default config file should be located within rmw_zenoh_cpp/config/.
+static const std::unordered_map<ConfigurableEntity,
+  std::pair<const char *, const char *>> envar_map = {
+  {ConfigurableEntity::Session,
+    {"ZENOH_SESSION_CONFIG_URI", "DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5"}},
+  {ConfigurableEntity::Router, {"ZENOH_ROUTER_CONFIG_URI", "DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5"}}
+};
```
clalancette: I think this map can be put into `zenoh_config.cpp`.  That will have the big benefit of it not being included with every compilation unit.

clalancette (changes requested):
I've left a few pretty minor things to change.  Otherwise, this looks good to me.

README.md
Yadunund: I assume those terms will be clarified in the design document once your PR lands! Hence the link here to refer to that document

rmw_zenoh_cpp/src/detail/zenoh_config.hpp
Yadunund: https://github.com/ros2/rmw_zenoh/pull/119/commits/2594ddb2228ba91a1279036b349004320f13aef4

README.md
Yadunund: https://github.com/ros2/rmw_zenoh/pull/119/commits/dbc20c07e7699a14b5a097af11026379a47d3a96

clalancette (approved):

PR #121, opened 2024-03-04
As per #855 issue, use initializer list

shivangvijay:
https://github.com/ros-planning/moveit2/issues/855, Making code cleaner for rmw_zenoh repository.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/message_type_support.cpp
+++ b/rmw_zenoh_cpp/src/detail/message_type_support.cpp
@@ -24,11 +24,10 @@
 #include "message_type_support.hpp"
 #include "type_support_common.hpp"
 
-MessageTypeSupport::MessageTypeSupport(const message_type_support_callbacks_t * members)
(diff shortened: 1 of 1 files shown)


Yadunund:
@shivangvijay thanks for the PR but i'm afraid these are not valid changes. There is no member `name` to initialize and the original implementation has no issues imo. If you would like to contribute, please look at the Issues page for things to implement.

shivangvijay:
Sorry I found my implementation mistake. I wanted to start with a 'good first issue'. However, I couldn't test it locally because I have ROS2 Humble, and I encountered issues while building rmw_zenoh. I didn't invest much effort into resolving this issue in Humble because the documentation specifies that the repository only supports ROS2 Rolling and Iron. 

There are no issues with the original implementation; I simply wanted to make the code cleaner. I will create another 'good first issue'-related pull request. If you believe it will make the code cleaner, please accept it. I have started exploring rmw_zenoh repository issues.

PR #123, opened 2024-03-04
Attach sequence number, publisher GID, and source timestamp to public…

clalancette:
…ations.

That way, the subscriptions can pull them out of the attachment and pass it to the upper layers.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -14,26 +14,41 @@
 
 #include <zenoh.h>
 
+#include <condition_variable>
 #include <cstring>
+#include <memory>
 #include <mutex>
 #include <optional>
+#include <string>
 #include <utility>
 
 #include "rcpputils/scope_exit.hpp"
 #include "rcutils/logging_macros.h"
 
+#include "attachment_helpers.hpp"
 #include "rmw_data_types.hpp"
 
 ///==============================================================================
-saved_msg_data::saved_msg_data(zc_owned_payload_t p, uint64_t recv_ts, const uint8_t pub_gid[16])
-: payload(p), recv_timestamp(recv_ts)
+saved_msg_data::saved_msg_data(
+  zc_owned_payload_t p,
+  uint64_t recv_ts,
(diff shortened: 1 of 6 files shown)


rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -775,6 +776,48 @@ rmw_return_loaned_message_from_publisher(
   return RMW_RET_UNSUPPORTED;
 }
 
+static z_owned_bytes_map_t create_map_and_set_sequence_num(
```
Yadunund, line 779: Should we include this in an anonymous namespace?

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -232,6 +247,7 @@ std::unique_ptr<ZenohReply> rmw_client_data_t::pop_next_reply()
 }
 
 //==============================================================================
+
```
Yadunund: Nit: Unintended newline?

Yadunund (approved):
LGTM with minor comments to optionally address. The changes here will allow us to very easily detect `MESSAGE_LOST` events!

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
clalancette: Oops, good call.  Will fix that one.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: We aren't consistent here.  Sometimes we use the anonymous namespace, sometimes we use `static`.  They both accomplish the same thing.  I personally find the `static` to be easier to read (as you don't have to look at the namespace to find out whether it is a public function or not), but I don't feel strongly about it.

For now, I'm going to leave this as-is.  I'll open a separate issue to make this consistent throughout the codebase.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
clalancette: Fixed in fa6f257.

Issue #111, opened 2024-02-16
Perform slash replacement when converting ros topic names to zenoh key epressions


Yadunund:
[mangle_name](https://github.com/ros2/rmw_zenoh/blob/60c4380c9c6e43bbdde45f5092fcfeccdfc4fc6d/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp#L123) is currently used to replace `/` found in namespaced topic names, types, etc when generating keyexpressions for liveliness tokens.

We should invoke it within [ros_topic_name_to_zenoh_key](https://github.com/ros2/rmw_zenoh/blob/60c4380c9c6e43bbdde45f5092fcfeccdfc4fc6d/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L81) for better consistency.

Yadunund:
Closed by #128

PR #127, opened 2024-03-08
Added docker file and docker compose to dockerize rolling

shivangvijay:
I am working on issue #111, but I am unable to test it because my system only has ROS Humble, and the build is failing in that ROS version. Therefore, for testing purposes, I have dockerized the code for ROS Rolling. Let me know if this is helpful, and then I will update the README file with the Docker steps.

The output look like:-
Screenshot from 2024-03-08 19-15-10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM osrf/ros:rolling-desktop
+SHELL ["/bin/bash", "-c"]
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+RUN apt update
+COPY . /app/src/rmw_zenoh
+WORKDIR /app
+RUN rosdep install --from-paths src --ignore-src --rosdistro rolling -y
+RUN source /opt/ros/rolling/setup.bash && source "$HOME/.cargo/env" && colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
+
+SHELL ["/bin/bash", "-c"]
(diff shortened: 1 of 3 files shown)


clalancette:
Long-term, we aren't going to want to host a Dockerfile here; we have Dockerfiles for all of ROS 2 available, and we'll want `rmw_zenoh` to be part of that.  Short-term, we aren't quite ready for this kind of thing.  So I'm going to mark this is a draft for now (since it is in the same status as #68).

shivangvijay:
Ok @clalancette . But why the build pipeline failed? I didn't change any core stack just added 3 files (docker related).

clalancette:
> Ok @clalancette . But why the build pipeline failed? I didn't change any core stack just added 3 files (docker related).

That's because we are in the middle of a transition to Ubuntu 24.04, and are still working through some problems.  It has nothing to do with this pull request directly.

shivangvijay:
> > Ok @clalancette . But why the build pipeline failed? I didn't change any core stack just added 3 files (docker related).
> 
> That's because we are in the middle of a transition to Ubuntu 24.04, and are still working through some problems. It has nothing to do with this pull request directly.

Ok Ok. I don't see any ticket of this issue, If you raise that I can work on this, I have Ubuntu 24.04.

Yadunund:
With `rmw_zenoh` merged into `ros2.repos`, it should be available in the official OSRF docker images for Rolling and future distros. We don't want to host any Dockerfiles for older distros here so closing this one out.

PR #128, opened 2024-03-09
Solve of issue #111

shivangvijay:
Based on my understanding, I attempt to address issue #111. If it doesn't align with your expectations, please provide further clarification on this matter.


rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -122,6 +122,9 @@ class Entity
 /// Replace "/" instances with "%".
 std::string mangle_name(const std::string & input);
 
+/// Replace "/" instances with "%".
+char * mangle_name(const char * input);
```
Yadunund: There is no need to overload the `mangle_name` function. The objective of the ticket #111 is to reuse existing methods (or modify if needed) within `ros_topic_name_to_zenoh_key`

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -79,20 +79,22 @@ namespace
...
 
   size_t start_offset = 0;
   size_t topic_name_len = strlen(topic_name);
   size_t end_offset = topic_name_len;
 
+  topic_name = liveliness::mangle_name(topic_name);
+
   if (topic_name_len > 0) {
-    if (topic_name[0] == '/') {
+    if (topic_name[0] == '%') {
```
Yadunund: The whole point of relying on `mangle_name` is to avoid doing any manual replacement like this. Once mangled, the `topic_name` can be used as is since there will no longer be any `/` characters in the string.
So we don't need to strip any characters after we mangle the topic name.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -446,6 +446,24 @@ std::string mangle_name(const std::string & input)
   return output;
 }
 
+///=============================================================================
+char * mangle_name(const char * input)
```
Yadunund: Unnecessary code duplication.

Yadunund (changes requested):
@shivangvijay appreciate the effort to address the outstanding issue but the implementation here is not quite what is expected. I highly recommend commenting on the ticket to discuss implementation strategies before working on a PR in case something is unclear. It's also a good way to let others know you're working on this ticket to prevent duplicated efforts. Please look at my comments below.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
shivangvijay: Understood.

shivangvijay:
Thank you, @Yadunund , for your guidance. In the future, I will comment on the ticket.

I am unable to understand why my build pipeline is failing, I am not changing anything in the CMakelists.txt file. Please let me know what I am doing wrong.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
shivangvijay: I tried to resolve this. Please review it.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
shivangvijay: I tried to resolve this. Please review it.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -79,28 +79,20 @@ namespace
...
-      // Strip the leading '/'
-      start_offset = 1;
-    }
-    if (topic_name[end_offset - 1] == '/') {
-      // Strip the trailing '/'
-      end_offset -= 1;
-    }
-  }
+  std::string topic_nm(topic_name);
+  topic_nm = liveliness::mangle_name(topic_nm);
```
Yadunund: Simplify?
```suggestion
  topic_nm = liveliness::mangle_name(topic_name);
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -79,28 +79,20 @@ namespace
 // the old string into it. If this becomes a performance problem, we could consider
 // modifying the topic_name in place. But this means we need to be much more
 // careful about who owns the string.
-z_owned_keyexpr_t ros_topic_name_to_zenoh_key(const char * const topic_name, size_t domain_id)
+z_owned_keyexpr_t ros_topic_name_to_zenoh_key(const char * topic_name, size_t domain_id)
```
Yadunund: The original `const` qualification seem right to me. We don't want to modify the contents of the original pointer. Is there a reason this change?

Yadunund (changes requested):

rmw_zenoh_cpp/src/rmw_zenoh.cpp
shivangvijay: I need to update your code as 
```c++
    std::string topic_nm = liveliness::mangle_name(topic_name);
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
shivangvijay: I changed this for my earlier implementation but not required for current implementation. I made new commit.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -87,20 +87,11 @@ z_owned_keyexpr_t ros_topic_name_to_zenoh_key(const char * const topic_name, siz
...
-    if (topic_name[0] == '/') {
-      // Strip the leading '/'
-      start_offset = 1;
-    }
-    if (topic_name[end_offset - 1] == '/') {
-      // Strip the trailing '/'
-      end_offset -= 1;
-    }
-  }
+  std::string topic_nm = liveliness::mangle_name(topic_name);
```
Yadunund: And update the call below with the new variable name.
```suggestion
  const std::string mangled_topic_name = liveliness::mangle_name(topic_name);
```

Yadunund (changes requested):
Thanks for addressing the feedback. One more minor change.

shivangvijay:
@Yadunund Is there any work remaining in this PR? Is not get merged till now.

Yadunund:
@shivangvijay sorry for the delay. 

I think we should add a `/` separator between the domain id and mangled topic name such that the keyepxression would look like `0/%chatter`. This way, if we ever need to get the original constituents  of a keyexpression, we can split over `/`. We adopt this approach for the lifecycle token's keyexpression.

So could you append a `/` to the domain id string?

shivangvijay:
> @shivangvijay sorry for the delay.
> 
> I think we should add a `/` separator between the domain id and mangled topic name such that the keyepxression would look like `0/%chatter`. This way, if we ever need to get the original constituents of a keyexpression, we can split over `/`. We adopt this approach for the lifecycle token's keyexpression.
> 
> So could you append a `/` to the domain id string?

If I add '/' in the domain ID string, I am facing an issue. I have attached a screenshot for your reference.
rmw

I tried below thing:-
Screenshot from 2024-03-27 18-34-31
But the return of z_keyexpr is not a "z_owned_keyexpr_t" type.

shivangvijay:
Now, I am exploring [https://github.com/eclipse-zenoh/zenoh-cpp](https://github.com/eclipse-zenoh/zenoh-cpp) to understand z_owned_keyexpr_t and implementation of z_keyexpr, and how to convert it's return value to z_owned_keyexpr_t.

Yadunund:
Why not simply avoid calling `z_keyexpr_join`?

```
const std::string keyexpr_str = std::to_string(domain_id) + "/" + liveliness::mangle_name(topic_name);
return z_keyexpr(keyexpr_str.c_str());
```

shivangvijay:
> Why not simply avoid calling `z_keyexpr_join`?
> 
> ```
> const std::string keyexpr_str = std::to_string(domain_id) + "/" + liveliness::mangle_name(topic_name);
> return z_keyexpr(keyexpr_str.c_str());
> ```

z_keyexpr return z_keyexpr_t type not z_owned_keyexpr_t type.

Error

Yadunund:
Sorry I meant [z_keyexpr_new](https://github.com/eclipse-zenoh/zenoh-c/blob/acfd36b44fe16e66c588e3261e536169723f1545/include/zenoh_commons.h#L1528).

Yadunund (approved):

PR #131, opened 2024-03-20
Attach sequence number and publisher GID to serialized messages

### LINKED ISSUE #130: Unable to obtain publisher GID and sequence number errors
Hi,

Thanks for all your work on the Zenoh RMW! We are very excited to try it out. 

I'm getting the following errors when using data from a rosbag through `ros2 bag play`. These errors show up in any node that subscribes to topics published over Zenoh. It seems like these errors go away when subscribing to "live" topics rather than those played back through `ros2 bag`. Do you have any idea what the root cause could be? 

```
[ERROR] [1710953784.261021973] [rmw_zenoh_cpp]: Unable to obtain sequence number from the attachment.
[ERROR] [1710953784.261040865] [rmw_zenoh_cpp]: Unable to obtain publisher GID from the attachment.
```

AlexDayCRL:
Fixes #130 by attaching the GID and sequence number to the message in `rmw_publish_serialized_message` mirroring `rmw_publish`. Tested by building locally and publishing messages from SQLite3 bag and subscribing with `ros2 topic hz`

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -1044,6 +1044,20 @@ rmw_publish_serialized_message(
     return RMW_RET_ERROR;
   }
 
+  uint64_t sequence_number = publisher_data->get_next_sequence_number();
+
+  z_owned_bytes_map_t map = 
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This looks great.  I pushed a small change to fix trailing whitespace.  Once CI finishes I'll merge it in.

Issue #126, opened 2024-03-07
Revisit `zenoh_router_check`


clalancette:
During `rmw_init`, we end up making a call to see if the Zenoh router is available: https://github.com/ros2/rmw_zenoh/blob/ae142228c946147bd44ecb8696f46a7eb7208746/rmw_zenoh_cpp/src/detail/zenoh_router_check.cpp#L51-L81 .

However, in a very busy system that check might timeout, even though it would (eventually) find the router.

There are several ways we could handle this:
1.  Increase the timeout for `z_info_routers_zid`.  This has the problems of a) not knowing how to do that, and b) choosing a timeout value that is short enough not to hang if the router isn't available, but long enough to deal with busy systems.
2.  Wait forever for a router to appear.  This is very ROS 1-like behavior.
3.  Do router detection asynchronously.  During initialization we would query for the router, but we wouldn't wait around for a response.  Assuming that the router "eventually" responded, we would have a callback that marked that fact.  If the router didn't respond in some timeout (this could be many seconds), then we would print a warning message to the user.

There may be additional ways to deal with this.

fujitatomoya:
@Yadunund /assign

Yadunund:
Closed by #135

PR #135, opened 2024-03-22
Retry router connection

Yadunund:
Addresses #126 

This PR updates `rmw_init()` such that it blocks until a Zenoh router is discovered in `ROUTER_CONNECTION_RETRY_MAX` attempts. For now this value is not externally configurable and is set to `std::numeric_limits<uint64_t>::max()` which effectively retries indefinitely.

To test run a `talker` node without starting the router. The terminal should contain logs to remind the user to start the router. And once the router is started on a different terminal, execution should resume.

```bash
# terminal 1
yadunund@ubuntu-22-04:~/ws_rmw$ ros2 run demo_nodes_cpp talker
[ERROR] [1711097721.962485478] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711097722.962627606] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711097723.962784124] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
```

```bash
# terminal 2
yadunund@ubuntu-22-04:~/ws_rmw$ ros2 run rmw_zenoh_cpp rmw_zenohd 
Enter 'q' to quit...
```

Then 
```bash
# terminal 1
[ERROR] [1711098130.624961769] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711098131.625107213] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711098132.625249714] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711098133.625404414] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711098134.625538271] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711098135.625676527] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[ERROR] [1711098136.625823771] [rmw_zenoh_cpp]: Unable to connect to a Zenoh router. Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?
[INFO] [1711098137.626023133] [rmw_zenoh_cpp]: Successfully connected to a Zenoh router with id cb1d1e4b8534f374740e5ed92edab2b3!
[INFO] [1711098139.637644881] [talker]: Publishing: 'Hello World: 1'
[INFO] [1711098140.637620937] [talker]: Publishing: 'Hello World: 2'

```


clalancette (commented):
So I like the idea.  However, there are two issues that I see here.

1.  I do think we should make this configurable through an environment variable.  That is, I like this as a default behavior, but I think we should also allow the user to set the number of times to check (including maybe a way to disable the check completely).
2. When I was testing this out with rclpy applications, I couldn't kill off the rclpy application.  That is, I would run `ros2 run demo_nodes_py talker` with no router, and have it be in the loop saying "Unable to connect to a Zenoh router".  But then hitting Ctrl-C would do nothing.  Doing the same thing with rclcpp, I was able to kill off the application.  This is likely a bug somewhere in rclpy, but it is a pretty poor user experience right now.  Not sure what we should do about that.

Yadunund:
Thanks for the feedback, Chris. Wondering if we should have two separate envars : one to decide whether or not to check for a router's presence and the other for retry count. Else we could have a single envar `ZENOH_ROUTER_CHECK_ATTEMPTS` which if set to 0 will skip the router check and will attempt to connect with the router as many times as defined if greater than zero. 

Regarding the rclpy issue, it wonder if its related to when we register signal handlers. We should definitely fix it but not a huge blocker imo since the user terminate with `CRTL+Z` or can start the router and then hit Ctrl+C if needed.

Yadunund:
Alright I changed the implementation to be configurable via `ZENOH_ROUTER_CHECK_ATTEMPTS`. If set to <= 0, we skil the check. Else check as many times as specified. If unset, check indefinitely.

README.md
```diff
@@ -66,28 +66,43 @@ ros2 run demo_nodes_cpp listener
...
+
+### Checking for a Zenoh router.
+The `ZENOH_ROUTER_CHECK_ATTEMPTS` environment variable can be used to configure if and how a `Zenoh session` checks for the presence of a `Zenoh router`.
+The behavior is explained in the table below.
+
+
+| ZENOH_ROUTER_CHECK_ATTEMPTS |                                                 Session behavior                                                 |
+|:---------------------------:|:----------------------------------------------------------------------------------------------------------------:|
+|            unset            |                                                             Indefinitely waits for connection to a Zenoh router. |
+|             <= 0            |                                                                                        Skips Zenoh router check. |
```
clalancette: I'll suggest that we change this behavior slightly.  In particular, if it is unset or `0`, then it indefinitely waits, and if it is less than 0, we skip the check.  That feels a little more symmetric to me, and means that you can always set some value and get the behavior you want.

clalancette (changes requested):
I really like the idea here.  I've left one more piece of feedback, then I think we can get this in.

README.md
Yadunund: Done https://github.com/ros2/rmw_zenoh/commit/6366c9d360d5fe6d7fd2a3cceabf9a165cceb0f7

clalancette (approved):

PR #134, opened 2024-03-21
Sync past tokio bump and drop thread restriction

Yadunund:
Undo thread limits imposed by https://github.com/ros2/rmw_zenoh/pull/95 by syncing to `zenoh` version with `tokio` async runtime.

To be tested.

PR #136, opened 2024-03-25
Patch cargo toolchain for noble

Yadunund:
This PR allows `rmw_zenoh` to build on Ubuntu Noble with all dependencies resolved by `rosdep`, ie, users do no need to manually install cargo like we [currently specify](https://github.com/ros2/rmw_zenoh?tab=readme-ov-file#setup).

This is achieved by patching the [zenoh-c/CMakeLists.txt](https://github.com/eclipse-zenoh/zenoh-c/blob/bd0fe98772b7d65e8fb71eb0369d7c85c9b09c73/CMakeLists.txt#L201-L202) file during the vendor step to drop the `+stable` toolchain specification. Further, we pin the version of cargo to be used as `1.75.0` on all platforms since this is the versions of cargo available via apt in Noble.

Also created a temporary rolling CI job that passes to demonstrate that the fix works. This is interim until docker image for rolling+noble is [ready](https://github.com/osrf/docker_images/issues/728).

CI is working again 🎉


zenoh_c_vendor/CMakeLists.txt
```diff
@@ -24,13 +24,29 @@ find_package(ament_cmake_vendor_package REQUIRED)
 # when expanded.
 set(ZENOHC_CARGO_FLAGS "--no-default-features$<SEMICOLON>--features=zenoh/transport_tcp zenoh/shared-memory")
 
+# Always force rustc 1.75.0.
+# This this is default version of rustc available via apt in Ubuntu Noble.
```
clalancette:
```suggestion
# This is the default version of rustc available via apt in Ubuntu Noble.
```

zenoh_c_vendor/CMakeLists.txt
```diff
@@ -24,13 +24,29 @@ find_package(ament_cmake_vendor_package REQUIRED)
...
+# We explictly use the +1.75.0 toolchain when building on Ubuntu Jammy.
+set(ZENOHC_CARGO_CHANNEL "1.75.0")
+
+# Apply the patch the zenoh-c CMakeLists.txt only if building an ROS 2 version
```
clalancette:
```suggestion
# Apply the patch to the zenoh-c CMakeLists.txt only if building an ROS 2 version
```

zenoh_c_vendor/CMakeLists.txt
```diff
@@ -24,13 +24,29 @@ find_package(ament_cmake_vendor_package REQUIRED)
...
+# newer than Iron. The patch removes the ability to set ZENOHC_CARGO_CHANNEL
+# via cmake args
+set(pathces "")
```
clalancette:
```suggestion
set(patches "")
```

clalancette (changes requested):
Besides my changes inline, I think that we should consider approaching this slightly differently.

In particular, it is not whether we are on Iron or Rolling which determines whether we need to remove the ZENOHC_CARGO_CHANNEL, but rather whether we are using the toolchain from rustup, or from Ubuntu (this will also be a factor when we go to make this work on Windows).

Is there any way for us to tell whether the cargo we are using supports the various channels or not?  That would be closer to what we are trying to accomplish here.

Yadunund:
@clalancette thanks for the feedback. I agree with the logic proposed. I've implemented this by relying on cmake's `execute_process` function where I check if `cargo +stable` returns an error. If it does, the patch is applied.

JEnoch:
https://github.com/eclipse-zenoh/zenoh-c/issues/300 has been created to address this `+stable` flag issue in `zenoh-c/CMakeLists.txt`

JEnoch:
https://github.com/eclipse-zenoh/zenoh-c/pull/301 makes zenoh-c CMakeLists.txt to no longer use `+stable` by default:
https://github.com/eclipse-zenoh/zenoh-c/blob/dbc88773fdc34ec32b86e667a6369880ccbc84b8/CMakeLists.txt#L30

However this was pushed after switching Zenoh to Tokio which is currently [causing some troubles](https://github.com/ros2/rmw_zenoh/pull/134#issuecomment-2024955800)

README.md
```diff
@@ -15,10 +15,12 @@ For information about the Design please visit [design](docs/design.md) page.
 
 ## Setup
 
-Install latest rustc.
-> Note: The version of rustc that can be installed via apt is outdated.
+Install latest rustc via `rustup` if building on Ubuntu Jammy.
+Skip this step if building on Ubuntu Noble as `cargo` and `rustc` will be
```
ruffsl, line 19: This is cool! When/how did that happen with rosdep? Or does Noble ship debs for rust build tools?

Yadunund: The debs are available in Jammy too but the versions are very dated so we install them via `rustup` to compile zenoh. Noble ships with cargo/rustc `1.75.0` which works out of the box for our use case.

Yadunund:
@clalancette @JEnoch 
I've taken the diff from https://github.com/eclipse-zenoh/zenoh-c/pull/301 and created a local patch file such that we apply the diff from the PR when we vendor `zenoh-c`.
I think this a good way to benefit from changes upstream while not pinning the zenoh-c hash to a version with tokio just yet.

clalancette (approved):
Looks reasonable to me to patch this for now.

PR #139, opened 2024-03-26
Run CI on all matrix jobs despite failure

Yadunund:
So that iron jobs run even if rolling fails due to interim docker issue


clalancette (approved):

Yadunund:
Closing in favor of #136

PR #141, opened 2024-03-26
FastCDR version bump

AlexDayCRL:
the ros2 rolling repo recently bumped FastCDR from version 1.1.x to 2.2.x. This moves around some of the constants for FastBuffers. Specifically it renames getSerializedDataLength to get_serialized_data_length and moves where the DDS_CDR serialization flag lives. 

References:
Rolling repos update: https://github.com/ros2/ros2/pull/1530/files
FastCDR Include Dirs @ 1.1.x: https://github.com/eProsima/Fast-CDR/tree/1.1.x/include/fastcdr
FastCDR Include Dirs @ 2.2.x: https://github.com/eProsima/Fast-CDR/tree/2.2.x/include/fastcdr

### DIFF
--- a/rmw_zenoh_cpp/CMakeLists.txt
+++ b/rmw_zenoh_cpp/CMakeLists.txt
@@ -27,6 +27,7 @@ find_package(zenohc REQUIRED)
 
 add_library(rmw_zenoh_cpp SHARED
   src/detail/attachment_helpers.cpp
+  src/detail/cdr.cpp
   src/detail/event.cpp
   src/detail/identifier.cpp
   src/detail/graph_cache.cpp
--- a/rmw_zenoh_cpp/src/detail/cdr.cpp
+++ b/rmw_zenoh_cpp/src/detail/cdr.cpp
@@ -1,0 +1,42 @@
+// Copyright 2024 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "fastcdr/Cdr.h"
+#include "fastcdr/FastBuffer.h"
+#include "fastcdr/config.h"
+
+#include "cdr.hpp"
+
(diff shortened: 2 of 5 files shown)


AlexDayCRL:
this is also stacked on top of `yadu/noble` and rebased to `rolling`.

Yadunund:
Thanks for spear heading this initiative. One thing to keep in mind is that we need the code to compile for Iron on Jammy with FastCDR `1.x.x` and Rolling on Noble with FastCDR `2.x.x`. We can achieve this in two ways
1.  Branch off `rolling` to create an `iron` branch for `rmw_zenoh`. Then merge this into `rolling`. All future PRs merged into `rolling` will need to be carefully backported to `iron`.
2. We implement this change by
  - Checking version of `FastCDR` found  a `find_package()` call within `rmw_zenoh_cpp/CMakeLists.txt` and then injecting a compile definition into the codebase, say `-DFAST_CDR_2` if `2.x.x` is found.
  - Updating `rmw_zenoh.cpp` such that it relies on `#ifdef FAST_CDR_2` blocks to invoke the `2.x.x` API or else the `1.x.x` ones.

Personally I'm in favor on 2 since `rmw_zenoh` is still under heavy development and I rely on Iron binaries for Nav2, moveit, open-rmf, etc to test functionality. @clalancette what do you think?

clalancette:
So I rebased this onto the latest, and I also went with strategy 2 as laid out by @Yadunund .  For now, we can keep all of our development on the `rolling` branch, though eventually we are going to want to branch off for `iron` (and soon `jazzy`).  In any case, what I did in 8dee384442cf0fef700ef97f08e2a15ed560170d should paper over those differences, and in my testing, it makes this build for both Fast-CDR v1 and v2.

Please take a look.

Yadunund (approved):
LGTM. Tested locally with rolling from source and Fast-CDR 2.x.
I'm surprised we didn't need to include a `add_compile_definitions(-DFASTCDR_VERSION_MAJOR ${FASTCDR_MAJOR_VERSION})` in the `CMakeLists.txt` file.

clalancette:
> I'm surprised we didn't need to include a `add_compile_definitions(-DFASTCDR_VERSION_MAJOR ${FASTCDR_MAJOR_VERSION})` in the `CMakeLists.txt` file.

Fast-CDR already generates this during its own installation phase and puts it in `config.h` file, which we can then reuse.  So no need for our own definition.

Yadunund:
> > I'm surprised we didn't need to include a `add_compile_definitions(-DFASTCDR_VERSION_MAJOR ${FASTCDR_MAJOR_VERSION})` in the `CMakeLists.txt` file.
> 
> Fast-CDR already generates this during its own installation phase and puts it in `config.h` file, which we can then reuse. So no need for our own definition.

Ah that explains it.

PR #142, opened 2024-03-26
Fix memory leak in `rmw_publisher`

AlexDayCRL:
The msg_bytes capture for the lambda to clean up memory was captured by value which meant it couldn't actually discard the memory. Passing by reference allows the allocator to dealloc the memory.


Yadunund (approved):
LGTM.

PR #150, opened 2024-04-04
Move graph guard condition destruction to rmw_context_fini.

clalancette:
That's because during rmw_shutdown(), the graph guard condition may still be in use in rwm_wait() and friends. Fix this by moving the destruction of it later, into rmw_context_fini.


Yadunund (approved):
This makes sense! Thanks for the fix.

PR #155, opened 2024-04-13
test_rclcpp in ci

Yadunund:


MichaelOrlov:
@Yadunund Could you please handle this issue? 
We decided to assign this issue to you at the latest waffle meeting. However, I don't have permission to change the Assignee field.

Yadunund:
@CihatAltiparmak would you be willing to finish up what i started here? ie, follow-up #256 by updating the CI such that when it builds the Rolling source job, it also builds and tests `test_rclcpp`? The tricky part is getting the zenoh router to run in the github action along with the tests.

CihatAltiparmak:
With a big pleasure. I was planning it as well. Seems like tricky part of Zenoh router is solved by 2 PRs so that tty was removed. I will start tomorrow.

Yadunund:
> With a big pleasure. I was planning it as well. Seems like tricky part of Zenoh router is solved by 2 PRs so that tty was removed. I will start tomorrow.

Thanks so much!

CihatAltiparmak:
Superseded by #271 Looks like zenoh router runs without problem. Seems like easier than i guess.

PR #153, opened 2024-04-13
Fix race conditions in rmw_wait and map queries to clients

Yadunund:
Fixes https://github.com/ros2/rmw_zenoh/issues/152.

PR #160, opened 2024-04-18
Fix a race condition in rmw_wait.

Yadunund:
This PR cherry picks the fix to rmw_wait from #153 

To very briefly explain, rmw_wait:
1.  Checks to see if any of the entities (subscriptions, clients, etc) have data ready to go.
2.  If they have data ready to go, then we skip attaching the condition variable and waiting.
3.  If they do not have data ready to go, then we attach the condition variable to all entities, take the condition variable lock, and call wait_for/wait on the condition variable.
4.  Regardless of whether we did 3 or 4, we check every entity to see if there is data ready, and mark that as appropriate in the wait set.

There is a race in all of this, however.  If data comes in after we've checked the entity (1), but before we've attached the condition variable (3), then we will never be woken up.  In most cases, this means that we'll wait the full timeout for the wait_for, which is not what we want.

Fix this by adding another step to 3.  In particular, after we've locked the condition variable mutex, check the entities again. Since we change the entities to *also* take the lock before we notify, this ensures that the entities cannot make changes that get lost.


MichaelOrlov:
@clalancette A friendly ping to handle this issue.  
At the waffle meeting, we decided to assign it to you, but I don't have permission to do so.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3109,20 +3109,21 @@ rmw_destroy_wait_set(rmw_wait_set_t * wait_set)
...
-  rmw_subscriptions_t * subscriptions,
-  rmw_guard_conditions_t * guard_conditions,
-  rmw_services_t * services,
-  rmw_clients_t * clients,
-  rmw_events_t * events)
+  const rmw_subscriptions_t * const subscriptions,
+  const rmw_guard_conditions_t * const guard_conditions,
+  const rmw_services_t * const services,
+  const rmw_clients_t * const clients,
+  const rmw_events_t * const events)
```
clalancette, line 3118: With the latest changes here, I think this entire function could be const, i.e.:

```suggestion
  const rmw_events_t * const events) const
```

clalancette: Ah, never mind, this is a free function, so can't be const.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3109,20 +3109,21 @@ rmw_destroy_wait_set(rmw_wait_set_t * wait_set)
...
 {
   static_cast<void>(events);
 
   if (guard_conditions) {
     for (size_t i = 0; i < guard_conditions->guard_condition_count; ++i) {
       GuardCondition * gc = static_cast<GuardCondition *>(guard_conditions->guard_conditions[i]);
       if (gc != nullptr) {
-        if (gc->get_and_reset_trigger()) {
+        if (gc->get_trigger()) {
```
mjcarroll, line 3126: Just for posterity.  What I _think_ was happening here is that checking for triggered conditions wasn't idempotent and was actually mutating the underlying conditions.  Now we are only _reading_ the condition (and not resetting the trigger) each time we are checking.

clalancette: > Just for posterity. What I _think_ was happening here is that checking for triggered conditions wasn't idempotent and was actually mutating the underlying conditions. Now we are only _reading_ the condition (and not resetting the trigger) each time we are checking.

Yep, you are exactly right.

clalancette: Just as an FYI, I've come up with a completely different approach to implementing `rmw_wait` on https://github.com/ros2/rmw_zenoh/tree/clalancette/rewrite-rmw-wait .  It needs more work, but I also think that it has  more promise than this PR to be much more performant.  I haven't opened a PR yet since it is still WIP.

Yadunund:
Superseded by https://github.com/ros2/rmw_zenoh/pull/203

PR #168, opened 2024-05-04
Add rmw_zenoh_cpp namespace

### LINKED ISSUE #60: Missing rmw_zenoh_cpp namespace
We should probably add some namespaces to the cpp code to better organize the code and avoid possible clashing in the future

```cpp
namespace rmw_zenoh_cpp
namespace detail
```

Yadunund:
This PR adds the `rmw_zenoh_cpp` namespace to all internal definitions except `rmw_context_impl_s` as per [expectation defined in rmw/init.h](https://github.com/ros2/rmw/blob/222b7c9a97789aed2d29ac63295a3fe493d5a72c/rmw/include/rmw/init.h#L30-L34).

Fixes https://github.com/ros2/rmw_zenoh/issues/60 but without a `detail` namespace. I felt it is not needed but let me know otherwise.


rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -836,7 +838,8 @@ rmw_ret_t GraphCache::publisher_count_matched_subscriptions(
   *subscription_count = 0;
   GraphNode::TopicMap::const_iterator topic_it = graph_topics_.find(publisher->topic_name);
   if (topic_it != graph_topics_.end()) {
-    rmw_publisher_data_t * pub_data = static_cast<rmw_publisher_data_t *>(publisher->data);
+    rmw_zenoh_cpp::rmw_publisher_data_t * pub_data =
+      static_cast<rmw_zenoh_cpp::rmw_publisher_data_t *>(publisher->data);
```
clalancette: Since we are already in the namespace, I don't think we need this change, right?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -871,7 +874,8 @@ rmw_ret_t GraphCache::subscription_count_matched_publishers(
   *publisher_count = 0;
   GraphNode::TopicMap::const_iterator topic_it = graph_topics_.find(subscription->topic_name);
   if (topic_it != graph_topics_.end()) {
-    rmw_subscription_data_t * sub_data = static_cast<rmw_subscription_data_t *>(subscription->data);
+    rmw_zenoh_cpp::rmw_subscription_data_t * sub_data =
+      static_cast<rmw_zenoh_cpp::rmw_subscription_data_t *>(subscription->data);
```
clalancette: Same here.

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -102,7 +104,7 @@ std::unique_ptr<saved_msg_data> rmw_subscription_data_t::pop_next_message()
     return nullptr;
   }
 
-  std::unique_ptr<saved_msg_data> msg_data = std::move(message_queue_.front());
+  std::unique_ptr<rmw_zenoh_cpp::saved_msg_data> msg_data = std::move(message_queue_.front());
```
clalancette, line 125: This change shouldn't be necessary.

(same goes for all of the changes within this namespace below)

rmw_zenoh_cpp/src/detail/rmw_data_types.hpp
```diff
@@ -316,15 +318,15 @@ class rmw_client_data_t final
 
   size_t get_next_sequence_number();
 
-  void add_new_reply(std::unique_ptr<ZenohReply> reply);
+  void add_new_reply(std::unique_ptr<rmw_zenoh_cpp::ZenohReply> reply);
```
clalancette, line 325: This change shouldn't be necessary, as we are in the namespace (same below).

clalancette (changes requested):
I've left a few things to fix, but generally it looks good to me.

rmw_zenoh_cpp/src/detail/graph_cache.cpp
Yadunund: Done.

clalancette (approved):

PR #169, opened 2024-05-05
Support RMW_EVENT_MESSAGE_LOST

Yadunund:
This PR brings support for `RMW_EVENT_MESSAGE_LOST`.

It's best to test with https://github.com/ros2/demos/pull/679.

Run the `message_lost_talker` node to publish at 1000Hz. See https://github.com/ros2/demos/tree/rolling/quality_of_service_demo#message-lost-status-event-demo for more details.

Also removes unused variables from #103


clalancette (approved):
This generally looks OK to me.

I will say that I have some misgivings about adding this.  It seems kind of fragile, and I worry that we'll receive this event more than we expect.  But there are no concrete problems I see, so I think we should put this in and see what happens.

PR #171, opened 2024-05-10
Include topic type and hash in key expression

### LINKED ISSUE #117: Crash when deserializing message of matching topic but different type.
Presently the key expressions for pub/subs do not include the topic type. Therefore with pub/subs created with same topic name and different type, the sub will still receive serialized payload form the pub but an error may be thrown when the sub fails to deserialize the payload into the sub type.
We can either include the type in the key expressions or just handle the error on the sub side.

Steps to reproduce: 

```bash
#terminal 1
ros2 topic pub /chatter std_msgs/msg/Float32 '{data: 6.0}'
```

```bash
#terminal 2
ros2 topic echo /chatter std_msgs/msg/String
```

The following error will be thrown in terminal 2.

```bash
>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'Fast CDR exception deserializing message of type std_msgs::msg::dds_::String_., at /home/yadunund/ws_rmw/src/rmw_alternative/rmw_zenoh_cpp/src/detail/type_support.cpp:133'

with this new error message:

  'could not deserialize ROS message, at /home/yadunund/ws_rmw/src/rmw_alternative/rmw_zenoh_cpp/src/rmw_zenoh.cpp:1567'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
failed to take message from subscription: could not deserialize ROS message, at /home/yadunund/ws_rmw/src/rmw_alternative/rmw_zenoh_cpp/src/rmw_zenoh.cpp:1567, at /home/yadunund/ros2_rolling/src/ros2/rcl/rcl/src/rcl/subscription.c:562
```

Yadunund:
Fix https://github.com/ros2/rmw_zenoh/issues/117 while factoring in https://github.com/ros2/rmw_zenoh/issues/201#issuecomment-2166445736.

The keyexprs now adopt the format

```
<ros_domain_id>/<topic_name>/<topic_type>/<topic_hash>
``` 

Topic names are no longer mangled.

I will follow up with a separate PR to also add the `topic_hash` to the liveliness tokens so that we can include the hash in the endpoint info when querying topic info.

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -102,10 +102,13 @@ z_owned_keyexpr_t ros_topic_name_to_zenoh_key(
   const char * const topic_type,
   const char * const type_hash)
 {
-  const std::string keyexpr_str = std::to_string(domain_id) + "/" +
-    strip_slashes(topic_name) + "/" +
-    std::string(topic_type) + "/" +
-    std::string(type_hash);
+  std::string keyexpr_str = std::to_string(domain_id);
+  keyexpr_str += "/";
+  keyexpr_str += strip_slashes(topic_name);
+  keyexpr_str += "/";
+  keyexpr_str += topic_type;
+  keyexpr_str += "/";
+  keyexpr_str += type_hash;
 
   return z_keyexpr_new(keyexpr_str.c_str());
 }
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -69,9 +69,10 @@
 
 namespace
 {
-
 //==============================================================================
-// A function to take ros topic names and convert them to valid Zenoh keys.
+// A function that generates a key expression for message transport of the format
+// <ros_domain_id>/<topic_name>/<topic_type>/<topic_hash>
+
 // In particular, Zenoh keys cannot start or end with a /, so this function
 // will strip them out.
 // The Zenoh key is also prefixed with the ros_domain_id.
@@ -79,11 +80,31 @@ namespace
 // the old string into it. If this becomes a performance problem, we could consider
 // modifying the topic_name in place. But this means we need to be much more
 // careful about who owns the string.
-z_owned_keyexpr_t ros_topic_name_to_zenoh_key(const char * const topic_name, size_t domain_id)
+z_owned_keyexpr_t ros_topic_name_to_zenoh_key(
+  const std::size_t domain_id,
+  const char * const topic_name,
+  const char * const topic_type)
 {
-  const std::string keyexpr_str = std::to_string(domain_id) +
-    "/" +
-    rmw_zenoh_cpp::liveliness::mangle_name(topic_name);
+  auto strip_slashes =
+    [](const char * const str) -> std::string
+    {
+      std::string ret = std::string(str);
+      const std::size_t len = strlen(str);
+      std::size_t start = 0;
+      std::size_t end = len - 1;
+      if (str[0] == '/') {
+        ++start;
+      }
+      if (str[end] == '/') {
+        --end;
+      }
+      return ret.substr(start, end - start);
+    };
+
+  const std::string keyexpr_str = std::to_string(domain_id) + "/" +
+    strip_slashes(topic_name) + "/" +
+    std::string(topic_type);
+
   return z_keyexpr_new(keyexpr_str.c_str());
 }
 
@@ -568,7 +589,9 @@ rmw_create_publisher(
     });
 
   z_owned_keyexpr_t keyexpr = ros_topic_name_to_zenoh_key(
-    topic_name, node->context->actual_domain_id);
+    node->context->actual_domain_id,
+    topic_name,
+    publisher_data->type_support->get_name());
   auto always_free_ros_keyexpr = rcpputils::make_scope_exit(
     [&keyexpr]() {
       z_keyexpr_drop(z_move(keyexpr));
@@ -1360,7 +1383,9 @@ rmw_create_subscription(
 
   z_owned_closure_sample_t callback = z_closure(rmw_zenoh_cpp::sub_data_handler, nullptr, sub_data);
   z_owned_keyexpr_t keyexpr = ros_topic_name_to_zenoh_key(
-    topic_name, node->context->actual_domain_id);
+    node->context->actual_domain_id,
+    topic_name,
+    sub_data->type_support->get_name());
   auto always_free_ros_keyexpr = rcpputils::make_scope_exit(
     [&keyexpr]() {
       z_keyexpr_drop(z_move(keyexpr));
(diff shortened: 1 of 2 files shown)


rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -69,21 +69,43 @@
...
 namespace
 {
-
 //==============================================================================
-// A function to take ros topic names and convert them to valid Zenoh keys.
+// A function that generates a key expression for message transport of the format
+// <ros_domain_id>/<topic_name>/<topic_type>/<topic_hash>
 // In particular, Zenoh keys cannot start or end with a /, so this function
 // will strip them out.
 // The Zenoh key is also prefixed with the ros_domain_id.
```
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -69,21 +69,43 @@
...
 // Performance note: at present, this function allocates a new string and copies
 // the old string into it. If this becomes a performance problem, we could consider
 // modifying the topic_name in place. But this means we need to be much more
 // careful about who owns the string.
-z_owned_keyexpr_t ros_topic_name_to_zenoh_key(const char * const topic_name, size_t domain_id)
-{
-  const std::string keyexpr_str = std::to_string(domain_id) +
-    "/" +
-    rmw_zenoh_cpp::liveliness::mangle_name(topic_name);
```
clalancette, line 86: Just to make sure I understand properly; we are no longer using `mangle_name` here so that we stick with slashes wherever possible for performance reasons, right?  The downside here is that it won't be possible to look at a keyexpression and tell which parts are the topic name, and which are Zenoh splits, but I guess I'm OK with that for now.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -69,21 +69,43 @@
...
+      std::size_t start = 0;
+      std::size_t end = len - 1;
+      if (str[0] == '/') {
+        ++start;
+      }
+      if (str[end] == '/') {
+        --end;
+      }
+      return ret.substr(start, end - start);
+    };
```
clalancette: I think it would be nicer to make this a separate function, rather than a lambda here.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -567,8 +590,23 @@ rmw_create_publisher(
...
+  char * type_hash_c_str = nullptr;
+  rcutils_ret_t stringify_ret = rosidl_stringify_type_hash(
+    publisher_data->type_hash,
+    *allocator,
+    &type_hash_c_str);
+  if (RCUTILS_RET_BAD_ALLOC == stringify_ret) {
+    RMW_SET_ERROR_MSG("Failed to allocate type_hash_c_str.");
+    return nullptr;
+  }
+
```
clalancette, line 612: We need to free the `type_hash_c_str` when we are done with it, something like:

```
  auto free_type_hash_c_str = rcpputils::make_scope_exit(
    [&type_hash_c_str]() {
      allocator.deallocate(type_hash_c_str, allocator.state);
    });
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2080,6 +2123,33 @@ rmw_create_client(
...
+  char * type_hash_c_str = nullptr;
+  rcutils_ret_t stringify_ret = rosidl_stringify_type_hash(
+    client_data->type_hash,
+    *allocator,
+    &type_hash_c_str);
+  if (RCUTILS_RET_BAD_ALLOC == stringify_ret) {
+    RMW_SET_ERROR_MSG("Failed to allocate type_hash_c_str.");
+    return nullptr;
+  }
+
```
clalancette, line 2154: Don't forget to free the type_hash_c_str when we are done with it:

```
  auto free_type_hash_c_str = rcpputils::make_scope_exit(
    [&type_hash_c_str]() {
      allocator.deallocate(type_hash_c_str, allocator.state);
    });
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2590,8 +2661,39 @@ rmw_create_service(
...
+  char * type_hash_c_str = nullptr;
+  rcutils_ret_t stringify_ret = rosidl_stringify_type_hash(
+    service_data->type_hash,
+    *allocator,
+    &type_hash_c_str);
+  if (RCUTILS_RET_BAD_ALLOC == stringify_ret) {
+    RMW_SET_ERROR_MSG("Failed to allocate type_hash_c_str.");
+    return nullptr;
+  }
+
```
clalancette, line 2711: Don't forget to free the type_hash_c_str when we are done with it:

```
  auto free_type_hash_c_str = rcpputils::make_scope_exit(
    [&type_hash_c_str]() {
      allocator.deallocate(type_hash_c_str, allocator.state);
    });
```

clalancette (changes requested):

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Yes that's exactly right.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/171/commits/5796e7426454f4ac73ce37a28a00954cf327c32c

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/171/commits/5796e7426454f4ac73ce37a28a00954cf327c32c

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -69,21 +69,44 @@
...
+  const std::size_t len = strlen(str);
+  std::size_t start = 0;
+  std::size_t end = len - 1;
+  if (str[0] == '/') {
+    ++start;
+  }
+  if (str[end] == '/') {
+    --end;
+  }
+  return ret.substr(start, end - start);
```
clalancette: I think this is incorrect.  Consider a str being passed in here that is "/foo".  In that case, `len` will be 4, `start` will begin at 0, and end will begin at 3.  Now `str[0]` is indeed `/`, so start is now `1`.  When we do the `ret.substr`, we'll only copy in `end - start`, `3 - 1`, or 2 bytes.  So that means we'll only get `fo`, when we meant to get `foo`.  Thus I think this needs to be:

```suggestion
  return ret.substr(start, end - start + 1);
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -69,21 +69,44 @@
...
+  const char * const topic_type,
+  const char * const type_hash)
 {
-  const std::string keyexpr_str = std::to_string(domain_id) +
-    "/" +
-    rmw_zenoh_cpp::liveliness::mangle_name(topic_name);
+  const std::string keyexpr_str = std::to_string(domain_id) + "/" +
+    strip_slashes(topic_name) + "/" +
+    std::string(topic_type) + "/" +
+    std::string(type_hash);
```
clalancette: We can avoid at least one allocation here by doing:

```
  std::string keyexpr_str = std::to_string(domain_id);
  keyexpr_str += "/";
  keyexpr_str += strip_slashes(topic_name);
  keyexpr_str += "/";
  keyexpr_str += topic_type;
  keyexpr_str += "/";
  keyexpr_str += type_hash;
```

clalancette (changes requested):
I've left two things inline to fix; once those are done, I think we are good to go here.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Thanks for the feedback. Done https://github.com/ros2/rmw_zenoh/pull/171/commits/50d7339473b4c80e328d69f1b2cbd6e096e0a02f

clalancette (approved):
This looks great to me, thanks for the fixups.

PR #172, opened 2024-05-13
git is required in buildstep

Timple:
It's probably not coming in via `ament_cmake_vendor_package` as that package supports multiple vcs.

### DIFF
--- a/zenoh_c_vendor/package.xml
+++ b/zenoh_c_vendor/package.xml
@@ -8,7 +8,6 @@
   <license>Apache License 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
-  <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>ament_cmake_vendor_package</buildtool_depend>
 
   <build_depend>cargo</build_depend>
--- a/zenoh_c_vendor/package.xml
+++ b/zenoh_c_vendor/package.xml
@@ -8,12 +8,12 @@
   <license>Apache License 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
-
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>ament_cmake_vendor_package</buildtool_depend>
 
   <build_depend>cargo</build_depend>
   <build_depend>clang</build_depend>
+  <build_depend>git</build_depend>
 
   <export>
     <build_type>ament_cmake</build_type>


Yadunund (approved):

christophebedard:
I did a search for `>git</` in my ROS 2 source workspace and I only see `<buildtool_depend>`, not `<build_depend>`:

```
$ ag '>git</' src/
src/eclipse-iceoryx/iceoryx/iceoryx_posh/package.xml
14:    <buildtool_depend>git</buildtool_depend>

src/ros2/rosbag2/mcap_vendor/package.xml
13:  <buildtool_depend>git</buildtool_depend>

src/ament/google_benchmark_vendor/package.xml
19:  <buildtool_depend>git</buildtool_depend>

src/eProsima/foonathan_memory_vendor/package.xml
14:  <buildtool_depend>git</buildtool_depend>
```

So should it be `<buildtool_depend>` or `<build_depend>`?

Yadunund:
@christophebedard you're right `buildtool_depend` is the more appropriate one. We should also update the `cargo` dep to be `buildtool_depend`. Do you mind opening a PR?

Timple:
My rationale was: It's used by `cargo` which is `build_depend`.

Although I'm not that sure anymore it was cargo which used it.

Yadunund:
Actually `git` should be used by `ament_vendor` [here](https://github.com/ros2/rmw_zenoh/blob/c76c9627ca8cf8c133124368f2f1ad06092b7027/zenoh_c_vendor/CMakeLists.txt#L28) to clone in `zenoh-c`. But I think this is transitively satisfied by the `python3-vcstool` dep defined [here](https://github.com/ament/ament_cmake/blob/79cc237f8eb819edf4c1c624b56451e0a05a45f8/ament_cmake_vendor_package/package.xml#L19).

clalancette:
> Actually `git` should be used by `ament_vendor` [here](https://github.com/ros2/rmw_zenoh/blob/c76c9627ca8cf8c133124368f2f1ad06092b7027/zenoh_c_vendor/CMakeLists.txt#L28) to clone in `zenoh-c`. But I think this is transitively satisfied by the `python3-vcstool` dep defined [here](https://github.com/ament/ament_cmake/blob/79cc237f8eb819edf4c1c624b56451e0a05a45f8/ament_cmake_vendor_package/package.xml#L19).

The thing is, @Timple is right; `python3-vcstool` has no `git` dependency:

```
ubuntu-24.04:~$ sudo apt-cache depends python3-vcstool
python3-vcstool
  Depends: python3-pkg-resources
  Depends: python3-yaml
  Depends: <python3:any>
    python3
  Depends: python3-setuptools
  Conflicts: <python-vcstool>
```

Yadunund:
Ah thanks for the clarification @clalancette.

Issue #179, opened 2024-05-24
Add a note on how users can build zenoh with other feature flags


Yadunund:
https://github.com/ros2/rmw_zenoh/blob/93ba28e0959940200ebbd0066d3433a426088ef2/zenoh_c_vendor/CMakeLists.txt#L25

Yadunund:
Closed by #180

PR #180, opened 2024-05-24
Add TLS and compression features by default

JEnoch:
Both TLS and compression are important features available in Zenoh.
This PR make them build by default for `rmw_zenoh`.

It also updates a comment wrt. threading and Tokio. This comment became obsolete after the switch of Zenoh from async_std to Tokio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ For information about the Design please visit [design](docs/design.md) page.
 
 Build `rmw_zenoh_cpp`
 
+>Note: By default, we vendor and compile `zenoh-c` with a subset of `zenoh` features.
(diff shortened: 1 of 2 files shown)


JEnoch:
@Yadunund do you think the comment I changed in this PR is good enough for #179 ?

BTW, the compression we enabled in config during our test today was actually not active, because the feature wasn't built... :sweat_smile:

Yadunund (approved):
Thanks for adding these features. Just pushed some minor tweaks. 

> BTW, the compression we enabled in config during our test today was actually not active, because the feature wasn't built... 😅

Whoops!

Issue #184, opened 2024-05-29
ros2 node list and ros2 topic list not working


berndpfrommer:
After your very nice youtube presentation today I took a stab at it, compiled and overlayed the rmw_zenoh_cpp workspace.
Talker and listener are working fine, but when I open a third terminal and try to run
``ros2 node list``
no nodes are shown and ``ros2 topic list`` shows only
```
/parameter_events
/rosout
```
Are those commands supposed to work?

Yadunund:
Hi @berndpfrommer,

Thanks for trying out `rmw_zenoh`.

That sounds really strange. 

This is silly of me to ask but did you `export RMW_IMPLEMENTATION=rmw_zenoh_cpp` in the third terminal? It will also help to ensure any `ros2 daemons` started previously are properly terminated before switching to `rmw_zenoh` since the cli tools query the `daemon` for ROS graph information.  Running the below command before running any nodes with `rmw_zenoh` would work.

```bash
pkill -9 -f ros && ros2 daemon stop
```

berndpfrommer:
RMW_IMPLEMENTATION was set correctly. Killing the ros2 daemon did the trick!
Thanks!

Yadunund:
Glad to hear!

PR #185, opened 2024-05-29
document procedure to kill stale ros2 daemons

berndpfrommer:
This is related to issue #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ ros2 run rmw_zenoh_cpp rmw_zenohd
 
 > Note: Without the zenoh router, nodes will not be able to discover each other since multicast discovery is disabled by default in the node's session config. Instead, nodes will receive discovery information about other peers via the zenoh router's gossip functionality. See more information on the session configs [below](#config).
 
+### Clear stale ros2 daemons
+```bash
+pkill -9 -f ros && ros2 daemon stop
+```
+Without this step the ROS2 CLI commands (e.g. ``ros2 node list``) may
+not work properly since they could still be using a ros2 daemon that
+is running on a different RMW.
+
 ### Run the `talker`
 ```bash
 # terminal 2
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ ros2 run rmw_zenoh_cpp rmw_zenohd
 ```bash
 pkill -9 -f ros && ros2 daemon stop
 ```
-Without this step the ROS2 CLI commands (e.g. ``ros2 node list``) may
+Without this step, ROS 2 CLI commands (e.g. `ros2 node list`) may
 not work properly since they could still be using a ros2 daemon that
 is running on a different RMW.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,14 @@ ros2 run rmw_zenoh_cpp rmw_zenohd
 
 > Note: Without the zenoh router, nodes will not be able to discover each other since multicast discovery is disabled by default in the node's session config. Instead, nodes will receive discovery information about other peers via the zenoh router's gossip functionality. See more information on the session configs [below](#config).
 
+### Clear stale ros2 daemons
+```bash
+pkill -9 -f ros && ros2 daemon stop
+```
+Without this step the ROS2 CLI commands (e.g. ``ros2 node list``) may
```
Yadunund:
```suggestion
Without this step, ROS 2 CLI commands (e.g. `ros2 node list`) may
```

README.md
```diff
@@ -47,6 +47,14 @@ ros2 run rmw_zenoh_cpp rmw_zenohd
...
+not work properly since they could still be using a ros2 daemon that
```
Yadunund:
```suggestion
not work properly since they would query ROS graph information from the ROS 2 daemon that
```

README.md
```diff
@@ -47,6 +47,14 @@ ros2 run rmw_zenoh_cpp rmw_zenohd
...
+is running on a different RMW.
```
Yadunund:
```suggestion
may have been started with different a RMW.
```

README.md
```diff
@@ -47,6 +47,14 @@ ros2 run rmw_zenoh_cpp rmw_zenohd
 
 > Note: Without the zenoh router, nodes will not be able to discover each other since multicast discovery is disabled by default in the node's session config. Instead, nodes will receive discovery information about other peers via the zenoh router's gossip functionality. See more information on the session configs [below](#config).
 
+### Clear stale ros2 daemons
```
Yadunund:
```suggestion
### Terminate ROS 2 daemon started with another RMW
```

Yadunund (changes requested):
minor nitpicks

Timple:
Maybe not the correct place to ask. But might ROS be improved such that the daemon is only spawned for rmw's that need it?

berndpfrommer:
Accepted your suggested changes, LGTM.

Yadunund (approved):

Yadunund:
> Maybe not the correct place to ask. But might ROS be improved such that the daemon is only spawned for rmw's that need it?

It may be worth having a discussion on https://github.com/ros2/ros2cli/tree/rolling. But imo even `rmw_zenoh` benefits from the daemon (which will be started automatically when there isn't one running and your run a `ros2 topic/node` command) since these graph queries can be made without initializing a new ROS context and hence waiting to connect with the Zenoh router; instead, we can query this information from the daemon node that has already connected to the router. Perhaps ` ros2 shutdown` command will be helpful to ensure all ROS 2 nodes and daemons are terminated before users switch to another RMW.

Issue #188, opened 2024-05-30
zenohd should log router ID to console


berndpfrommer:
Is there  any reason why not to have the router log its id to the console? It would make it easier to verify that the clients indeed connect to the right router. Right now, the zenohd only produces this output:
```
ros2 run rmw_zenoh_cpp rmw_zenohd
Enter 'q' to quit...
```
Whereas the client is more talkative:
```
 Successfully connected to a Zenoh router with id a9d4ace930f270fb60876a363d75844c.
```
When I had trouble getting things to work (see issue #184), the first question was: what router is the client connecting to? I know the daemon's log verbosity can be increased with the right env variable, but if it had just logged its ID on startup, that would have saved me from finding out how to change the log verbosity.

Yadunund:
Hi @berndpfrommer,

You can run the bridge with  envar `export RUST_LOG=zenoh=debug,zenoh::net::routing::queries=trace` and it should print out all the information you need. Let me know if that helps.

berndpfrommer:
Thanks, I got it from the logs already. But you can probably tell that the above line is not super intuitive, and if things don't work, people have to first find out how to enable the logging. If you want to keep the default output minimal, a "--debug" flag as argument to zenohd would also suffice.
At the moment I have no problems with rmw_zenoh_cpp. Works great for me. Just wanted to make a suggestion to improve the out-of-the-box experience of rmw_zenoh_cpp.

Yadunund:
Thanks for the valuable feedback! I'm okay with having the router print a log each time a session connects to it! We can have more details logged when `--debug` is passed.
I'd be more than happy to review a PR with these changes.

PR #189, opened 2024-06-01
print router id on startup

berndpfrommer:
Minimal PR related to #188 
This was enough for me to realize that it would take too long to ramp up (learning the basics of rust, loaning etc) to do a better job here. Pulling in the liveliness utils to print out the router id is ugly as well (but still beats code replication).
I agree with @Yadunund that having the router ids of connected routers logged would be much better, but this is the best I can do in a limited amount of time. Please close if you plan on doing it right in the near future.
Your work on rmw_zenoh is much appreciated.

### DIFF
--- a/rmw_zenoh_cpp/src/zenohd/main.cpp
+++ b/rmw_zenoh_cpp/src/zenohd/main.cpp
@@ -32,6 +32,7 @@
 #include <ament_index_cpp/get_package_share_directory.hpp>
 
 #include "../detail/zenoh_config.hpp"
+#include "../detail/liveliness_utils.hpp"
 
 #include "rmw/error_handling.h"
 
@@ -188,7 +189,9 @@ int main(int argc, char ** argv)
     printf("Unable to open router session!\n");
     return 1;
   }
-
+  printf(
+    "router id: %s\n",
+    rmw_zenoh_cpp::liveliness::zid_to_str(z_info_zid(z_session_loan(&s))).c_str());
 #ifdef _WIN32
   SetConsoleCtrlHandler(quit, TRUE);
 #else
--- a/rmw_zenoh_cpp/src/zenohd/main.cpp
+++ b/rmw_zenoh_cpp/src/zenohd/main.cpp
@@ -190,7 +190,7 @@ int main(int argc, char ** argv)
     return 1;
   }
   printf(
-    "Started Zenoh router with id: %s.\n",
+    "Started Zenoh router with id %s.\n",
     rmw_zenoh_cpp::liveliness::zid_to_str(z_info_zid(z_session_loan(&s))).c_str());
 #ifdef _WIN32
   SetConsoleCtrlHandler(quit, TRUE);
--- a/rmw_zenoh_cpp/src/zenohd/main.cpp
+++ b/rmw_zenoh_cpp/src/zenohd/main.cpp
@@ -190,7 +190,7 @@ int main(int argc, char ** argv)
     return 1;
   }
   printf(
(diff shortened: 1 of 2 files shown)


rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -188,7 +189,9 @@ int main(int argc, char ** argv)
     printf("Unable to open router session!\n");
     return 1;
   }
-
+  printf(
+    "router id: %s\n",
```
Yadunund:
```suggestion
    "Started Zenoh router with id: %s.\n",
```

Yadunund (changes requested):
Thanks a lot for this addition. I think it's the best we can do for now until the zenoh-c API allows one to register a callback that will trigger anytime a peer connects. For now there is [z_info_peers_zid](https://zenoh-c.readthedocs.io/en/0.11.0.2/api.html#c.z_info_peers_zid) but I believe it's a blocking call and does a check for peers at the time of invocation and not throughout the life of the session.

Left a minor suggestion to consider but LGTM otherwise.

JEnoch:
Right, [`z_info_peers_zid()`](https://zenoh-c.readthedocs.io/en/0.11.0.2/api.html#c.z_info_peers_zid) returns the connected peers' ids at the time of call. You could call it periodically to log changes, but that requires to maintain a state and compare at each call.

An alternative is to activate Zenoh logs via the `RUST_LOG` environment variable, that could be set via an option before Session creation in `main.cpp`:
- `RUST_LOG=zenoh=info` for Session initialisation logs (i.e. id and listening endpoints)
- `RUST_LOG=zenoh=info,zenoh_transport=debug` to add connectivity events logging

Side note: reviewing this PR I saw that [`rmw_zenoh_cpp::liveliness::zid_to_str`](https://github.com/ros2/rmw_zenoh/blob/60b131ea969cc9711e08efd9e44e7fef6485d96e/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L222) returns a hexadecimal string assuming `zid` is a big endian u128. Actually in Zenoh it's by convention a little endian u128. 
https://github.com/ros2/rmw_zenoh/pull/190 fixes this.

Yadunund:
Thanks for the information @JEnoch! 

> An alternative is to activate Zenoh logs via the RUST_LOG environment variable, that could be set via an option before Session creation in main.cpp:
RUST_LOG=zenoh=info for Session initialisation logs (i.e. id and listening endpoints)
RUST_LOG=zenoh=info,zenoh_transport=debug to add connectivity events logging

^ I think it's definitely worth adding these to the README for users who want more details.

JEnoch:
Done via https://github.com/ros2/rmw_zenoh/pull/191

@Yadunund  do you think it's good enough, or shall we add to the router a `--debug` option that automatically set `RUST_LOG` in main.cpp (if not already defined) ?
Or better, a cumulative `-v` option setting `RUST_LOG` to different values ?

berndpfrommer:
I think adding the -v option would be a good idea. The additional documentation in #191 is also most welcome.
I'd be happy to see this PR closed as obsolete.

Yadunund:
Adding the `-v` flag as highlighted above sounds great. I'm still good with merging this PR in the meanwhile.

JEnoch:
#197 adds the `-v` to the router.
I think it's still worth merging this PR which prints the router id without `-v`.
Using `-v` will also print it, but along other Zenoh startup logs (listening endpoints).

Yadunund (approved):

PR #192, opened 2024-06-04
Change Node config to listen only on loopback

JEnoch:
In [DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5](https://github.com/ros2/rmw_zenoh/blob/868b64ef3fd64e4f52d76f9933308791a915b6de/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5#L27) no listen endpoints is set. For a `peer` configuration, this implies that Zenoh will by default open a listening TCP socket on a random port (chosen by the OS) and on ANY network interface.
This means that each Node can be directly connected on TCP via any interface, bypassing the router and its possible security configurations (TLS and access control).

This PR changes the default configuration to make each Node to listen for TCP connection only on the loopback interface: `tcp/localhost:0` (port `0` meaning let the OS choose a random port)

### DIFF
--- a/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
+++ b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
@@ -24,7 +24,7 @@
   /// peers, or client can use to establish a zenoh session.
   listen: {
     endpoints: [
-      // "tcp/localhost:7447",
+      "tcp/localhost:0",
     ],
   },
   /// Configure the scouting mechanisms and their behaviours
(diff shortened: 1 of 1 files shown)


rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
```diff
@@ -24,7 +24,7 @@
   /// peers, or client can use to establish a zenoh session.
   listen: {
     endpoints: [
-      // "tcp/localhost:7447",
+      "tcp/localhost:0",
```
Yadunund, line 30: Could you add the explanation in the PR description as a comment above this line?

Yadunund (approved):

PR #197, opened 2024-06-05
-v option for zenohd

JEnoch:
Following discussion in https://github.com/ros2/rmw_zenoh/pull/189 this PR makes `zenohd` to support a `-v` command line option to set the logs verbosity level.

Internally, it set the `RUST_LOG` environment variable a value that depends on the number of `-v` options in the command line.
If `RUST_LOG` is already defined when starting the command, the `-v` options are ignored. This allows user to finely tune the logging if he wants.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,12 +122,14 @@ In this example, the `Zenoh router` will connect to the `Zenoh router` running o
 
 ### Logging
 
-The core of Zenoh is implemented in Rust and uses a logging library that can be configured via a `RUST_LOG` environment variable.
-This variable can be configured independently for each Node and the Zenoh router.
-For instance:
+The core of Zenoh is implemented in Rust and uses a logging library that can be configured via a `RUST_LOG` environment variable. This variable can be configured independently for each Node and for the Zenoh router.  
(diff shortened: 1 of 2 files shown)


clalancette (commented):
Given that we already have the ability to set RUST_LOG as an environment variable, adding the command-line option seems redundant to me.  I feel like it just makes it harder to describe to users.  Can we just document the RUST_LOG environment variable more fully and get rid of the command-line parsing?

JEnoch:
What do you think is missing in the [current documentation of RUST_LOG](https://github.com/ros2/rmw_zenoh/blob/66ec842030fda0cd381835f8ed2577f331fb5729/README.md#logging) ?

clalancette:
> What do you think is missing in the [current documentation of RUST_LOG](https://github.com/ros2/rmw_zenoh/blob/66ec842030fda0cd381835f8ed2577f331fb5729/README.md#logging) ?

The only thing I was thinking was to add the `=trace` portion from this PR.

But we discussed it in the meeting today, and we decided to close this for now.  If you think it is useful to add that to the README, please feel free to open another PR.

PR #203, opened 2024-06-20
Fix race condition and improve performance of rmw_wait

Yadunund:


clalancette:
It turns out that this way of doing things is not really feasible, given the design of `rmw_wait`.  In particular, because this is a per-context condition_variable, it can spuriously wake us up for things that we are not interested in.  That would require us to do a check to see if we should go back to sleep, waiting on something we do care about.

Because of that, I'm going to close out this PR.  #205 is the start of a series of PRs that should fix this in a different way.

PR #204, opened 2024-06-20
Fix checking of the guard condition.

clalancette:
Previously, when we were checking if a guard condition was triggered, we would actually check and reset it. But that means that when we later on went to set the appropriate things in the wait_set, we wouldn't actually set the guard_condition true, since it was false by that point.

Instead, add in two separate methods; the one that just returns whether it is triggered, which we call when checking, and the one that checks and resets, which we call when setting it in the wait_set.

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -3143,14 +3143,12 @@ static bool has_triggered_condition(
   rmw_clients_t * clients,
   rmw_events_t * events)
 {
-  static_cast<void>(events);
-
   if (guard_conditions) {
     for (size_t i = 0; i < guard_conditions->guard_condition_count; ++i) {
       rmw_zenoh_cpp::GuardCondition * gc =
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3143,14 +3143,12 @@ static bool has_triggered_condition(
   rmw_clients_t * clients,
   rmw_events_t * events)
 {
-  static_cast<void>(events);
```
Yadunund, line 3146: wow does this mean we were never actually checking for events?

Yadunund (approved):
Thanks for updating this! I hope you don't mind this push https://github.com/ros2/rmw_zenoh/pull/204/commits/d830d6f5812c908135d841cfb44e5892e6b3207e.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: No, the code for checking events is down below.  We just forgot to remove this (dead) line of code when we actually did implement events.

PR #207, opened 2024-06-21
Improve performance for waits on subs, clients, services, and guards.

clalancette:
Basically by combining the check/attach and the detach/check, we can go from locking 4 times per entity to locking twice per entity, which is a large savings here.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/event.cpp
+++ b/rmw_zenoh_cpp/src/detail/event.cpp
@@ -104,22 +104,6 @@ void EventsManager::trigger_event_callback(rmw_zenoh_event_type_t event_id)
 }
 
 ///=============================================================================
-bool EventsManager::event_queue_is_empty(rmw_zenoh_event_type_t event_id) const
-{
-  if (event_id > ZENOH_EVENT_ID_MAX) {
-    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-      "RMW Zenoh is not correctly configured to handle rmw_zenoh_event_type_t [%d]. "
-      "Report this bug.",
-      event_id);
-    return true;
-  }
-
-  std::lock_guard<std::mutex> lock(event_mutex_);
-
-  return event_queues_[event_id].empty();
-}
-
-///=============================================================================
 std::unique_ptr<rmw_zenoh_event_status_t> EventsManager::pop_next_event(
   rmw_zenoh_event_type_t event_id)
 {
@@ -183,7 +167,7 @@ void EventsManager::add_new_event(
 }
 
 ///=============================================================================
-void EventsManager::attach_event_condition(
+bool EventsManager::queue_has_data_and_attach_condition_if_not(
   rmw_zenoh_event_type_t event_id,
   std::condition_variable * condition_variable)
 {
@@ -192,26 +176,36 @@ void EventsManager::attach_event_condition(
       "RMW Zenoh is not correctly configured to handle rmw_zenoh_event_type_t [%d]. "
       "Report this bug.",
       event_id);
-    return;
+    return false;
   }
 
   std::lock_guard<std::mutex> lock(event_condition_mutex_);
+
+  if (!event_queues_[event_id].empty()) {
+    return true;
+  }
+
   event_conditions_[event_id] = condition_variable;
+
+  return false;
 }
 
 ///=============================================================================
-void EventsManager::detach_event_condition(rmw_zenoh_event_type_t event_id)
+bool EventsManager::detach_condition_and_event_queue_is_empty(rmw_zenoh_event_type_t event_id)
 {
   if (event_id > ZENOH_EVENT_ID_MAX) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "RMW Zenoh is not correctly configured to handle rmw_zenoh_event_type_t [%d]. "
       "Report this bug.",
       event_id);
-    return;
+    return false;
   }
 
   std::lock_guard<std::mutex> lock(event_condition_mutex_);
+
   event_conditions_[event_id] = nullptr;
+
+  return event_queues_[event_id].empty();
 }
 
 ///=============================================================================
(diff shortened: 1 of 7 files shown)


rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -192,26 +176,36 @@ void EventsManager::attach_event_condition(
...
-void EventsManager::detach_event_condition(rmw_zenoh_event_type_t event_id)
+bool EventsManager::detach_condition_and_event_queue_is_empty(rmw_zenoh_event_type_t event_id)
 {
   if (event_id > ZENOH_EVENT_ID_MAX) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "RMW Zenoh is not correctly configured to handle rmw_zenoh_event_type_t [%d]. "
       "Report this bug.",
       event_id);
-    return;
+    return false;
```
Yadunund: Shouldn't this return `true` since a queue for a non-existent event is empty? 

```suggestion
    return true;
```

rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -183,7 +167,7 @@ void EventsManager::add_new_event(
 }
 
 ///=============================================================================
-void EventsManager::attach_event_condition(
+bool EventsManager::queue_has_data_and_attach_condition_if_not(
```
Yadunund, line 170: Suggest to standardize to `queue_is_empty` instead of `queue_has_data`.

```suggestion
bool EventsManager::queue_is_empty_and_attach_condition_if_not(
```

rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -192,26 +176,36 @@ void EventsManager::attach_event_condition(
       "RMW Zenoh is not correctly configured to handle rmw_zenoh_event_type_t [%d]. "
       "Report this bug.",
       event_id);
-    return;
+    return false;
```
Yadunund, line 179: In line with suggestion to rename to `queue_is_empty`

```suggestion
    return true;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -180,24 +182,25 @@ void rmw_subscription_data_t::add_new_message(
...
 ///=============================================================================
-bool rmw_service_data_t::query_queue_is_empty() const
-{
-  std::lock_guard<std::mutex> lock(query_queue_mutex_);
-  return query_queue_.empty();
-}
-
-///=============================================================================
-void rmw_service_data_t::attach_condition(std::condition_variable * condition_variable)
+bool rmw_service_data_t::queue_has_data_and_attach_condition_if_not(
```
Yadunund, line 185: Similarly stick to `queue_is_empty`

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -180,24 +182,25 @@ void rmw_subscription_data_t::add_new_message(
...
+  std::condition_variable * condition_variable)
 {
   std::lock_guard<std::mutex> lock(condition_mutex_);
+  if (!query_queue_.empty()) {
+    return true;
```
Yadunund, line 190:
```suggestion
    return false;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -180,24 +182,25 @@ void rmw_subscription_data_t::add_new_message(
...
+  }
   condition_ = condition_variable;
+
+  return false;
```
Yadunund, line 194:
```suggestion
  return true;
```

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -338,25 +341,25 @@ void rmw_client_data_t::add_new_reply(std::unique_ptr<ZenohReply> reply)
...
-bool rmw_client_data_t::reply_queue_is_empty() const
-{
-  std::lock_guard<std::mutex> lock(reply_queue_mutex_);
-
-  return reply_queue_.empty();
-}
-
-///=============================================================================
-void rmw_client_data_t::attach_condition(std::condition_variable * condition_variable)
+bool rmw_client_data_t::queue_has_data_and_attach_condition_if_not(
```
Yadunund, line 344:
```suggestion
bool rmw_client_data_t::queue_is_empty_and_attach_condition_if_not(
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3286,34 +3287,31 @@ rmw_wait(
     }
   }
 
+  // According to the documentation for rmw_wait in rmw.h, entries in the various arrays that have
+  // *not* been triggered should be set to NULL
+
```
```suggestion
```

Yadunund (changes requested):
This is awesome! Tested with a few applications and works great. Just felt it would be better to be consistent with the naming of functions. Personally I prefer to name these `queue_is_empty..()` everywhere as opposed to some being `queu_has_data` but let me know what you think!

clalancette:
> Personally I prefer to name these `queue_is_empty..()` everywhere as opposed to some being `queu_has_data` but let me know what you think!

I called it `queue_has_data` for two reasons:
1.  To make the conditionals inside of `rmw_wait` use the `true` path, instead of having to negate them.  I find that a bit easier to read.
2.  To better distinguish the "check-and-attach" API calls from the "detach-and-check" API calls.  Just looking at `rmw_subscription_data_t` as an example, we currently have `queue_has_data_and_attach_condition_if_not`, along with the counterpart `detach_condition_and_queue_is_empty`.  If we were to switch everything to "queue is empty", then we would have `queue_is_empty_and_attach_condition_if_so` along with `detach_condition_and_queue_is_empty`.  They are just a bit too similar for my taste.

Those are just the reasons I chose what I did, but I'm not tied to that.  If you feel that `queue_is_empty_and_attach_condition_if_so` is clearer, I can go ahead and change it.  Thoughts?

Yadunund (approved):
Thanks for explaining the rationale. It makes more sense to me now!

clalancette:
> Thanks for explaining the rationale. It makes more sense to me now!

Thanks!  I'm going to go ahead and merge this one in, but in the future if we want to change it, it should be easy enough.  On to the next `rmw_wait` fix :).

PR #210, opened 2024-06-24
Add remaining qos fields to liveliness tokens

Yadunund:
Also include deadline, lifespan and liveliness QoS settings to the liveliness tokens.

With these changes I can get [rcl/test_info_by_topic](https://github.com/ros2/rcl/blob/rolling/rcl/test/rcl/test_info_by_topic.cpp) to pass.


rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -161,13 +173,34 @@ std::vector<std::string> split_keyexpr(
...
+  // Durability.
   keyexpr += std::to_string(qos.durability);
   keyexpr += QOS_DELIMITER;
+  // History.
   keyexpr += std::to_string(qos.history);
-  keyexpr += QOS_HISTORY_DELIMITER;
+  keyexpr += QOS_COMPONENT_DELIMITER;
   keyexpr += std::to_string(qos.depth);
+  // Deadline.
+  keyexpr += QOS_DELIMITER;
```
clalancette: This is going to seem really pedantic, but I'd prefer if this is moved up two lines.  The reason is so that the delimiter is always attached to the end of the thing it is delimiting.  At the end of the day, it won't matter too much, but I find it easier to read:

```suggestion
  keyexpr += std::to_string(qos.depth);
  keyexpr += QOS_DELIMITER;
  // Deadline.
```

(alternatively, you could leave this QOS_DELIMITER where it is, but then change the Reliability, Durability, and History QOS_DELIMITER to be in the one below it.  Either is fine, I just want it to be consistent one way or the other)

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -161,13 +173,34 @@ std::vector<std::string> split_keyexpr(
 std::string qos_to_keyexpr(rmw_qos_profile_t qos)
 {
   std::string keyexpr = "";
```
clalancette, line 175: One overall thought I'm having here: would it be more performant to assume a QoS, and then only send the "differences" in the liveliness token here?

That is, we start out by assuming that all topics have a QoS of RELIABLE, VOLATILE, KEEP_LAST, DEPTH 10, DEADLINE 0, LIFESPAN 0, LIVELINESS POLICY 0, LIVELINESS LEASE 0.  If the QoS is exactly that, we put nothing in the QoS field here.  If the topic happens to use BEST_EFFORT, but the rest of the QoS is the same, then we only have to encode and send BEST_EFFORT.

This would be a lot more work to parse on the receiving side, so I'm not sure it is worth it.  But I'm curious what your thoughts are (and maybe what @JEnoch thinks about it).

clalancette (changes requested):
I have one minor comment, and one thought for improving the efficiency in the future.  Feel free to punt on improving the efficiency, but if you do that I suggest making an issue about it.

Otherwise, this looks great!

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: This is a great idea. I was planning to work on a separate PR to drop the reliance on [rmw_dds_common::qos_profile_get_best_available_for_topic_publisher](https://github.com/ros2/rmw_zenoh/blob/61f2a704a17e6e45834274b15cff72a53bd00554/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L552-L553) and other such functions within our codebase by having our own implementation of a SYSTEM_DEFAULT qos. 
I could combine that effort with the suggestion above to only transmit QoS info in the liveliness token if there is a deviation from the default.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Totally valid feedback. Done https://github.com/ros2/rmw_zenoh/pull/210/commits/17f841924d1ccf092d97c55809e26202c9b436b6

clalancette (approved):
Looks great to me!

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
JEnoch: Yes, that's what I did in the liveliness tokens for `zenoh-plugin-ros2dds`:
- [`qos_to_key_expr()`](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/blob/a742b36d34953fda689a1710ab1dd923279a784e/zenoh-plugin-ros2dds/src/liveliness_mgt.rs#L197): if a QoS has default value (i.e. `== None`) it's an empty string between 2 separators (`:` char)
- [`key_expr_to_qos()`](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/blob/a742b36d34953fda689a1710ab1dd923279a784e/zenoh-plugin-ros2dds/src/liveliness_mgt.rs#L229): [`str::split(':')`](https://doc.rust-lang.org/std/primitive.str.html#method.split) returns an iterator on each "QoS as a string representation". If the string is empty, that's the default QoS value.

PR #211, opened 2024-06-25
Fix a race condition in rmw_wait.

clalancette:
In the current code, there is a race condition in rmw_wait. This happens because of the following sequence:

1.  Without taking a lock, we check if any of the entities in the wait_set are ready, and if not attach the condition_variable to it.
2.  Then we take the lock, and go to sleep on the condition_variable.

However, doing step 1 takes time, and we check in a particular order: guard_conditions, events, subscriptions, services, clients.  The race happens because a subscription may come in after we've checked subscriptions in the above list (while we are checking services and clients).  In that case, we'll unnecessarily go to sleep on the condition_variable, even though something is ready.  This increases the latency.

To solve the issue, we still do all of the checking and attaching unlocked.  However, we update the "notify" method of each entity so that it takes the condition_variable lock, and in addition to kicking the condition_variable it sets a boolean in the wait_set structure to "true".  Then, in rmw_wait(), after we have finished attaching, we take the lock, and set a predicate on the condition_variable so that it will quit if wait_set_data->triggered is true.  Thus, if one of the entities became ready after we checked, we'll notice it and not go to sleep.

This also allows us to deal with "spurious" wakeups, where the condition_variable was woken up even though nothing in this wait_set became ready.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Open Source Robotics Foundation, Inc.
+// Copyright 2024 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -25,6 +25,29 @@ namespace rmw_zenoh_cpp
 
 struct rmw_wait_set_data_t
 {
+  // The combination of condition_variable, condition_mutex, and triggered are used to make sure
+  // there isn't a race in rmw_wait().  That could happen because of the following sequence:
+  //
+  // 1. Without taking a lock, rmw_wait() checks if any of the entities in the wait_set are ready,
+  //    and if not attachs the condition_variable to it.
+  // 2. It then takes the lock, and sleeps on the condition_variable.
+  //
+  // However, doing step 1 takes time, and checks in a particular order: guard_conditions, events,
+  // subscriptions, services, clients. The race could happen because a subscription may come in
+  // subscriptions have been checked in the above list (while services and clients are being
+  // checked). In that case, rmw_wait() will unnecessarily go to sleep on the condition_variable,
+  // even though something is ready. This increases the latency.
+  //
+  // To solve the issue, rmw_wait() still does all of the checking and attaching unlocked. However,
+  // the "notify" method of each entity takes the condition_mutex lock, and in addition to kicking
+  // the condition_variable it sets "triggreed" in this structure to "true". Then after rmw_wait()
+  // has finished attaching, it takes the lock and sets a predicate on the condition_variable so
+  // that it will quit if wait_set_data->triggered is true. Thus, if one of the entities became
+  // ready after we checked, the condition_variable will notice it and not go to sleep.
+  //
+  // This also deals with "spurious" wakeups, where the condition_variable was woken up even though
+  // nothing in this wait_set became ready.
+
   std::condition_variable condition_variable;
   std::mutex condition_mutex;
 
--- a/rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
@@ -1,0 +1,37 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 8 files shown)


rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
clalancette, line 1: I had to split this out into its own file to avoid a circular `#include` between `rmw_data_types.hpp` and `events.hpp`.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3416,13 +3410,22 @@ rmw_wait(
...
-          lock, std::chrono::nanoseconds(wait_timeout->nsec + RCUTILS_S_TO_NS(wait_timeout->sec)));
+          lock,
+          std::chrono::nanoseconds(wait_timeout->nsec + RCUTILS_S_TO_NS(wait_timeout->sec)),
+          [wait_set_data]() {return wait_set_data->triggered;});
       }
     }
+
+    // We reset here (while still holding the lock) because we only cared about this through this
+    // particular call.
+    wait_set_data->triggered = false;
```
clalancette: This is a very important line of code to note.  If we forget to do this, then *every* time we come in here, we'll skip the wait, which is obviously bad for performance.

I should point out that there is another way to do this, which is to make `triggered` a stack variable and pass that into all of the `attach_condition` calls.  While that would avoid this particular problem, it seemed unnecessary, so I didn't do it.  If we feel like that is a better way to go, I can plumb that in.

clalancette:
I should also mention the invaluable thoughts of @wjwwood and @mjcarroll when coming up with this.

rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
```diff
@@ -0,0 +1,37 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
```
Yadunund:
```suggestion
// Copyright 2024 Open Source Robotics Foundation, Inc.
```

rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
```diff
@@ -0,0 +1,37 @@
...
+#define DETAIL__RMW_WAIT_SET_DATA_HPP_
+
+#include <condition_variable>
+#include <mutex>
+
+#include "rmw/rmw.h"
+
+namespace rmw_zenoh_cpp
+{
+
```
Yadunund, line 25: Could we add the text from your PR description as a docstring here? It will be very useful.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: > I should point out that there is another way to do this, which is to make triggered a stack variable and pass that into all of the attach_condition calls. While that would avoid this particular problem, it seemed unnecessary, so I didn't do it. If we feel like that is a better way to go, I can plumb that in.

Personally I think it's okay leaving it as is but we could add this as a potential todo comment?

Yadunund (commented):
This is awesome! The implementation makes so much sense in retrospect but the rmw API does not make this apparent 😅 

Minor comments before we can merge!

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: > Personally I think it's okay leaving it as is but we could add this as a potential todo comment?

I'll put a comment in there explaining this, but not a TODO.  It's not clear to me that we'll ever change it.

clalancette: Fixed in [b6074f4](https://github.com/ros2/rmw_zenoh/pull/211/commits/b6074f42a8534cc8683addbb13ca66c58e736d72)

rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
clalancette: Added in [b6074f4](https://github.com/ros2/rmw_zenoh/pull/211/commits/b6074f42a8534cc8683addbb13ca66c58e736d72)

rmw_zenoh_cpp/src/detail/rmw_wait_set_data.hpp
clalancette: Fixed in [b6074f4](https://github.com/ros2/rmw_zenoh/pull/211/commits/b6074f42a8534cc8683addbb13ca66c58e736d72)

clalancette:
Ready for another round of review!

Yadunund (approved):

PR #217, opened 2024-06-27
Cleanup the implementation of rmw_init_options_copy.

clalancette:
In particular, make sure to properly clean up if we fail partway through any of the steps.

I noticed this while reviewing the code for `rmw_init_options_copy`.


Yadunund (approved):
Nice catch!

PR #219, opened 2024-06-27
Make sure to *not* undeclare the liveliness token for clients.

clalancette:
That is, we were forgetting to cancel the undeclare of the liveliness token in client initialization, so we would create the token and immediately remove it.  Fix that here, so it only gets removed at rmw_destroy_client time.


Yadunund (approved):
Thanks for the fix!

PR #221, opened 2024-06-27
Implement rmw_take_sequence.

clalancette:
While this isn't currently used by rclcpp and rclpy, it *is* used inside of rcl, specifically in testing.  Thus we need to implement it to pass all rcl tests.

To reduce code duplication, this commit refactors __rmw_take() to handle this case.  In particular, we rename __rmw_take() to __rmw_take_one(), with no error checking.  Then we change rmw_take() and rmw_take_with_info() to do the error checking and call __rmw_take_one().  Finally we implement rmw_take_sequence() by calling __rwm_take_one() in a loop.

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -1701,37 +1701,17 @@ rmw_subscription_get_content_filter(
   return RMW_RET_UNSUPPORTED;
 }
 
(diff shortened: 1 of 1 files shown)


Yadunund (approved):
The implementations looks great to me!

On a side note: I wonder if there is a way to align the datatypes of the`taken` arg in `rmw_take_with_info` and `rmw_take_sequence` esp given one is `bool * ` and another as `size_t`. The latter should be an array of bools imo.

PR #223, opened 2024-06-29
Switch to custom logging macros to avoid deadlocks

Yadunund:
Address logging related deadlock discovered in https://github.com/ros2/rmw_zenoh/issues/182.

@clalancette I went with a more C++ implementation here by relying on the popular [fmt](https://github.com/fmtlib/fmt/) library which is apparently more performant than `printf` and `ostringstream `. Let me know if you prefer following `rcutils/logging.h` type of implementation with a specifiable allocator. We could also directly invoke [rcutils_logging_console_output_handler](https://github.com/ros2/rcutils/blob/f69abc1f9d5871843df2ad3870faed43cf35df6e/src/logging.c#L1373).

### DIFF
--- a/rmw_zenoh_cpp/src/detail/logging.hpp
+++ b/rmw_zenoh_cpp/src/detail/logging.hpp
@@ -39,28 +39,7 @@ public:
     RCUTILS_LOG_SEVERITY level,
     const char * name,
     const char * message,
-    ...) const
-  {
-    if (level >= threshold_level_) {
-      rcutils_time_point_value_t now;
-      rcutils_ret_t ret = rcutils_system_time_now(&now);
-      if (ret != RCUTILS_RET_OK) {
-        RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to get timestamp while doing a console logging.\n");
-        return;
-      }
-      static rcutils_log_location_t log_location = {__func__, __FILE__, __LINE__};
-      va_list args;
-      va_start(args, message);
-      rcutils_logging_console_output_handler(
-        &log_location,
-        level,
-        name,
-        now,
-        message,
-        &args
-      );
-    }
-  }
+    ...) const;
 
 private:
   RCUTILS_LOG_SEVERITY threshold_level_;
--- a/rmw_zenoh_cpp/src/detail/logging.hpp
+++ b/rmw_zenoh_cpp/src/detail/logging.hpp
@@ -1,0 +1,84 @@
+// Copyright 2024 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DETAIL__LOGGING_HPP_
+#define DETAIL__LOGGING_HPP_
+
+#include <fmt/chrono.h>
+#include <fmt/color.h>
+#include <fmt/printf.h>
+
+#include <iostream>
+#include <mutex>
+#include <string>
+#include <sstream>
+
+namespace rmw_zenoh_cpp
+{
+///=============================================================================
+enum LogLevel
+{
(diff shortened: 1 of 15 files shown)


clalancette:
> @clalancette I went with a more C++ implementation here by relying on the popular [fmt](https://github.com/fmtlib/fmt/) library which is apparently more performant than `printf` and `ostringstream `

The problem with bringing in this dependency is that we'll have to deal with it on Windows.  Also, in general, the more dependencies this has, the harder it will be to get `rmw_zenoh` into the core.  Finally, my understanding is that either C++-20 or C++-23 will have this directly, so we can get this "for free" when we upgrade in the future.  For all of those reasons, I think we should stick with printf-style logging for now.

> We could also directly invoke [rcutils_logging_console_output_handler](https://github.com/ros2/rcutils/blob/f69abc1f9d5871843df2ad3870faed43cf35df6e/src/logging.c#L1373).

That's a really good idea, I'd prefer that.  While it won't get us *all* of the features of the `rcutils` logger, it will at least move us in that direction.

Yadunund:
Thanks for the quick feedback! I'll update to rely on `rcutils` functions.

Yadunund:
@clalancette I've updated accordingly. 

I've ran the problematic case from #182 several times and it works most of the time*.
```bash
# terminal 1
ros2 run rclcpp_components component_container
```

```bash
# terminal 2
for i in {1..100}; do echo "ros2 component load --node-namespace /talker${i} /ComponentManager composition composition::Talker"; done | parallel
```

`*` once in a while, the command in terminal 2 does not exit cleanly. However running `ros2 node list | wc -l` returns `101` indicating that 100 nodes were indeed injected into the component container. All topics are also available. So perhaps the issue there is related to shutdown?

rmw_zenoh_cpp/src/detail/logging.hpp
```diff
@@ -0,0 +1,71 @@
...
+      rcutils_logging_console_output_handler(
+        &log_location,
+        level,
+        name,
+        now,
+        message,
+        &args
+      );
+    }
+  }
```
clalancette: Can we move this entire implementation into logging.cpp?  That way not every file that includes this will have to compile this code.

rmw_zenoh_cpp/src/detail/logging_macros.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "logging.hpp"
+
+///=============================================================================
+// We define custom logging marcos to log to console as it was discovered
+// that relying on RCUTILS_LOG_X_NAMED functions also end up logging over /rosout
+// which can lead to deadlocks in rmw_zenoh especially when multiple threads
+// invoke GraphCache::parse_put() and GraphCache::parse_del() functions.
+// See https://github.com/ros2/rmw_zenoh/issues/182 for more details.
+// Note: Consider adding a set_rmw_log_level() API to rmw which can override the
+// default log level at the RMW layer.
```
clalancette: This actually already exists: https://github.com/ros2/rmw_zenoh/blob/0eb456032355a2be9965e5171e09286046c8a0a1/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L3874-L3879 .

We just don't currently implement it, and I don't believe `rcl` or anything else currently calls it.

clalancette (changes requested):
Overall, this looks great!  I've left a couple of pieces of minor feedback inline.

One other thing we might want to consider is adding a GitHub action here to check if a PR contains `RCUTILS_LOG_*`, or does `#include <rcutils/logging_macros.h>`.  That would help ensure that we don't add these by accident in the future.

rmw_zenoh_cpp/src/detail/logging.hpp
Yadunund: Whoops forgot to do that after getting rid of the templated function. Done https://github.com/ros2/rmw_zenoh/pull/223/commits/afe137813e93cd4c3a85524c7d2067fdd9ec6880

rmw_zenoh_cpp/src/detail/logging_macros.hpp
Yadunund: Oh nice! I've implemented it in https://github.com/ros2/rmw_zenoh/pull/223/commits/73af6701512578f96103e2060aa5bd6ce3ebb2d7 since connext and fastrtps implement it too.

clalancette (approved):
This is fantastic!  I'll note that the new `check_logging.yaml` job is currently failing, because of course this PR *does* contain RCUTILS_LOG macros (that we are removing).  It would be nice if there was a `diffDoesNotAdd` configuration variable, but after this PR all further ones shouldn't have this problem.  So I'm going ahead and approving this.

PR #226, opened 2024-07-01
Move the event_map out of the header file.

clalancette:
The main reason for this is that putting it in the header file means that it is duplicated into every compilation unit that includes it.  This is slow, and in theory can cause issues if a pointer to the map is passed between compilation units.

Instead, hide the map inside of the events.cpp file, and add a free function that converts between an RMW event type and a Zenoh event type.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/event.cpp
+++ b/rmw_zenoh_cpp/src/detail/event.cpp
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <deque>
+#include <memory>
+#include <mutex>
+#include <unordered_map>
 #include <utility>
 
 #include "event.hpp"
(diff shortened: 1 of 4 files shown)


rmw_zenoh_cpp/src/detail/event.hpp
```diff
@@ -53,17 +51,7 @@ enum rmw_zenoh_event_type_t
 /// Helper value to indicate the maximum index of events supported.
 #define ZENOH_EVENT_ID_MAX rmw_zenoh_event_type_t::ZENOH_EVENT_PUBLICATION_MATCHED
```
Yadunund, line 52: Should we move this to the translation unit too?

rmw_zenoh_cpp/src/detail/event.cpp
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <deque>
+#include <memory>
```
Yadunund, line 16: Do we need to specify these headers again if they're included in the header file?

Yadunund (approved):

rmw_zenoh_cpp/src/detail/event.hpp
clalancette: As it stands, this is still needed by graph_cache.cpp.  I don't see a good way to remove this right now, so I'd prefer to leave it.

rmw_zenoh_cpp/src/detail/event.cpp
clalancette: > Do we need to specify these headers again if they're included in the header file?

My general thinking here is that yes, if the implementation uses the signatures, it should also include it.  Essentially I believe in "include-what-you-use" in both the header file and the implementation file.

rmw_zenoh_cpp/src/detail/event.hpp
Yadunund: Sounds good.

rmw_zenoh_cpp/src/detail/event.cpp
Yadunund: Gotcha

clalancette:
Going ahead and merging this one, thanks for the review!

Issue #186, opened 2024-05-29
malloc(): unaligned tcache chunk detected error


CihatAltiparmak:
## Tested On
* ROS VERSION: ros iron from docker image
* latest version of rmw_zenoh_cpp

## Description 
Hello, 
I want to report below bug occured when is run the tests of [ros2_control](https://github.com/ros-controls/ros2_control) and [ros2_controllers](https://github.com/ros-controls/ros2_controllers) by using [this dockerfile](https://gist.github.com/CihatAltiparmak/2a660fc25af9aba23c54aa2e3882be38#file-ros2_control_testing-dockerfile). It seems like error occurs in [this line of one of joint_trajectory_controller package's test files](https://github.com/ros-controls/ros2_controllers/blob/d2e926be5d1ecc31fc5966322c59afdd17bb1c3f/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L212) . I tried to look at all issues to find relevant issue but i couldn't find so i decided to open a new issue here. For more, you can also look at [this comment](https://github.com/moveit/moveit2/issues/2844#issuecomment-2135924753)

```
[Thread 0x7904777fe640 (LWP 20062) exited]
[       OK ] OnlyEffortTrajectoryControllers/TestTrajectoryActionsTestParameterized.test_allow_nonzero_velocity_at_trajectory_end_true/0 (2017 ms)
[ RUN      ] OnlyEffortTrajectoryControllers/TestTrajectoryActionsTestParameterized.test_allow_nonzero_velocity_at_trajectory_end_true/1
malloc(): unaligned tcache chunk detected

Thread 1 "test_trajectory" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=133060925205056) at ./nptl/pthread_kill.c:44
44	./nptl/pthread_kill.c: No such file or directory.
(gdb) backtrace 
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=133060925205056) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=133060925205056) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=133060925205056, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007904a997b476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
[...]

PR #228, opened 2024-07-01
Fix replies coming in after a client is destroyed.

clalancette:
It turns out that in Zenoh, once a query is in-flight (via
the z_get zenoh-c API), there is no way to cancel it.  With
that in mind, consider a sequence of events:
    
1.  rmw_create_client
2.  rmw_send_request
3.  rmw_destroy_client
    
If all of that happens before the query callback is called,
we'll have freed up the structure that we passed to the zenoh
closure (client_data), and thus the structure will access
already freed memory.
    
The fix for this is relatively complicated.  First of all,
anytime that there is a call to rmw_send_request (which calls z_get()),
it increases a per-client counter.  rmw_destroy_client() checks to
to see if that counter is 0.  If it is, we know both that there are
no queries in flight, and also that there will be no new ones (because
we are cleaning the client up).  Thus the structure is freed directly.
If the counter is greater than 0, then there is at least one query
in flight and it is not safe to free the structure.  However, the
client_data structure is marked as "shutdown" at this point.  There
will not be any *new* requests for this client, but the in-flight ones
still need to be dealt with.
    
For the in-flight ones, every time client_data_handler() (the
rmw_zenoh_cpp callback for queries) is called, it first decrements the
number of in-flight queries.  If the client is shutdown, and the number
of in-flight queries is 0, then it is safe to free the client_data
structure.  If the client is shutdown but there are other queries in
flight, no actual work is done except for the decrement.
    
There is one case which is not handled here at all, and that has
to do with timeouts.  Currently the rmw_zenoh_cpp client query timeout
is set to essentially infinite.  Thus, if a query is in-flight, but never
returns, the memory corresponding to that client will be leaked.
However, this is already an existing problem; this patch changes that
from a UB to a memory leak.

In my testing, this fixes the problems pointed out in #186

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -388,6 +388,36 @@ std::unique_ptr<ZenohReply> rmw_client_data_t::pop_next_reply()
 }
 
 //==============================================================================
+// See the comment about the "num_in_flight" class variable in the rmw_client_data_t class
+// for the use of this method.
+void rmw_client_data_t::increment_in_flight_callbacks()
+{
+  std::lock_guard<std::mutex> lock(in_flight_mutex_);
+  num_in_flight_++;
+}
+
+//==============================================================================
+// See the comment about the "num_in_flight" class variable in the rmw_client_data_t class
+// for the use of this method.
+bool rmw_client_data_t::shutdown_and_query_in_flight()
+{
+  std::lock_guard<std::mutex> lock(in_flight_mutex_);
+  is_shutdown_ = true;
+
+  return num_in_flight_ > 0;
+}
+
+//==============================================================================
+// See the comment about the "num_in_flight" class variable in the rmw_client_data_t structure
+// for the use of this method.
+bool rmw_client_data_t::decrement_queries_in_flight_and_is_shutdown(bool & queries_in_flight)
+{
+  std::lock_guard<std::mutex> lock(in_flight_mutex_);
+  queries_in_flight = --num_in_flight_ > 0;
+  return is_shutdown_;
+}
+
+//==============================================================================
 void sub_data_handler(
   const z_sample_t * sample,
   void * data)
@@ -525,6 +555,20 @@ void client_data_handler(z_owned_reply_t * reply, void * data)
     );
     return;
   }
+
+  // See the comment about the "num_in_flight" class variable in the rmw_client_data_t class for
+  // why we need to do this.
+  bool queries_in_flight = false;
+  bool is_shutdown = client_data->decrement_queries_in_flight_and_is_shutdown(queries_in_flight);
+
+  if (is_shutdown) {
+    if (!queries_in_flight) {
+      client_data->context->options.allocator.deallocate(
+        client_data, client_data->context->options.allocator.state);
+    }
+    return;
+  }
+
   if (!z_reply_check(reply)) {
     RMW_ZENOH_LOG_ERROR_NAMED(
       "rmw_zenoh_cpp",
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -28,6 +28,7 @@
 #include "rcpputils/scope_exit.hpp"
 
 #include "rmw/error_handling.h"
+#include "rmw/impl/cpp/macros.hpp"
 
 #include "attachment_helpers.hpp"
 #include "rmw_data_types.hpp"
@@ -417,6 +418,12 @@ bool rmw_client_data_t::decrement_queries_in_flight_and_is_shutdown(bool & queri
   return is_shutdown_;
 }
 
+bool rmw_client_data_t::is_shutdown() const
+{
+  std::lock_guard<std::mutex> lock(in_flight_mutex_);
+  return is_shutdown_;
+}
+
 //==============================================================================
 void sub_data_handler(
   const z_sample_t * sample,
@@ -558,14 +565,7 @@ void client_data_handler(z_owned_reply_t * reply, void * data)
 
   // See the comment about the "num_in_flight" class variable in the rmw_client_data_t class for
   // why we need to do this.
-  bool queries_in_flight = false;
-  bool is_shutdown = client_data->decrement_queries_in_flight_and_is_shutdown(queries_in_flight);
-
-  if (is_shutdown) {
-    if (!queries_in_flight) {
-      client_data->context->options.allocator.deallocate(
-        client_data, client_data->context->options.allocator.state);
-    }
+  if (client_data->is_shutdown()) {
     return;
   }
 
@@ -594,4 +594,30 @@ void client_data_handler(z_owned_reply_t * reply, void * data)
   // Since we took ownership of the reply, null it out here
   *reply = z_reply_null();
 }
+
+void client_data_drop(void * data)
+{
+  auto client_data = static_cast<rmw_client_data_t *>(data);
+  if (client_data == nullptr) {
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to obtain client_data_t "
+    );
+    return;
+  }
+
+  // See the comment about the "num_in_flight" class variable in the rmw_client_data_t class for
+  // why we need to do this.
+  bool queries_in_flight = false;
+  bool is_shutdown = client_data->decrement_queries_in_flight_and_is_shutdown(queries_in_flight);
+
+  if (is_shutdown) {
+    if (!queries_in_flight) {
+      RMW_TRY_DESTRUCTOR(client_data->~rmw_client_data_t(), rmw_client_data_t, );
+      client_data->context->options.allocator.deallocate(
+        client_data, client_data->context->options.allocator.state);
+    }
+  }
+}
+
 }  // namespace rmw_zenoh_cpp
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -2378,7 +2378,6 @@ rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
     return RMW_RET_INVALID_ARGUMENT);
 
   // CLEANUP ===================================================================
-  z_drop(z_move(client_data->zn_closure_reply));
   z_drop(z_move(client_data->keyexpr));
   zc_liveliness_undeclare_token(z_move(client_data->token));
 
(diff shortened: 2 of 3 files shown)


clalancette:
I should note that this was on the advice of the Zenoh developers in Discord.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2464,11 +2463,13 @@ rmw_send_request(
...
   opts.value.payload = z_bytes_t{data_length, reinterpret_cast<const uint8_t *>(request_bytes)};
-  client_data->zn_closure_reply =
+  z_owned_closure_reply_t zn_closure_reply =
     z_closure(rmw_zenoh_cpp::client_data_handler, nullptr, client_data);
   z_get(
-    z_loan(context_impl->session), z_loan(
-      client_data->keyexpr), "", &client_data->zn_closure_reply, &opts);
+    z_loan(context_impl->session),
+    z_loan(client_data->keyexpr), "",
+    z_move(zn_closure_reply),
```
Yadunund, line 2511: Wondering if it makes sense for the `rmw_client` to own the callback. What happens when the callback does not get triggered yet (the service is busy), but then the client is destroyed? 

Would it be better to keep the original implementation but ensure we free the callback with `z_reply_null` when we call `rmw_client_fini`?

According to the [z_get documentation](https://zenoh-c.readthedocs.io/en/0.11.0/api.html#c.z_get):
> callback – The callback function that will be called on reception of replies for this query. Note that the reply parameter of the callback is passed by mutable reference, but will be dropped once your callback exits to help you avoid memory leaks. If you’d rather take ownership, please refer to the documentation of [z_reply_null()](https://zenoh-c.readthedocs.io/en/0.11.0/api.html#c.z_reply_null)

Will closing the Zenoh session ensure all callbacks are freed?
cc: @JEnoch

clalancette: > Wondering if it makes sense for the `rmw_client` to own the callback. What happens when the callback does not get triggered yet (the service is busy), but then the client is destroyed?

Right, so that is a good question.  And the answer is that it is up to `rmw_zenoh_cpp` to deal with this situation, by not freeing the underlying memory until all in-flight callbacks are accounted for.  I'll point out that this is already a problem in `rmw_zenoh_cpp`; if a callback for a query comes in after `rmw_destroy_client` has been called, we'll access freed memory.  I have an upcoming patch which fixes that, which is built on top of this.  If you think it is better, I can combine that into this PR.

> Would it be better to keep the original implementation but ensure we free the callback with `z_reply_null` when we call `rmw_client_fini`?

I'm not 100% sure, but I believe that this is a different thing.  This PR is concerned with the ownership of the closure itself.  To see what I am talking about, consider that in almost all other places where we call into the Zenoh API, we use either `z_loan` or `z_move`.  The code before this PR was using a raw pointer reference, which is a code smell.

What you are talking about is the ownership of the reply.  There, we are already taking ownership of the memory during the callback (via `z_reply_null`), so I don't think we need to make a change.  Does that make sense?

Also, it might be worthwhile to get feedback from @JEnoch and @OlivierHecart, since they were the ones that suggested this to me to begin with.

Yadunund: Thanks for clarifying. I think it would be good to combine those changes in this PR as well.

> What you are talking about is the ownership of the reply.

Ah yeah sorry I got things mixed up but my original concern still stands on whether the callback gets freed if the client is destroyed before we receive a reply.

clalancette: > Ah yeah sorry I got things mixed up but my original concern still stands on whether the callback gets freed if the client is destroyed before we receive a reply.

This is a question for the Zenoh developers, but I believe not.  That's the reason for doing the `z_move`, so that zenoh-c/Zenoh proper is responsible for the lifetime of the closure.  I'll also point out that in my tests with valgrind, this didn't show a problem.

clalancette:
OK, so I've now updated this PR to include all of the fixes necessary to fix our use-after-free bug here.  I've also updated the title and the initial description to explain what all of this is doing.  And there is another comment in the code explaining why we need all of this.  Please take another look and let me know what you think.

Yadunund (approved):
Wow this is quite the tricky problem. Thanks for finding this suitable fix.

clalancette:
> Wow this is quite the tricky problem. Thanks for finding this suitable fix.

Yeah, agreed.  This one is tricky.  Let's hold off on merging this one for a little bit to give the Zenoh developers a chance to take a look and comment.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2464,11 +2467,17 @@ rmw_send_request(
...
+    &opts);
+
+  // See the comment about the "num_in_flight" class variable in the rmw_client_data_t class for
+  // why we need to do this.
+  client_data->increment_in_flight_callbacks();
```
OlivierHecart: Shouldn't `increment_in_flight_callbacks` be called before `z_get` so that we are sure `rmw_destroy_client` is not called in parallel between `z_get` and `increment_in_flight_callbacks` ? I don't know the rest of the code enough to know if `rmw_destroy_client` and `rmw_send_request`can be called concurrently or not.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2464,11 +2467,17 @@ rmw_send_request(
   // and any number.
   opts.consolidation = z_query_consolidation_latest();
   opts.value.payload = z_bytes_t{data_length, reinterpret_cast<const uint8_t *>(request_bytes)};
-  client_data->zn_closure_reply =
+  z_owned_closure_reply_t zn_closure_reply =
```
OlivierHecart, line 2506: Shouldn't we check that `is_shutdown_` is `false` before issuing new queries for this client ? (Maybe there is another check somewhere else that already prevents that I'm not aware of)

rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
```diff
@@ -522,6 +552,20 @@ void client_data_handler(z_owned_reply_t * reply, void * data)
     );
     return;
   }
+
```
OlivierHecart, line 565: This decrement should be done in the `drop` function of the `z_owned_closure_reply_t`. There is no guarantee that the `call` function will be called only once per query, while `drop` will be called once per query at the end.

OlivierHecart (changes requested):
`num_in_flight_` should be decremented in the drop function of the `z_owned_closure_reply_t`.

clalancette:
> `num_in_flight_` should be decremented in the drop function of the `z_owned_closure_reply_t`.

Thanks.  This was a very helpful comment; I didn't realize that `drop` was one of the parameters to the closure creation.  That is absolutely the right place to free the memory.

However, looking at this closer leads me to another question.  As you can see in this PR, we create a new closure and give ownership of it to Zenoh via `z_get` for every query:

```
  z_owned_closure_reply_t zn_closure_reply =
    z_closure(rmw_zenoh_cpp::client_data_handler, nullptr, client_data);
  z_get(
    z_loan(context_impl->session),
    z_loan(client_data->keyexpr), "",
    z_move(zn_closure_reply),
    &opts);
```

(this is more-or-less required by the `z_get` API).  But the documentation in `zenoh-c` about closures says this:

```
 * It is guaranteed that:
 *
 *   - `call` will never be called once `drop` has started.
 *   - `drop` will only be called **once**, and **after every** `call` has ended.
 *   - The two previous guarantees imply that `call` and `drop` are never called concurrently.
```

That suggests to me that it is possible to use the same closure in multiple `z_get` calls.  Is that correct?  Would it be better for us to create the closure when we initialize the client (`rmw_create_client`), and then reuse it for every `z_get` with this client (`rmw_send_request`)?  If so, what would that look like, since the API seems to require us to `z_move` the closure?

OlivierHecart:
> > `num_in_flight_` should be decremented in the drop function of the `z_owned_closure_reply_t`.
> 
> Thanks. This was a very helpful comment; I didn't realize that `drop` was one of the parameters to the closure creation. That is absolutely the right place to free the memory.
> 
> However, looking at this closer leads me to another question. As you can see in this PR, we create a new closure and give ownership of it to Zenoh via `z_get` for every query:
> 
> ```
>   z_owned_closure_reply_t zn_closure_reply =
>     z_closure(rmw_zenoh_cpp::client_data_handler, nullptr, client_data);
>   z_get(
>     z_loan(context_impl->session),
>     z_loan(client_data->keyexpr), "",
>     z_move(zn_closure_reply),
>     &opts);
> ```
> 
> (this is more-or-less required by the `z_get` API). But the documentation in `zenoh-c` about closures says this:
> 
> ```
>  * It is guaranteed that:
>  *
>  *   - `call` will never be called once `drop` has started.
>  *   - `drop` will only be called **once**, and **after every** `call` has ended.
>  *   - The two previous guarantees imply that `call` and `drop` are never called concurrently.
> ```
> 
> That suggests to me that it is possible to use the same closure in multiple `z_get` calls. Is that correct? Would it be better for us to create the closure when we initialize the client (`rmw_create_client`), and then reuse it for every `z_get` with this client (`rmw_send_request`)? If so, what would that look like, since the API seems to require us to `z_move` the closure?

I believe that what creates the confusion here is that in Zenoh one single `z_get` can lead to 0, 1 or multiple replies. Which results in the API to 0, 1 or multiple calls to `call` and finally one call to `drop` to indicate to the user that there are no more replies for this single `z_get`. With that in mind, the documentation does not necessarily suggests that the same closure can be used in multiple `z_get`. 
So I believe that you indeed should not reuse the same closure for multiple `z_get`. I'll double check that and come back to you if otherwise but for now you can go with this assumption.

clalancette:
> So I believe that you indeed should not reuse the same closure for multiple `z_get`. I'll double check that and come back to you if otherwise but for now you can go with this assumption.

Fantastic, thank you for that explanation.  I'll continue with that assumption, then.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: This is a good point.  We probably actually have a larger problem here in that we don't have any kind of locking at all for concurrent calls of the APIs related to clients (and servers, and subscriptions, and publishers).  That is something we should look into.

For now, I am going to move this `increment` ahead of the `z_get` call as you suggest.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: Yes, good point.  I'll add this in as well.

clalancette:
All right, I believe I responded to all comments now.  Please take another look when you get a chance.

OlivierHecart (approved):

PR #233, opened 2024-07-01
Move all private functions to use an anonymous namespace.

### LINKED ISSUE #124: Revisit anonymous namespaces vs. static functions
C++ has two different ways to create compilation-unit-local functions: using an anonymous namespace (the modern C++ way), or by putting `static` in front of the function (the C way).

Currently in rmw_zenoh_cpp we use a mixture of these two styles.  We should be consistent here, and just choose one.  My personal preference is for `static` in front of the function, as it is easier to tell at a glance that a function is local-only.  But I'm open to other opinions.

This comes out of a discussion in https://github.com/ros2/rmw_zenoh/pull/123#discussion_r1512108913

clalancette:
This is just so we consistently use C++-style anonymous namespaces, instead of C-style "static" functions.  No functional change.

This should fix #124

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_data_types.cpp
@@ -33,7 +33,9 @@
 #include "rmw_data_types.hpp"
 
 ///=============================================================================
-static size_t hash_gid(const uint8_t * gid)
+namespace
+{
+size_t hash_gid(const uint8_t * gid)
 {
   std::stringstream hash_str;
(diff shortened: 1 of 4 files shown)


Yadunund (approved):

Timple:
Just curious, what changed your mind with regards to the preference for `static`?

clalancette:
> Just curious, what changed your mind with regards to the preference for `static`?

We just had more anonymous functions, so it was less work to convert `static` -> anonymous rather than anonymous -> `static`.

Issue #212, opened 2024-06-25
Drop usage of `rmw_dds_common::qos_profile_get_best_available_X`


Yadunund:
At the same time explore transmitting QoS info only if there is a deviation from the system default. See https://github.com/ros2/rmw_zenoh/pull/210#discussion_r1652776472 for context.

clalancette:
We've now dropped our usage of `rmw_dds_common` completely, so I'm going to go ahead and close this out.

PR #234, opened 2024-07-02
Implement custom functions to adapt qos

Yadunund:
Partially addresses #212.

This PR gets `rcl/test_get_actual_qos` tests to pass. The main "fixs" here are to ensure any SYSTEM_DEFAULT values are assigned to those expected by the tests. Specifically
* Ensuring `deadline`, `lifespan`, and `liveliness_lease_durations` values are updated from their default `RMW_DURATION_UNSPECIFIED` to `RMW_DURATION_INFINITE`.
* `liveliness` is set to `RMW_QOS_POLICY_LIVELINESS_AUTOMATIC`.

I also updated the message callbacks to only pop from the queue if `history` is not `KEEP_ALL`.

Lastly, also moved all QoS related functionality to a new header. In a subsequent PR, I will update the liveliness token generation for the QoS portion to check against the `default_qos` before appending to the liveliness token.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/qos.cpp
+++ b/rmw_zenoh_cpp/src/detail/qos.cpp
@@ -1,0 +1,134 @@
+// Copyright 2024 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "qos.hpp"
+
+// Define defaults for various QoS settings.
+#define RMW_ZENOH_DEFAULT_HISTORY RMW_QOS_POLICY_HISTORY_KEEP_LAST;
+// If the depth field in the qos profile is set to 0, the RMW implementation
+// has the liberty to assign a default depth. The zenoh transport protocol
+// is configured with 256 channels so theoretically, this would be the maximum
+// depth we can set before blocking transport. A high depth would increase the
+// memory footprint of processes as more messages are stored in memory while a
+// very low depth might unintentionally drop messages leading to a poor
+// out-of-the-box experience for new users. For now we set the depth to 42,
+// a popular "magic number". See https://en.wikipedia.org/wiki/42_(number).
+#define RMW_ZENOH_DEFAULT_HISTORY_DEPTH 42;
+#define RMW_ZENOH_DEFAULT_RELIABILITY RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+#define RMW_ZENOH_DEFAULT_DURABILITY RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
+#define RMW_ZENOH_DEFAULT_DEADLINE RMW_DURATION_INFINITE;
+#define RMW_ZENOH_DEFAULT_LIFESPAN RMW_DURATION_INFINITE;
+#define RMW_ZENOH_DEFAULT_LIVELINESS RMW_QOS_POLICY_LIVELINESS_AUTOMATIC;
+#define RMW_ZENOH_DEFAULT_LIVELINESS_LEASE_DURATION RMW_DURATION_INFINITE;
+
+namespace rmw_zenoh_cpp
+{
+///=============================================================================
+QoS::QoS()
+{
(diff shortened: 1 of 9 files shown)


rmw_zenoh_cpp/src/detail/qos.hpp
```diff
@@ -0,0 +1,56 @@
...
+/// Signature matching rmw_get_publishers_info_by_topic and rmw_get_subscriptions_info_by_topic
+using GetEndpointInfoByTopicFunction = std::function<rmw_ret_t(
+      const rmw_node_t *,
+      rcutils_allocator_t *,
+      const char *,
+      bool,
+      rmw_topic_endpoint_info_array_t *)>;
+
+//==============================================================================
+class QoS
```
clalancette, line 36: This class feels awkward to me.

I wonder whether we need to have the class at all.  It seems to me that in `qos.cpp`, we could instead have:

```
static rmw_qos_profile_t default_qos =
{
  RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT,
  42,
  RMW_QOS_POLICY_RELIABILITY_RELIABLE,
  RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
  RMW_DURATION_INFINITE,
  RMW_DURATION_INFINITE,
  RMW_QOS_POLICY_LIVELINESS_AUTOMATIC,
  RMW_DURATION_INFINITE,
  false,
};
```

And then have `best_available_qos` be a free function, something like:

```
rmw_ret_t best_available_qos(rmw_qos_profile_t * qos_profile)
{
  // pull data from default_qos as necessary
}
```

That said, I don't know exactly what your plans are for upcoming PRs.

rmw_zenoh_cpp/src/detail/qos.cpp
```diff
@@ -0,0 +1,134 @@
...
+// has the liberty to assign a default depth. The zenoh transport protocol
+// is configured with 256 channels so theoretically, this would be the maximum
+// depth we can set before blocking transport. A high depth would increase the
+// memory footprint of processes as more messages are stored in memory while a
+// very low depth might unintentionally drop messages leading to a poor
+// out-of-the-box experience for new users. For now we set the depth to 42,
+// a popular "magic number". See https://en.wikipedia.org/wiki/42_(number).
+#define RMW_ZENOH_DEFAULT_HISTORY_DEPTH 42;
+#define RMW_ZENOH_DEFAULT_RELIABILITY RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+#define RMW_ZENOH_DEFAULT_DURABILITY RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
```
clalancette, line 29: This is a bit of a surprising default to me.  I would think that this would either be `RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT`, or, if we need this to be concrete, `RMW_QOS_POLICY_DURABILITY_VOLATILE`.  Can you explain the reasoning behind `TRANSIENT_LOCAL` here?

clalancette (commented):
Overall, this is an improvement, and fixes the test as you say.  I left two things inline to consider.

rmw_zenoh_cpp/src/detail/qos.hpp
Yadunund: I was planning to update [qos_to_keyexpr](https://github.com/ros2/rmw_zenoh/blob/9ede38d1d84e9b7f08971d56d1843624b5052240/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L174) to call `QoS()::get().default_qos()` and compare each field in the default with that of the input qos. Hence a class. I could also just define the `static rmw_qos_profile_t default_qos` as you mentioned here in the header file. Let me know if you have a preference.

rmw_zenoh_cpp/src/detail/qos.cpp
Yadunund: The default is set in [rmf_dds_common::qos_profile_get_best_available_for_publisher/subscription](https://github.com/ros2/rmw_dds_common/blob/0cf1ce476af39f8706004ea5432d00645aa93a66/rmw_dds_common/src/qos.cpp#L500-L507). This function is used in `rmw_zenoh` prior to this PR and also in other DDS based RMWs.

clalancette: Ah, got it.  Thanks for the explanation, in that case, nothing to do here.

rmw_zenoh_cpp/src/detail/qos.hpp
clalancette: I see.  In that case you would have to export `rmw_qos_profile_t default_qos` anyway, since to re-implement `qos_to_keyexpr` you need access to each individual field.

Overall, even though I'm not a huge fan of this class, it isn't all that important in the end.  Thus I'm going to approve this anyway so we can get the improvements here.  We can always consider re-implementing this in the future if we think of a better way to do this.

clalancette (approved):

PR #245, opened 2024-07-15
fix: log the function name, the file name, and the line number with macros

YuanYuYuan:
We should log the information of function names, file names, and line numbers with macros rather than doing so inside a function. 

Taking `export RCUTILS_CONSOLE_OUTPUT_FORMAT="[{severity} {time}] [{name}]: {message} ({function_name}() at {file_name}:{line_number})"` as an example,

- Current

```txt
[ERROR 1721033624.485865217] [Logging Test]: Message (log_named() at /workspace/src/rmw_zenoh/rmw_zenoh_cpp/src/detail/logging.cpp:50)
```

- After fixed
```txt
[ERROR 1721034005.478704138] [Logging Test]: Message (rmw_init() at /workspace/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_init.cpp:95)
```

### DIFF
--- a/rmw_zenoh_cpp/src/detail/logging.cpp
+++ b/rmw_zenoh_cpp/src/detail/logging.cpp
@@ -36,6 +36,9 @@ void Logger::set_log_level(RCUTILS_LOG_SEVERITY new_level)
 
 void Logger::log_named(
   RCUTILS_LOG_SEVERITY level,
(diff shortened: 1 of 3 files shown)


Yadunund (approved):
Thanks for the fix!

Issue #238, opened 2024-07-04
[multi-host] failed to initialize rosout publisher: unable to create zenoh publisher cache


berndpfrommer:
Scenario is as follows. Two machines, connected to the same wifi router. I'm running a ``zenohd`` on one host, and want to query the ``zenohd`` from the other host.

Both running the latest version of rmw_zenoh_cpp (Wed Jul 3rd, commit is 747a72ee774c65113b6f738fd8b87c586ae21e1b).

Host A [192.168.17.155]   < --- wifi -- >  Host B [ 192.168.17.249]  (runs zenoh_hd)

On Host A, running ``ros2 topic list`` fails:
```
ZENOH_SESSION_CONFIG_URI=${HOME}/.config/rmw_zenoh_session_config.json5 ros2 topic list
[INFO] [1720111205.046398983] [rmw_zenoh_cpp]: Successfully connected to a Zenoh router with id c3505398a7e2fc5f5d6f4d64204756.
[ERROR] [1720111206.056194866] [rmw_zenoh_cpp]: create publisher liveliness: 1
failed to initialize rosout publisher: unable to create zenoh publisher cache, at /opt/ros/src/rolling/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_zenoh.cpp:633, at ./src/rcl/publisher.c:117
```
The ``rmw_zenoh_session_config.json5`` file on Host A has this content:
```
{
  connect: {
    endpoints: [
      "tcp/192.168.17.249:7447",
    ],
  },
}
```
The debug messages on Host B indicate that the connection is made successfully, but then a read error occurs:
```
2024-07-04T16:46:50.515086Z DEBUG  rx-0 ThreadId(06) zenoh_transport::unicast::universal::link: RX task failed: Read error on TCP link [::ffff:192.168.17.249]:7447 => [::ffff:192.168.17.155]:60024: early eof at /home/pfrommer/.cargo/git/checkouts/zenoh-cc237f2570fab813/93f93d2/io/zenoh-links/zenoh-link-tcp/src/unicast.rs:130.
2024-07-04T16:46:50.515166Z DEBUG net-0 ThreadId(03) zenoh_transport::unicast::universal::transport: [cde320992ffe7790db9c0d9ea63f3029] Closing transport with peer: 4f3c8070ef6f52be764a440643a6b6f3
[...]

PR #246, opened 2024-07-15
Activate Zenoh logs by default

JEnoch:
Currently no logs from Zenoh are shown, neither "error" level, neither "warn" level.
The user would have to define the `RUST_LOG` environment variable to enable those logs.

This PR checks automatically defines the `RUST_LOG` environment variable before opening a Zenoh Session, if not already defined in the environment.
- For the Router it's set by default to `RUST_LOG=info` (similar level than the original Zenoh Router)
- For the Nodes it's set by default to `RUST_LOG=warn`

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_init.cpp
+++ b/rmw_zenoh_cpp/src/rmw_init.cpp
@@ -42,10 +42,6 @@ extern "C"
 // TODO(clalancette): Make this configurable, or get it from the configuration
 #define SHM_BUFFER_SIZE_MB 10
 
-// Environment variable for Zenoh logging
-const char * ZENOH_LOG_ENV_VAR = "RUST_LOG";
-const char * ZENOH_LOG_DEFAULT_LEVEL = "warn";
-
 namespace
 {
 void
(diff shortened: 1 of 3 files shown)


JEnoch:
This is a follow up of [this discussion](https://github.com/ros2/rmw_zenoh/issues/238#issuecomment-2211175116)

rmw_zenoh_cpp/src/rmw_init.cpp
```diff
@@ -42,6 +42,10 @@ extern "C"
 // TODO(clalancette): Make this configurable, or get it from the configuration
 #define SHM_BUFFER_SIZE_MB 10
 
+// Environment variable for Zenoh logging
+const char * ZENOH_LOG_ENV_VAR = "RUST_LOG";
```
Yadunund: My preference would be to move these definitions into `detail/zenoh_config.hpp` which is already included by `rmw_init.cpp` and `zenohd/main.cpp` below. Then you can define a `ZENOH_LOG_INFO_LEVEL_STR` and `ZENOH_LOG_DEBUG_LEVEL_STR` and use these in the `setenv` call.

Yadunund (changes requested):
Thanks for updating the implementation! Please find my feedback below.

rmw_zenoh_cpp/src/rmw_init.cpp
JEnoch: Done!

Yadunund (approved):
Pushed a small change to return an `rmw_ret_t` enum in `rmw_init` but otherwise LGTM!

PR #247, opened 2024-07-23
Pass `-r` to rosdep during Setup

fuzzypixelz:
On Ubuntu 22.04.4 LTS, with ROS2 Rolling installed via [Debian packages](https://docs.ros.org/en/rolling/Installation/Ubuntu-Install-Debians.html#ubuntu-debian-packages), I get the following error when following the README [Setup](https://docs.ros.org/en/rolling/Installation/Ubuntu-Install-Debians.html#ubuntu-debian-packages) instructions:

```console
$ rosdep install --ignore-src --from-paths src --rosdistro rolling -y
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
zenoh_c_vendor: No definition of [ament_cmake_vendor_package] for OS version [jammy]
rmw_zenoh_cpp: No definition of [rmw] for OS version [jammy]
```

While skimming some [ROS2-related Github issues](https://github.com/IntelRealSense/realsense-ros/issues/2309) it seems that passing `-r` which means "Continue installing despite errors." resolves the issue (at least for me).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ See [zenoh_c_vendor/CMakeLists.txt](./zenoh_c_vendor/CMakeLists.txt) for more de
 mkdir ~/ws_rmw_zenoh/src -p && cd ~/ws_rmw_zenoh/src
 git clone https://github.com/ros2/rmw_zenoh.git
 cd ~/ws_rmw_zenoh
-rosdep install --from-paths src --ignore-src --rosdistro <DISTRO> -y # replace <DISTRO> with ROS 2 distro of choice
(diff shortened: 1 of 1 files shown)


clalancette:
So the problem with that setup is that we don't support ROS 2 Rolling on Ubuntu 22.04 anymore; it only supports Ubuntu 24.04.  That's why rosdep was failing for you.

I'm going to suggest that you either stick with ROS 2 Iron (the last supported release on Ubuntu 22.04), or use ROS 2 Rolling on Ubuntu 24.04. which should fix the error for you.

In either case, I don't think we need a change here, so I'm going to close this out.

fuzzypixelz:
@clalancette Can we mention this in the README then?

PR #248, opened 2024-07-23
construct and destruct rmw_node_data_t

Yadunund:
Construct and destruct node->data after allocating memory. When we add non-POD data fields to `rmw_node_data_t`, we will run into issues when accessing fields.


clalancette (approved):
This looks great!  Thanks for putting this together.

PR #250, opened 2024-07-25
Use an ordered_map to store the TopicMap.

clalancette:
Some of the tests in ROS 2 assume that the order in which entities (pub, sub, clients, services) are added to the system is the same order in which they will be returned from calls like get_topic_names_and_types().  To facilitate that, use an ordered_map for the mapping between a topic name and the topic type, which remembers insertion order.

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -19,6 +19,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: uncrustify
-      run: /ros_entrypoint.sh ament_uncrustify rmw_zenoh_cpp/
+      run: /ros_entrypoint.sh ament_uncrustify --exclude rmw_zenoh_cpp/src/detail/ordered_hash.hpp rmw_zenoh_cpp/src/detail/ordered_map.hpp rmw_zenoh_cpp/
     - name: cpplint
-      run: /ros_entrypoint.sh ament_cpplint rmw_zenoh_cpp/
+      run: /ros_entrypoint.sh ament_cpplint --exclude rmw_zenoh_cpp/src/detail/ordered_hash.hpp rmw_zenoh_cpp/src/detail/ordered_map.hpp rmw_zenoh_cpp/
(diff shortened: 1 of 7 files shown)


.github/workflows/style.yaml
```diff
@@ -19,6 +19,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: uncrustify
-      run: /ros_entrypoint.sh ament_uncrustify rmw_zenoh_cpp/
+      run: /ros_entrypoint.sh ament_uncrustify --exclude rmw_zenoh_cpp/src/detail/ordered_hash.hpp rmw_zenoh_cpp/src/detail/ordered_map.hpp rmw_zenoh_cpp/
```
Yadunund, line 22: I guess we could delete this workflow and update build.yaml to additionally run `colcon test`? Or do we want to keep this since it will catch style issues earlier?

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -80,7 +81,7 @@ struct GraphNode
   // Map topic type to QoSMap
   using TopicTypeMap = std::unordered_map<std::string, TopicQoSMap>;
   // Map topic name to TopicTypeMap
-  using TopicMap = std::unordered_map<std::string, TopicTypeMap>;
+  using TopicMap = tsl::ordered_map<std::string, TopicTypeMap>;
```
Yadunund, line 86: Could we add a note above on why we're using `tsl::ordered_map` here? ie highlight that tests expect topics to be returned in same order as insertion

Yadunund (approved):
Looks great to me!

.github/workflows/style.yaml
clalancette: I'm up for either.  It is kind of nice to get "early" notification on simple linter failures, but it also requires us to do extra things like this.  I'm going to leave this as-is for now, and we can decide later on if we want to just rely on `colcon test` in the future.

rmw_zenoh_cpp/src/detail/graph_cache.hpp
clalancette: Good call.  Done in 6518e5f.

Yadunund (approved):

Issue #251, opened 2024-07-29
Console Doesn't Revert To Old Console Mode After Router Process Is Terminated


CihatAltiparmak:
Hello,

In router executable, SIGINT signal is handled but SIGTERM is not handled. This gives rise not to close router gracefully. If we directly send SIGTERM to router, router doesn't close session and KeyboardReader instance doesn't destroy. Additionaly, When it is threw out reading error from tty, we should close session before finishing process. I have a PR for it.

DarkusAlphaHydranoid:
PR is merged. Can be closed.

PR #252, opened 2024-07-29
Added SIGTERM handler for linux and closed session when reading error is caught

CihatAltiparmak:
Addresses #251

I have no chance to try it in windows machine. but seems windows handles all signals in one handler function.

### DIFF
--- a/rmw_zenoh_cpp/src/zenohd/main.cpp
+++ b/rmw_zenoh_cpp/src/zenohd/main.cpp
@@ -199,7 +199,7 @@ int main(int argc, char ** argv)
     printf("Unable to open router session!\n");
     return 1;
   }
-  auto close_session = rcpputils::make_scope_exit(
+  auto always_close_session = rcpputils::make_scope_exit(
     [&session]() {
       z_close(z_move(session));
     });
--- a/rmw_zenoh_cpp/src/zenohd/main.cpp
+++ b/rmw_zenoh_cpp/src/zenohd/main.cpp
@@ -36,6 +36,8 @@
 
 #include "rmw/error_handling.h"
 
+#include "rcpputils/scope_exit.hpp"
+
 static bool running = true;
 
 class KeyboardReader final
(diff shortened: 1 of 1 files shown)


rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -217,6 +218,8 @@ int main(int argc, char ** argv)
       c = keyreader.readOne();
     } catch (const std::runtime_error &) {
       perror("read():");
+
+      z_close(z_move(s));
```
Yadunund: to prevent duplicated `z_close(z_move(s));` calls (here and below in line 233), I wonder if it's better to wrap this statement within a `rcpputils::scope_exit` block like https://github.com/ros2/rmw_zenoh/blob/fcfed30ca7ba629a137bc48fab0502c71e515839/rmw_zenoh_cpp/src/rmw_init.cpp#L203-L206

rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -217,6 +218,8 @@ int main(int argc, char ** argv)
       c = keyreader.readOne();
     } catch (const std::runtime_error &) {
       perror("read():");
+
```
Yadunund:
```suggestion
```

Yadunund (changes requested):
Thanks for the enhancement! Left a minor comment to consider

rmw_zenoh_cpp/src/zenohd/main.cpp
CihatAltiparmak: Done it in this commit ( https://github.com/ros2/rmw_zenoh/pull/252/commits/0566dc2a4736931e8c61c3bd370cee78ee07716b )

rmw_zenoh_cpp/src/zenohd/main.cpp
CihatAltiparmak: Added scope_exit functionality in this commit ( https://github.com/ros2/rmw_zenoh/pull/252/commits/0566dc2a4736931e8c61c3bd370cee78ee07716b )

CihatAltiparmak:
I'm ready for second review @Yadunund , but i want to mention just one point. If i don't remove `perror` in router executable, terminal looks like below when i send SIGTERM to `ros2` (assuming that executes `ros2 run rmw_zenoh_cpp rmw_zenohd`) 
. But the relavant processes are closed. This behavior is not related to my branch.
Screenshot from 2024-07-31 11-33-24

rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -192,20 +194,26 @@ int main(int argc, char ** argv)
...
-    rmw_zenoh_cpp::liveliness::zid_to_str(z_info_zid(z_session_loan(&s))).c_str());
+    rmw_zenoh_cpp::liveliness::zid_to_str(z_info_zid(z_session_loan(&session))).c_str());
 #ifdef _WIN32
   SetConsoleCtrlHandler(quit, TRUE);
 #else
   signal(SIGINT, quit);
+  signal(SIGTERM, quit);
 #endif
 
+  auto close_session = rcpputils::make_scope_exit(
```
Yadunund: Sorry one tiny nitpick, can we move this `close_session` instantiation right after the `if (!z_check(session))` block? Just to be consistent with other parts of the codebase.

CihatAltiparmak: Done it with this commit (https://github.com/ros2/rmw_zenoh/pull/252/commits/35c5ba0fb81878796ab75bfa705f241daf87fd1d)

Yadunund (approved):
Thanks for iterating so quickly!

PR #255, opened 2024-08-02
remove rmw_localhost_only_t.

CihatAltiparmak:
Hello, maybe it's also in your plan, but i wanted to open PR about deprecated usage of rmw_localhost_only_t just in case you have overlooked this PR ( https://github.com/ros2/rcl/pull/1169 )  in rcl. I have discovered this one while running ros-industrial ci in my repo and some errors happen like below in the `ROS_REPO=testing` case.

```shell
       Finished release [optimized] target(s) in 7m 07s
  ---
  Finished <<< zenoh_c_vendor [7min 14s]
  Starting >>> rmw_zenoh_cpp
  --- stderr: rmw_zenoh_cpp
  /root/upstream_ws/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_init_options.cpp: In function ‘rmw_ret_t rmw_init_options_init(rmw_init_options_t*, rcutils_allocator_t)’:
  /root/upstream_ws/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_init_options.cpp:50:17: error: ‘rmw_init_options_t’ {aka ‘struct rmw_init_options_s’} has no member named ‘localhost_only’
     50 |   init_options->localhost_only = RMW_LOCALHOST_ONLY_DEFAULT;
        |                 ^~~~~~~~~~~~~~
  /root/upstream_ws/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_init_options.cpp:50:34: error: ‘RMW_LOCALHOST_ONLY_DEFAULT’ was not declared in this scope
     50 |   init_options->localhost_only = RMW_LOCALHOST_ONLY_DEFAULT;
        |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
  /root/upstream_ws/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_init_options.cpp: In function ‘rmw_ret_t rmw_init_options_copy(const rmw_init_options_t*, rmw_init_options_t*)’:
  /root/upstream_ws/src/rmw_zenoh/rmw_zenoh_cpp/src/rmw_init_options.cpp:92:7: error: ‘rmw_init_options_t’ {aka ‘struct rmw_init_options_s’} has no member named ‘localhost_only’
     92 |   tmp.localhost_only = src->localhost_only;
        |       ^~~~~~~~~~~~~~
[...]

PR #256, opened 2024-08-03
Modify CI to track ros testing repo

CihatAltiparmak:
To prevent the errors stated in this PR ( https://github.com/ros2/rmw_zenoh/pull/255 ) , i have refactored the build CI so that it can be tracked the modifications, API changes etc. for rolling, jazzy and iron continuously. I don't know what type of approach this team has, however i wanted to come up with some option. It's ready to your service. It doesn't matter if you accepted or not. Btw to just modify container except this PR's idea can be another option. I can do it too.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,21 +4,52 @@ on:
   push:
     branches: [ rolling ]
   workflow_dispatch:
+  schedule:
+    # Run every morning to detect flakiness and broken dependencies
+    - cron: '03 5 * * *'
 defaults:
   run:
     shell: bash
 jobs:
-  test:
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        ROS_DISTRO: ['iron', 'jazzy', 'rolling']
-        ROS_REPO: ['main', 'testing']
-    timeout-minutes: 30
-    steps:
-      - uses: actions/checkout@v4
-      - uses: 'ros-industrial/industrial_ci@master'
-        env:
-          ROS_DISTRO: ${{ matrix.ROS_DISTRO }}
-          ROS_REPO: ${{ matrix.ROS_REPO }}
+  build_and_test_source_rolling:
+      runs-on: ubuntu-latest
+      container:
+        image: rostooling/setup-ros-docker:ubuntu-noble-latest
+      steps:
+      - name: Build and run tests
+        id: action-ros-ci
+        uses: ros-tooling/action-ros-ci@v0.3
+        with:
+          package-name: |
+            rmw_zenoh_cpp
+            zenoh_c_vendor
+          target-ros2-distro: rolling
+          vcs-repo-file-url: https://raw.githubusercontent.com/ros2/ros2/rolling/ros2.repos
+
+    build_and_test_binaries_jazzy:
+      runs-on: ubuntu-latest
+      container:
+        image: rostooling/setup-ros-docker:ubuntu-noble-latest
+      steps:
+      - name: Build and run tests
+        id: action-ros-ci
+        uses: ros-tooling/action-ros-ci@v0.3
+        with:
+          package-name: |
+            rmw_zenoh_cpp
+            zenoh_c_vendor
+          target-ros2-distro: jazzy
+
+    build_and_test_binaries_iron:
+      runs-on: ubuntu-latest
+      container:
+        image: rostooling/setup-ros-docker:ubuntu-noble-latest
+      steps:
+      - name: Build and run tests
+        id: action-ros-ci
+        uses: ros-tooling/action-ros-ci@v0.3
+        with:
+          package-name: |
+            rmw_zenoh_cpp
+            zenoh_c_vendor
+          target-ros2-distro: iron
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,7 +42,7 @@ jobs:
   build_and_test_binaries_iron:
     runs-on: ubuntu-latest
     container:
-      image: rostooling/setup-ros-docker:ubuntu-noble-latest
+      image: rostooling/setup-ros-docker:ubuntu-jammy-latest
     steps:
     - name: Build and run tests
       id: action-ros-ci
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,7 +11,11 @@ defaults:
   run:
     shell: bash
 jobs:
-  build_and_test_source_rolling:
+  build_and_test:
+      name: build_and_test_binaries_${{ matrix.ROS_DISTRO }}
+      strategy:
+        matrix:
+          ROS_DISTRO: [rolling, jazzy, iron]
       runs-on: ubuntu-latest
       container:
         image: rostooling/setup-ros-docker:ubuntu-noble-latest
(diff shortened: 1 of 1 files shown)


MichaelOrlov:
@Yadunund @clalancette Could you please review this PR?

Yadunund (changes requested):
Thanks for kicking off this initiative! 

I feel that the using binaries from the `testing` repo may not highlight issues soon enough. It relies on someone blooming changes from the core repos and then waiting for  the rosdistro PR to be merged to kick-off the build jobs.

On the other hand we expect core packages for released distros like jazzy, iron etc to be fairly stable so we don't need to test with the latest source code or binaries from the test repo.

What do you think of the following:
- Update the `build.yaml` workflow to rely on [setup-ros](https://github.com/ros-tooling/setup-ros).
  - If we're testing for `rolling`, the job will build `--packages-up-to rmw_zenoh_cpp` from **source**. 
  - Else, we configure `setup-ros` to use regular ROS 2 core binaries and build upto `rmw_zenoh_cpp`
- Configure the build job to run on a nightly schedule using a [cron config](https://github.com/open-rmf/rmf/blob/100453f07c74b48babc7333a25c6db342dd187b1/.github/workflows/build.yaml#L5-L6). This way the maintainers will be notified if any breaking changes get merged in core upstream repos.

clalancette:
Agreed with what @Yadunund laid out above.

ahcorde:
I made the changes @Yadunund 

 - source build for rolling `build / build_and_test_source_rolling `
 - binary build for jazzy `build / build_and_test_binaries_jazzy `
 - binary build for iron `build / build_and_test_binaries_iron`
 - Added cron too

ahcorde:
[rolling is failing](https://github.com/ros2/rmw_zenoh/actions/runs/10506648470/job/29106714188?pr=256) because of the `localhost_only` issue https://github.com/ros2/rmw_zenoh/pull/255

.github/workflows/build.yaml
```diff
@@ -4,27 +4,51 @@ on:
...
   workflow_dispatch:
+  schedule:
+    # Run every morning to detect flakiness and broken dependencies
+    - cron: '03 5 * * *'
 defaults:
   run:
     shell: bash
 jobs:
-  test:
+  build_and_test_source_rolling:
```
Yadunund: Instead of having three separate jobs like this, can we retain the `distro` matrix from earlier and setup one job only to run for each distro in the matrix? Within this job, we can have a special `if` condition to check if `distro == rolling` and if so, build from source. If you need access to both the ROS distro version and ubuntu version, the matrix can be defined [like so](https://github.com/open-rmf/rmf_ci_templates/blob/1ae7b4f64644ab2f8724e9a1e1b3d14803caa61e/.github/workflows/reusable_build.yaml#L13-L25).

Also it appears that your changes now pull a `rostooling/setup-ros-docker:ubuntu-x` image. How often is this image updated? And why not use `ros-tooling/setup-ros@v0.7` to setup from latest `ubuntu` base image [like so](https://github.com/ros-tooling/setup-ros?tab=readme-ov-file#setting-up-the-worker-installing-system-dependencies-and-ros-linux).

Yadunund (changes requested):
Thanks @ahcorde for the changes. Left a comment to reuse the same job across distros/os.

.github/workflows/build.yaml
Yadunund: Here is an example of using an `if` condition within the action: https://github.com/open-rmf/rmf_ci_templates/blob/1ae7b4f64644ab2f8724e9a1e1b3d14803caa61e/.github/workflows/reusable_build.yaml#L53

.github/workflows/build.yaml
```diff
@@ -4,27 +4,28 @@ on:
...
-    - name: build
-      run: /ros_entrypoint.sh colcon build
+  build_and_test:
+      name: build_and_test_binaries_${{ matrix.ROS_DISTRO }}
+      strategy:
+        matrix:
+          ROS_DISTRO: [rolling, jazzy, iron]
+      runs-on: ubuntu-latest
+      container:
+        image: rostooling/setup-ros-docker:ubuntu-noble-latest
```
Yadunund: We can't alwayse use `noble` since iron is distributed on `jammy`. Please see my previous comment on how the matrix can be updated to include distro and os. https://github.com/ros2/rmw_zenoh/pull/256#discussion_r1727744706

Not addressed form the same comment:
* Using If condition to include `vcs-repo-file-url` 
* Using ros-tooling/setup-ros@v0.7 instead of the `rostooling/setup-ros-docker:ubuntu` image

Yadunund (changes requested):

.github/workflows/build.yaml
CihatAltiparmak: Firstly sorry for some commits, but would using separate jobs for source and binaries be good instead of using if?  I am yet to complete my work. I will ping you.

Yadunund: I'd prefer to have everything in the same job. The CI here is a temporary measure until we start running the usual CI jobs using the ROS buildfarm.

CihatAltiparmak: I've applied, @Yadunund and i've readded the fail-fast param which @ahcorde has overlooked by deleting this. Would you review in your available time? (Inspired by https://github.com/ros-tooling/setup-ros/blob/a6ce30ecca1e5dcc10ae5e6a44fe2169115bf852/README.md?plain=1#L249-L252 and https://github.com/ros2/ros2_tracing/blob/a866b9c701311bc8200a00c949b4a0fff9803777/.github/workflows/test.yml#L68) Btw if you want, we can merge `rolling` branch to this branch so that all CI can pass.

ahcorde: thank you for the changes @CihatAltiparmak

ahcorde (approved):
I will wait for a final review from @Yadunund

.github/workflows/build.yaml
```diff
@@ -4,27 +4,46 @@ on:
...
-        distro: ['iron', 'jazzy', 'rolling']
+        include:
+          # Rolling (source)
+          - ROS_DISTRO: rolling
+            BUILD_TYPE: source
+            DOCKER_IMAGE: ubuntu:noble
+          # Jazzy (binary)
+          - ROS_DISTRO: jazzy
+            BUILD_TYPE: binary
+            DOCKER_IMAGE: ubuntu:noble
```
Yadunund: I think we should remove this `DOCKER_IMAGE` flag as it's one extra item to configure.
See commend below for relevant suggestion.

.github/workflows/build.yaml
```diff
@@ -4,27 +4,46 @@ on:
...
+          - ROS_DISTRO: iron
+            BUILD_TYPE: binary
+            DOCKER_IMAGE: ubuntu:jammy
+    env:
+      ROS2_REPOS_FILE_URL: 'https://raw.githubusercontent.com/ros2/ros2/${{ matrix.ROS_DISTRO }}/ros2.repos'
+    runs-on: ubuntu-latest
     container:
-      image: ros:${{ matrix.distro }}-ros-base
-    timeout-minutes: 30
+      image: ${{ matrix.DOCKER_IMAGE }}
```
Yadunund: Our CI times have increased  by 2mins by installed ROS binaries for cases where `BUILD_TYPE` is `binary`. Can we use an if condition here to use `ros:${{ matrix. ROS_DISTRO }}-ros-base` here if `matrix.BUILD_TYPE == 'binary'`; else `ubuntu:noble`. We won't require the DOCKER_IMAGE variable anymore.

Yadunund (changes requested):
It looks like CI time for binary jobs has increased by little over 2 mins due to the fact that we're installing ROS binaries as part of the workflow. Left a suggestion for how we could tackle this.

.github/workflows/build.yaml
CihatAltiparmak: I've modified based on your suggestion. But i've also had to install coverage tools in case of using binary because of crashing CI. I hope i've decreased CI check time. Finally, my approach is okay with binary builds, but It can create some problems in case that you want to build from iron source because iron fails in `ubuntu-noble` 
While modifying, i've utilized https://github.com/open-rmf/rmf_ci_templates/blob/1ae7b4f64644ab2f8724e9a1e1b3d14803caa61e/.github/workflows/reusable_build.yaml#L56-L59

Yadunund (approved):
LGTM! Thanks for iterating @CihatAltiparmak and for addressing the build error on Rolling!

Issue #249, opened 2024-07-25
Race conditions when accessing data


Yadunund:
The current implementation suffers from data races between threads that access the same data, eg. [rmw_publisher_data_t](https://github.com/ros2/rmw_zenoh/blob/fcfed30ca7ba629a137bc48fab0502c71e515839/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp#L94). There are essentially two classes of races:
1. One thread destructing the data object and another accessing is modifying it.
2. Two or more threads are modifying the data object.

### Proposal to address the issue
- [ ] Update all data classes to rely on getters and setters to access members while locking mutexes.
- [ ] The type erased data fields in rmw types, eg [rmw_publisher_t->data](https://github.com/ros2/rmw/blob/5ce844c23131113f14102ad45aac009142bafa9a/rmw/include/rmw/types.h#L128) will store the raw pointer to the RMW context. The assumption here is that the context will always be valid if an RMW entity is valid
- [ ] The [rmw_context_impl_s](https://github.com/ros2/rmw_zenoh/blob/fcfed30ca7ba629a137bc48fab0502c71e515839/rmw_zenoh_cpp/src/detail/rmw_data_types.hpp#L46) class will store sets/vectors of `rmw_x_data_t` objects of the various RMW entities (publishers, subs, clients, servers, etc). Two or more threads will lock the same mutex in the `rmw_context_impl_s` while accessing/modifying these objects.

clalancette:
I believe that this has been done with a combination of refactors from @Yadunund , plus the switch to zenoh-cpp.  Thus I'm going to close this out.

PR #258, opened 2024-08-06
Thread-safe access to graph cache

Yadunund:
Step 1 in addressing #249.

This PR updates the `rmw_context_impl_s` class to accurately manage the lifetime of its members while ensure thread-safe data access. For starters, it ensures that the Graph Cache updates and lookups are thread safe.

In subsequent PRs, I will update the `rmw_context_impl_s` class to store `rmw_publisher_data_t`, etc with member functions to manage/access their functionalities.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -54,6 +54,7 @@ void rmw_context_impl_s::graph_sub_data_handler(const z_sample_t * sample, void 
       break;
     case z_sample_kind_t::Z_SAMPLE_KIND_DELETE:
       data_ptr->graph_cache_->parse_del(keystr._cstr);
+      break;
     default:
       return;
   }
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -107,6 +107,16 @@ rmw_ret_t rmw_context_impl_s::Data::subscribe()
   }
   // Setup the liveliness subscriber to receives updates from the ROS graph
   // and update the graph cache.
+  // TODO(Yadunund): This closure is still not 100% thread safe as we are
+  // passing Data* as the type erased argument to z_closure. Thus during
+  // the execution of graph_sub_data_handler, the rawptr may be freed/reset
+  // by a different thread. When we switch to zenoh-cpp we can replace z_closure
+  // with a lambda that captures a weak_ptr<Data> by copy. The lambda and caputed
+  // weak_ptr<Data> will have the same lifetime as the subscriber. Then within
+  // graph_sub_data_handler, we would first lock to weak_ptr to check if the
+  // shared_ptr<Data> exits. If it does, then even if a different thread calls
+  // rmw_context_fini() to destroy rmw_context_impl_s, the locked
+  // shared_ptr<Data> would live on until the graph_sub_data_handler callback.
   auto sub_options = zc_liveliness_subscriber_options_null();
   z_owned_closure_sample_t callback = z_closure(
     rmw_context_impl_s::graph_sub_data_handler, nullptr,
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -70,7 +70,7 @@ void rmw_context_impl_s::graph_sub_data_handler(const z_sample_t * sample, void 
   }
 
   // Trigger the ROS graph guard condition.
-  rmw_ret_t rmw_ret = rmw_trigger_guard_condition(data_ptr->graph_guard_condition_);
+  rmw_ret_t rmw_ret = rmw_trigger_guard_condition(data_ptr->graph_guard_condition_.get());
   if (RMW_RET_OK != rmw_ret) {
     RMW_ZENOH_LOG_WARN_NAMED(
       "rmw_zenoh_cpp",
@@ -81,25 +81,23 @@ void rmw_context_impl_s::graph_sub_data_handler(const z_sample_t * sample, void 
 
 ///=============================================================================
 rmw_context_impl_s::Data::Data(
-  const rcutils_allocator_t * allocator,
   const std::string & enclave,
   z_owned_session_t session,
   std::optional<zc_owned_shm_manager_t> shm_manager,
   const std::string & liveliness_str,
-  std::shared_ptr<rmw_zenoh_cpp::GraphCache> graph_cache,
-  rmw_guard_condition_t * graph_guard_condition)
-: allocator_(allocator),
-  enclave_(std::move(enclave)),
+  std::shared_ptr<rmw_zenoh_cpp::GraphCache> graph_cache)
+: enclave_(std::move(enclave)),
   session_(std::move(session)),
   shm_manager_(std::move(shm_manager)),
   liveliness_str_(std::move(liveliness_str)),
   graph_cache_(std::move(graph_cache)),
-  graph_guard_condition_(graph_guard_condition),
   is_shutdown_(false),
   next_entity_id_(0),
   is_initialized_(false)
 {
-  // Do nothing.
+  graph_guard_condition_ = std::make_unique<rmw_guard_condition_t>();
+  graph_guard_condition_->implementation_identifier = rmw_zenoh_cpp::rmw_zenoh_identifier;
+  graph_guard_condition_->data = &guard_condition_data_;
 }
 
 ///=============================================================================
@@ -168,31 +166,15 @@ rmw_ret_t rmw_context_impl_s::Data::shutdown()
 ///=============================================================================
 rmw_context_impl_s::Data::~Data()
 {
-  RMW_TRY_DESTRUCTOR(
-    static_cast<rmw_zenoh_cpp::GuardCondition *>(
-      graph_guard_condition_->data)->~GuardCondition(),
-    rmw_zenoh_cpp::GuardCondition, );
-  if (rcutils_allocator_is_valid(allocator_)) {
-    allocator_->deallocate(graph_guard_condition_->data, allocator_->state);
-    allocator_->deallocate(graph_guard_condition_, allocator_->state);
-    graph_guard_condition_ = nullptr;
-  }
-
   auto ret = this->shutdown();
   static_cast<void>(ret);
 }
 
 ///=============================================================================
 rmw_context_impl_s::rmw_context_impl_s(
-  const rcutils_allocator_t * allocator,
   const std::size_t domain_id,
   const std::string & enclave)
 {
-  // Check if allocator is valid.
-  if (!rcutils_allocator_is_valid(allocator)) {
-    throw std::runtime_error("Invalid allocator passed to rmw_context_impl_s.");
-  }
-
   // Initialize the zenoh configuration.
   z_owned_config_t config;
   rmw_ret_t ret;
@@ -322,56 +304,15 @@ rmw_context_impl_s::rmw_context_impl_s(
       }
     });
 
-  // Initialize the guard condition.
-  rmw_guard_condition_t * graph_guard_condition =
-    static_cast<rmw_guard_condition_t *>(allocator->zero_allocate(
-      1, sizeof(rmw_guard_condition_t), allocator->state));
-  if (graph_guard_condition == NULL) {
-    throw std::runtime_error("failed to allocate graph guard condition");
-  }
-  auto free_guard_condition = rcpputils::make_scope_exit(
-    [graph_guard_condition, allocator]() {
-      allocator->deallocate(graph_guard_condition, allocator->state);
-    });
-  graph_guard_condition->implementation_identifier =
-    rmw_zenoh_cpp::rmw_zenoh_identifier;
-  graph_guard_condition->data =
-    allocator->zero_allocate(1, sizeof(rmw_zenoh_cpp::GuardCondition), allocator->state);
-  if (graph_guard_condition->data == NULL) {
-    throw std::runtime_error("failed to allocate graph guard condition data");
-  }
-  auto free_guard_condition_data = rcpputils::make_scope_exit(
-    [graph_guard_condition, allocator]() {
-      allocator->deallocate(graph_guard_condition->data, allocator->state);
-    });
-  RMW_TRY_PLACEMENT_NEW(
-    graph_guard_condition->data,
-    graph_guard_condition->data,
-    throw std::runtime_error("failed to initialize graph_guard_condition->data."),
-    rmw_zenoh_cpp::GuardCondition);
-  auto destruct_guard_condition_data = rcpputils::make_scope_exit(
-    [graph_guard_condition]() {
-      auto gc_data =
-      static_cast<rmw_zenoh_cpp::GuardCondition *>(graph_guard_condition->data);
-      RMW_TRY_DESTRUCTOR_FROM_WITHIN_FAILURE(
-        gc_data->~GuardCondition(),
-        rmw_zenoh_cpp::GuardCondition);
-    });
-
   close_session.cancel();
   free_shm_manager.cancel();
-  destruct_guard_condition_data.cancel();
-  free_guard_condition_data.cancel();
-  free_guard_condition.cancel();
 
   data_ = std::make_shared<Data>(
-    allocator,
     std::move(enclave),
     std::move(session),
     std::move(shm_manager),
     std::move(liveliness_str),
-    std::move(graph_cache),
-    graph_guard_condition);
+    std::move(graph_cache));
 
   ret = data_->subscribe_to_ros_graph();
   if (ret != RMW_RET_OK) {
@@ -404,7 +345,7 @@ std::optional<zc_owned_shm_manager_t> & rmw_context_impl_s::shm_manager()
 rmw_guard_condition_t * rmw_context_impl_s::graph_guard_condition()
 {
   std::lock_guard<std::mutex> lock(data_->mutex_);
-  return data_->graph_guard_condition_;
+  return data_->graph_guard_condition_.get();
 }
 
 ///=============================================================================
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -15,6 +15,8 @@
 #include "rmw_context_impl_s.hpp"
 
 #include <mutex>
+#include <optional>
+#include <string>
 #include <thread>
 #include <utility>
 
@@ -101,8 +103,9 @@ rmw_context_impl_s::Data::Data(
 }
 
 ///=============================================================================
-rmw_ret_t rmw_context_impl_s::Data::subscribe()
+rmw_ret_t rmw_context_impl_s::Data::subscribe_to_ros_graph()
 {
+  std::lock_guard<std::mutex> lock(mutex_);
   if (is_initialized_) {
     return RMW_RET_OK;
   }
@@ -120,8 +123,7 @@ rmw_ret_t rmw_context_impl_s::Data::subscribe()
   // shared_ptr<Data> would live on until the graph_sub_data_handler callback.
   auto sub_options = zc_liveliness_subscriber_options_null();
   z_owned_closure_sample_t callback = z_closure(
-    rmw_context_impl_s::graph_sub_data_handler, nullptr,
-    this->shared_from_this().get());
+    rmw_context_impl_s::graph_sub_data_handler, nullptr, this);
   graph_subscriber_ = zc_liveliness_declare_subscriber(
     z_loan(session_),
     z_keyexpr(liveliness_str_.c_str()),
@@ -186,6 +188,11 @@ rmw_context_impl_s::rmw_context_impl_s(
   const std::size_t domain_id,
   const std::string & enclave)
 {
+  // Check if allocator is valid.
+  if (!rcutils_allocator_is_valid(allocator)) {
+    throw std::runtime_error("Invalid allocator passed to rmw_context_impl_s.");
+  }
+
   // Initialize the zenoh configuration.
   z_owned_config_t config;
   rmw_ret_t ret;
@@ -366,7 +373,7 @@ rmw_context_impl_s::rmw_context_impl_s(
     std::move(graph_cache),
     graph_guard_condition);
 
-  ret = data_->subscribe();
+  ret = data_->subscribe_to_ros_graph();
   if (ret != RMW_RET_OK) {
     throw std::runtime_error("Unable to subscribe to ROS Graph updates.");
   }
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -15,15 +15,22 @@
 #include "rmw_context_impl_s.hpp"
 
 #include <utility>
+#include <thread>
 
 #include "guard_condition.hpp"
+#include "identifier.hpp"
 #include "liveliness_utils.hpp"
 #include "logging_macros.hpp"
 #include "zenoh_config.hpp"
+#include "zenoh_router_check.hpp"
 
 #include "rcpputils/scope_exit.hpp"
 #include "rmw/error_handling.h"
 #include "rmw/impl/cpp/macros.hpp"
+
+// Megabytes of SHM to reserve.
+// TODO(clalancette): Make this configurable, or get it from the configuration
+#define SHM_BUFFER_SIZE_MB 10
 
 ///=============================================================================
 void rmw_context_impl_s::graph_sub_data_handler(const z_sample_t * sample, void * data)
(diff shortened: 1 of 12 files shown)


rmw_zenoh_cpp/src/rmw_init.cpp
```diff
@@ -242,148 +165,94 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
...
-        z_drop(z_move(context->impl->shm_manager.value()));
+    [&shm_manager]() {
+      if (shm_manager.has_value()) {
+        z_drop(z_move(shm_manager.value()));
       }
     });
 
   // Initialize the guard condition.
-  context->impl->graph_guard_condition =
+  rmw_guard_condition_t * graph_guard_condition =
```
clalancette: In the new structure, it seems like the `rmw_context_impl_s` structure is the "owner" of the `graph_guard_condition`.  If I'm correct about that, then my suggestion here is to move the "construction" of the graph guard_condition (this code, and the next dozen lines or so) into the constructor of `rmw_context_impl_s`.  I think it makes more sense there.  What do you think?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -210,11 +211,7 @@ rmw_create_node(
     context->impl,
     "expected initialized context",
     return nullptr);
-  RMW_CHECK_FOR_NULL_WITH_MSG(
-    context->impl->enclave,
-    "expected initialized enclave",
-    return nullptr);
```
clalancette, line 172: I don't have a good idea of whether or not this check should be in place.  What was your reasoning for removing it?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,452 @@
...
+{
+  std::lock_guard<std::mutex> lock(data_->mutex_);
+  return data_->enclave_;
+}
+
+///=============================================================================
+z_session_t rmw_context_impl_s::session() const
+{
+  std::lock_guard<std::mutex> lock(data_->mutex_);
+  return z_loan(data_->session_);
```
clalancette, line 334: This lock isn't doing too much.  The problem is that as soon as the call to `rmw_context_impl_s::session()` returns, the lock is dropped.  At that point, the caller has a loan to the session, but if some other thread changes the session, the caller now has a dangling loan.  We'll need to do something more clever with sessions here in order to make it thread safe.

This same comment applies to the `shm_manager()` and `graph_guard_condition()` methods below (it doesn't apply to `enclave()` above because that is returning a copy).

clalancette (commented):
I've left a few things to think about.

rmw_zenoh_cpp/src/rmw_init.cpp
Yadunund: Yeah I wanted to do that  but was unsure how to handle the case when we need to return `RMW_RET_BAD_ALLOC` when memory allocation fails. Should the constructor throw an error? In #259 I was able to rely on the `make()` pattern but I don't see a way to do that here unless we switch to a singleton of `rmw_context_impl_s`.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Ah I now store the `enclave` as an `std::string` within `rmw_context_impl_s` so we don't need to check for nullptr. Maybe I can update to check if the string is empty?

rmw_zenoh_cpp/src/rmw_init.cpp
clalancette: I think the constructor can throw; we are careful to use `RMW_TRY_PLACEMENT_NEW` when constructing, which captures the exception and turns it into a `return` through the C API.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: Oh, I see.  Right, that makes sense.

I kind of feel like an empty enclave is allowed (when not using security).  In that case, I think what you have here is fine.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Yes I'm aware of this issue and this is a temporary workaround (I left a similar comment in the function that returns the `std::optional<shm_manager>`).

My goal in subsequent PRs is to add functions like `rmw_context_impl_s::add_publisher()`, `rmw_context_impl_s::publish()` etc which will create/handle the publisher/subscribption/client/service instance inside `rmw_context_impl_s` directly without returning anything. This way we will lock the `session` -> create the instance -> add it to an internal map like in #259.

rmw_zenoh_cpp/src/rmw_init.cpp
Yadunund: Ah that's pretty neat! Will update accordingly!

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: ie this comment https://github.com/ros2/rmw_zenoh/blob/a9b34e2ede699450cc5dd968e87f172827cdbb28/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp#L54-L58

rmw_zenoh_cpp/src/rmw_init.cpp
Yadunund: After learning that `RMW_TRY_PLACEMENT_NEW` elegantly handles exceptions, I've updated the constructor here to also instantiate the Zenoh router and shm manager. https://github.com/ros2/rmw_zenoh/pull/258/commits/a6fb52136f499dd411de284a912ef0cd99b2e85f

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,584 @@
...
+{
+  if (is_initialized_) {
+    return RMW_RET_OK;
+  }
+  // Setup the liveliness subscriber to receives updates from the ROS graph
+  // and update the graph cache.
+  auto sub_options = zc_liveliness_subscriber_options_null();
+  z_owned_closure_sample_t callback = z_closure(
+    rmw_context_impl_s::graph_sub_data_handler, nullptr,
+    this->shared_from_this().get());
```
clalancette: Isn't this equivalent to `this`?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,584 @@
...
+  graph_guard_condition_(graph_guard_condition),
+  is_shutdown_(false),
+  next_entity_id_(0),
+  is_initialized_(false)
+{
+  // Do nothing.
+}
+
+///=============================================================================
+rmw_ret_t rmw_context_impl_s::Data::subscribe()
```
clalancette: I think the method name here could be a little more descriptive.  `subscribe_to_graph_cache`, maybe?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,584 @@
...
+{
+  if (is_initialized_) {
```
clalancette, line 107: Do we need to take the `mutex_`, for guarding access to `is_initialized_` and `graph_cache_`?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,584 @@
...
+}
+
+///=============================================================================
+rmw_context_impl_s::Data::~Data()
+{
+  RMW_TRY_DESTRUCTOR(
+    static_cast<rmw_zenoh_cpp::GuardCondition *>(
+      graph_guard_condition_->data)->~GuardCondition(),
+    rmw_zenoh_cpp::GuardCondition, );
+  if (rcutils_allocator_is_valid(allocator_)) {
```
clalancette: We might consider getting rid of this check; if we've gotten this far, then the `allocator_` had to have been valid, right?  Maybe we should move this check into the constructor instead?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,584 @@
...
+    rmw_zenoh_cpp::zenoh_router_check_attempts();
+  if (configured_connection_attempts.has_value()) {
+    ret = RMW_RET_ERROR;
+    uint64_t connection_attempts = 0;
+    // Retry until the connection is successful.
+    while (ret != RMW_RET_OK && connection_attempts < configured_connection_attempts.value()) {
+      if ((ret = rmw_zenoh_cpp::zenoh_router_check(z_loan(session))) != RMW_RET_OK) {
+        ++connection_attempts;
+      }
+      std::this_thread::sleep_for(std::chrono::seconds(1));
```
clalancette: I know you are just porting this over, but I'm just realizing that one second is a long time to wait here.  I would much rather see us check more often, like every 100 milliseconds or so.  I don't think we should fix it in this PR, but I'm going to suggest opening an issue (or adding a comment to an existing issue, if one is appropriate) to reconsider this.

MichaelOrlov:
@Yadunund Friendly ping to follow up on this issue.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,584 @@
...
+  if (data_ptr == nullptr) {
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "[graph_sub_data_handler] Unable to lock data_wp."
+    );
+    return;
+  }
+
+  // Update the graph cache.
+  std::lock_guard<std::mutex> lock(data_ptr->mutex_);
```
ahcorde, line 57: #include `<mutex>`

MichaelOrlov:
Discussion from maintenance triage: Decided to assign this issue to @Yadunund

clalancette:
> `rmw_context_impl_s::publish()` etc

I think this is what I'm struggling with most in this change.  It just doesn't seem right to me to be having one giant class that encapsulates all functionality of the RMW.  It does fix the locking problem, but it doesn't seem very elegant.

If we ignore the threading problem for the moment, in my ideal world we'd have a `ContextImpl` class at the top-level.  During `rmw_create_node`, we'd call `ContextImpl::create_node()`, which would return a `Node` object.  During `rmw_create_publisher`, we'd call `Node::create_publisher()`, which would return a `Publisher` object.  During `rmw_publish`, we'd call `Publisher::publish()`.  (There would be similar classes for `Subscription`, `ServiceClient`, and `ServiceServer`).  And the `GraphCache` would be embedded inside of the `ContextImpl` (since that is indeed a session-wide entity).  In that class hierarchy, the functionality for each of these entities is encapsulated into its own class, and further they don't have to share one giant lock.  This is, incidentally, how zenoh-cpp works (though there is no `Node` entity there, so you ask the `Session` to create a `Publisher` object).

Now we have to think about the locking.  If we just did the above, it wouldn't be much different from what we have today, and it wouldn't fix our locking problems.  So what we need is for each entity to be able to query its "parent" on whether it is still alive.  For instance, in `rmw_publish`, we'd have to ask the `Node` whether the `Publisher` is still alive.  But before that, we have to actually ask the `ContextImpl` whether the `Node` is still alive.  (There's an argument to be made here that we have a similar problem with `ContextImpl`, but I'm not sure we can solve that with the current RMW API).

Thus, the `Node` class would keep a list of active `Publisher`, `Subscription`, `ServiceServer`, and `ServiceClient` objects.  The `ContextImpl` class would keep a list of active `Node` objects.  And the opaque data (`rmw_publisher->data`, etc) that we return from `rmw_create_publisher` (and siblings) would be a pointer to the `ContextImpl`, some kind of identifier for the `Node`, and some kind of identifier for the `Publisher`.  Then we'd have enough information to walk the entire hierarchy and determine whether the thing we want is still alive.  During an `rmw_destroy_publisher`, we'd remove it from the `Node` list, and thus further calls wouldn't work.  `rmw_destroy_subscription` is a bit harder, because a callback may race here; we'd have to make the callback ask the node whether the subscription is still alive at this point.

I admit I haven't totally thought through all of the implications here.  But I think this proposal creates a class hierarchy, but also allows us to add in additional checking/locking.  What do you think of the general idea?

Yadunund:
> > `rmw_context_impl_s::publish()` etc
> 
> I think this is what I'm struggling with most in this change. It just doesn't seem right to me to be having one giant class that encapsulates all functionality of the RMW. It does fix the locking problem, but it doesn't seem very elegant.
> 
> If we ignore the threading problem for the moment, in my ideal world we'd have a `ContextImpl` class at the top-level. During `rmw_create_node`, we'd call `ContextImpl::create_node()`, which would return a `Node` object. During `rmw_create_publisher`, we'd call `Node::create_publisher()`, which would return a `Publisher` object. During `rmw_publish`, we'd call `Publisher::publish()`. (There would be similar classes for `Subscription`, `ServiceClient`, and `ServiceServer`). And the `GraphCache` would be embedded inside of the `ContextImpl` (since that is indeed a session-wide entity). In that class hierarchy, the functionality for each of these entities is encapsulated into its own class, and further they don't have to share one giant lock. This is, incidentally, how zenoh-cpp works (though there is no `Node` entity there, so you ask the `Session` to create a `Publisher` object).
> 
> Now we have to think about the locking. If we just did the above, it wouldn't be much different from what we have today, and it wouldn't fix our locking problems. So what we need is for each entity to be able to query its "parent" on whether it is still alive. For instance, in `rmw_publish`, we'd have to ask the `Node` whether the `Publisher` is still alive. But before that, we have to actually ask the `ContextImpl` whether the `Node` is still alive. (There's an argument to be made here that we have a similar problem with `ContextImpl`, but I'm not sure we can solve that with the current RMW API).
> 
> Thus, the `Node` class would keep a list of active `Publisher`, `Subscription`, `ServiceServer`, and `ServiceClient` objects. The `ContextImpl` class would keep a list of active `Node` objects. And the opaque data (`rmw_publisher->data`, etc) that we return from `rmw_create_publisher` (and siblings) would be a pointer to the `ContextImpl`, some kind of identifier for the `Node`, and some kind of identifier for the `Publisher`. Then we'd have enough information to walk the entire hierarchy and determine whether the thing we want is still alive. During an `rmw_destroy_publisher`, we'd remove it from the `Node` list, and thus further calls wouldn't work. `rmw_destroy_subscription` is a bit harder, because a callback may race here; we'd have to make the callback ask the node whether the subscription is still alive at this point.
> 
> I admit I haven't totally thought through all of the implications here. But I think this proposal creates a class hierarchy, but also allows us to add in additional checking/locking. What do you think of the general idea?

Thanks for the additional thoughts here. I agree such a hierarchy would be ideal. What i'm not clear about is whether `Node` will have an API to return a `Publisher::SharedPtr`? My only concern here is that we need to ensure that the only thing that `rmw_destroy_publisher` does is remove the `Publisher::SharedPtr` from the container in which it is stored within `Node`. Then the destructor of `Publisher` takes care of all the cleaning up. So even if `rmw_publish` obtained this `Publisher::SharedPtr` from `Node`, and is thus keeping it alive while `rmw_destroy_publisher` is invoked in a separate thread, the `Publisher's` destructor will only be invoked after `rmw_publish` returns. This would also be the case if `Node` is destroyed  when `rmw_publish` is being invoked.

In any case it seems like this is a bridge we will need to cross only after we merge this PR; perhaps when addressing #259? So should we move the discussion there since the changes here simple make `rmw_context_impl_s` a concrete class with private data members?

Yadunund:
In https://github.com/ros2/rmw_zenoh/pull/258/commits/7cc52bdb76ea9f4f8fb9b290483d1073b78a88d9, I updated the `rmw_context_impl_s` class to return a `shared_ptr<GraphCache>`. The `graph_sub_data_handler()` cb should still be thread-safe with the new implementation as the callback and other methods that query the `shared_ptr<GraphCache>` should lock the same mutex within `GraphCache`. 
Hoping this addresses the concern of `rmw_context_impl_s` class embedding everything within it. I will update #259 in a similar way such that `rmw_context_impl_s` returns a `shared_ptr<NodeData>`.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp
```diff
@@ -0,0 +1,137 @@
...
+public:
+  // Constructor that internally initializees the Zenoh session and other artifacts.
+  // Throws an std::runtime_error if any of the initializations fail.
+  // The construction will block until a Zenoh router is detected.
+  // TODO(Yadunund): Make this a non-blocking call by checking for the Zenoh
+  // router in a separate thread. Instead block when creating a node if router
+  // check has not succeeded.
+  rmw_context_impl_s(
+    const rcutils_allocator_t * allocator,
+    const std::size_t domain_id,
```
ahcorde, line 44: include `<cstddef>`

ahcorde (changes requested):

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: The header is included in `rmw_context_impl_s.hpp`

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp
Yadunund: Since we're compiling with C++, `std::size_t` is part of the standard library and should not need any additional includes. Please correct me if i'm mistaken.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
clalancette: While true, I am always of the opinion that you should include-what-you-use.  So I'd add it in here as well.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,434 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rmw_context_impl_s.hpp"
+
+#include <utility>
+#include <thread>
```
clalancette:
```suggestion
#include <mutex>
#include <thread>
#include <utility>
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: https://github.com/ros2/rmw_zenoh/issues/277

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: When I move `allocator_->deallocate(graph_guard_condition_->data, allocator_->state);` outside of this check, i get a segfault when running some of the rcl tests. Here's the command to run the test directly (you will need to update the path to your rolling workspace)

```
/usr/bin/python3 "-u" "home/yadunund/ros2_rolling/install/ament_cmake_test/share/ament_cmake_test/cmake/run_test.py" "/home/yadunund/ros2_rolling/build/rcl/test_results/rcl/test_info_by_topic__rmw_zenoh_cpp.gtest.xml" "--package-name" "rcl" "--output-file" "/home/yadunund/ros2_rolling/build/rcl/ament_cmake_gtest/test_info_by_topic__rmw_zenoh_cpp.txt" "--env" "RMW_IMPLEMENTATION=rmw_zenoh_cpp" "--command" "/home/yadunund/ros2_rolling/build/rcl/test/test_info_by_topic" "--gtest_output=xml:/home/yadunund/ros2_rolling/build/rcl/test_results/rcl/test_info_by_topic__rmw_zenoh_cpp.gtest.xml"

```

Does this imply the allocator is being destroyed sooner than expected somewhere in the rcl layer?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/258/commits/6ad087c383d5b164869b7a5a3f1f52247722e5a4

I didn't include it initially as I thought we might run into a deadlock between this thread that calls `zc_liveliness_declare_subscriber` and the Zenoh thread which might subscribe to a liveliness token and then trigger `rmw_context_impl_s::graph_sub_data_handler()` callback which also locks the same mutex.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Updated to `subscribe_to_ros_graph` https://github.com/ros2/rmw_zenoh/pull/258/commits/6ad087c383d5b164869b7a5a3f1f52247722e5a4

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: It totally is. I forgot to switch it back to `this` after playing around with `shared/weak_from_this`
https://github.com/ros2/rmw_zenoh/pull/258/commits/6ad087c383d5b164869b7a5a3f1f52247722e5a4

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/258/commits/6ad087c383d5b164869b7a5a3f1f52247722e5a4

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
clalancette: > Does this imply the allocator is being destroyed sooner than expected somewhere in the rcl layer?

Yeah, it has to be.

That said, looking at this more closely, this has the problem that I had elsewhere, in that "another entity" allocated these, and then the "Data" class is deallocating them.  While that can work if you are careful about transferring ownership, I think it is just simpler in general to have the entity that allocated them deallocate them (no transfer of ownership).  With that in mind, I'll suggest either moving the construction of the graph_guard_condition into the Data constructor, *or* moving the destruction here into the rmw_context_impl_s destructor.

Yadunund: I ended up not using the `allocator` when instantiating the `graph_guard_condition` https://github.com/ros2/rmw_zenoh/pull/258/commits/5be2532930a7ccefcf761811a601a3c3297a065f.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,383 @@
...
+  z_owned_str_t shm_enabled = zc_config_get(z_loan(config), "transport/shared_memory/enabled");
+  auto always_free_shm_enabled = rcpputils::make_scope_exit(
+    [&shm_enabled]() {
+      z_drop(z_move(shm_enabled));
+    });
+
+  // Initialize the zenoh session.
+  z_owned_session_t session = z_open(z_move(config));
+  if (!z_session_check(&session)) {
+    throw std::runtime_error("Error setting up zenoh session");
```
ahcorde:
```suggestion
    throw std::runtime_error("Error setting up zenoh session.");
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -0,0 +1,383 @@
...
+    if (z_reply_is_ok(&reply)) {
+      z_sample_t sample = z_reply_ok(&reply);
+      z_owned_str_t keystr = z_keyexpr_to_string(sample.keyexpr);
+      // Ignore tokens from the same session to avoid race conditions from this
+      // query and the liveliness subscription.
+      graph_cache->parse_put(z_loan(keystr), true);
+      z_drop(z_move(keystr));
+    } else {
+      RMW_ZENOH_LOG_DEBUG_NAMED(
+        "rmw_zenoh_cpp", "[rmw_context_impl_s] z_call received an invalid reply\n");
```
ahcorde:
```suggestion
        "rmw_zenoh_cpp", "[rmw_context_impl_s] z_call received an invalid reply.\n");
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp
```diff
@@ -0,0 +1,137 @@
...
+  // Note: This is not thread-safe.
+  // TODO(Yadunund): Remove this API and instead include a publish() API
+  // that handles the shm_manager once the context manages publishers.
+  std::optional<zc_owned_shm_manager_t> & shm_manager();
+
+  // Get the graph guard condition.
+  rmw_guard_condition_t * graph_guard_condition();
+
+  // Get a unique id for a new entity.
+  size_t get_next_entity_id();
```
ahcorde, line 65: include `<cstddef>`

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp
```diff
@@ -0,0 +1,137 @@
...
+    // Graph cache.
+    std::shared_ptr<rmw_zenoh_cpp::GraphCache> graph_cache_;
+    // ROS graph liveliness subscriber.
+    z_owned_subscriber_t graph_subscriber_;
+    // Equivalent to rmw_dds_common::Context's guard condition
+    /// Guard condition that should be triggered when the graph changes.
+    std::unique_ptr<rmw_guard_condition_t> graph_guard_condition_;
+    // The GuardCondition data structure.
+    rmw_zenoh_cpp::GuardCondition guard_condition_data_;
+    /// Shutdown flag.
```
ahcorde:
```suggestion
    // Shutdown flag.
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp
```diff
@@ -0,0 +1,137 @@
...
+    // zenoh session config.
+    std::optional<zc_owned_shm_manager_t> shm_manager_;
+    // Liveliness keyexpr string to subscribe to for ROS graph changes.
+    std::string liveliness_str_;
+    // Graph cache.
+    std::shared_ptr<rmw_zenoh_cpp::GraphCache> graph_cache_;
+    // ROS graph liveliness subscriber.
+    z_owned_subscriber_t graph_subscriber_;
+    // Equivalent to rmw_dds_common::Context's guard condition
+    /// Guard condition that should be triggered when the graph changes.
```
ahcorde:
```suggestion
    // Guard condition that should be triggered when the graph changes.
```

ahcorde (approved):
Some minor fixes, otherwise LGTM

clalancette (approved):
This needs a rebase to fix conflicts, and we should fix @ahcorde 's comments, but this otherwise looks good to me.

PR #261, opened 2024-08-08
Support running zenohd without console access

nkoenig:
I was playing around with running `rmw_zenohd` in a launch file, and it was not happy because of the KeyboardReader's attempt to access the console.

So, I've essentially captured the runtime error, and continue running the zenoh router if the console is not available. 

I haven't tested this on windows, and I don't know if this would break something else or violate some design decision.

### DIFF
--- a/rmw_zenoh_cpp/src/zenohd/main.cpp
+++ b/rmw_zenoh_cpp/src/zenohd/main.cpp
@@ -12,9 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <condition_variable>
 #include <chrono>
 #include <cstdio>
 #include <cstring>
+#include <memory>
+#include <mutex>
 #include <stdexcept>
 #include <string>
 #include <thread>
@@ -39,6 +42,8 @@
 #include "rcpputils/scope_exit.hpp"
 
 static bool running = true;
+static std::mutex run_mutex;
+static std::condition_variable run_cv;
 
 class KeyboardReader final
 {
@@ -160,14 +165,18 @@ private:
 BOOL WINAPI quit(DWORD ctrl_type)
 {
   (void)ctrl_type;
+  std::scoped_lock lock(run_mutex);
   running = false;
+  run_cv.notify_one();
   return true;
 }
 #else
 void quit(int sig)
 {
   (void)sig;
+  std::scoped_lock lock(run_mutex);
   running = false;
+  run_cv.notify_one();
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


Timple:
Personally I would just drop all keyboard handling/capturing. It's a very a-typical thing to do for a daemon.

clalancette:
> Personally I would just drop all keyboard handling/capturing. It's a very a-typical thing to do for a daemon.

How do you handle shutting down cleanly, then?

Timple:
> How do you handle shutting down cleanly, then?

Handle the sigint that arises from pressing Ctrl+c.

clalancette:
> Handle the sigint that arises from pressing Ctrl+c.

Ah, fair.  I was thinking that we needed this code to deal with that part of it, but you are right, we don't.

The honest truth is that we hope to get rid of this daemon completely, but we aren't quite there yet.  I think in the meantime, yeah, I'd be on board with just removing the keyboard handling completely.

nkoenig:
> to get rid of this d



> Ah, fair. I was thinking that we needed this code to deal with that part of it, but you are right, we don't.

I'll go ahead and remove it.
 
> The honest truth is that we hope to get rid of this daemon completely, but we aren't quite there yet. I think in the meantime, yeah, I'd be on board with just removing the keyboard handling completely.

I'd be happy to help on this effort. I'm not a zenoh expert, but I have time to spare.

clalancette:
> I'll go ahead and remove it.

OK, great.  I'll review this part of it, since I think we should get this in.

> I'd be happy to help on this effort. I'm not a zenoh expert, but I have time to spare.

Ah, cool.  Here is some context.  Back in late 2023, we actually didn't have a custom router, and we were just using the router as provided by the upstream Zenoh project.  The major problem with that was that it required us to build Zenoh twice: once for building the `zenoh-c` bindings, and once so we got access to the router.  Since building Zenoh is pretty slow, this was less than ideal.  So we introduced the custom router here.  But it isn't ideal, because while it supports *most* of the features of the zenohd router, it doesn't support all of them.  And that leads to confusion.

My best idea on how to fix this is to change https://github.com/eclipse-zenoh/zenoh-c so that it can (optionally) also build the router.  Then in the vendor package here, we would turn on that option, and also install the router in the correct place so that `ros2 run` could find it.  There may be other solutions, but that was the one I came up with.

Does that make sense?

rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -39,135 +35,25 @@
...
-#endif
-};
+static std::mutex run_mutex;
+static std::condition_variable run_cv;
 
 #ifdef _WIN32
 BOOL WINAPI quit(DWORD ctrl_type)
 {
   (void)ctrl_type;
+  std::scoped_lock lock(run_mutex);
```
clalancette: So we can't (safely) take a mutex inside of a signal handler, because it is possible for the mutex to block.

Luckily, we don't need to take the mutex in order to run `cv.notify_one()`.  Indeed, the documentation for it says that it is a [pessimization](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) (see the Notes).  Thus I think this should be removed:

```suggestion
```

rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -39,135 +35,25 @@
...
   running = false;
+  run_cv.notify_one();
   return true;
 }
 #else
 void quit(int sig)
 {
   (void)sig;
+  std::scoped_lock lock(run_mutex);
```
clalancette: Same thing here.

```suggestion
```

rmw_zenoh_cpp/src/zenohd/main.cpp
```diff
@@ -39,135 +35,25 @@
...
-};
+static std::mutex run_mutex;
+static std::condition_variable run_cv;
 
 #ifdef _WIN32
 BOOL WINAPI quit(DWORD ctrl_type)
 {
   (void)ctrl_type;
+  std::scoped_lock lock(run_mutex);
   running = false;
```
clalancette, line 45: Actually, I think we don't even need the `running` variable anymore.  Just the fact that the condition variable wakes up in the main thread means that a signal handler was called, which I think should be enough here.

```suggestion
```

clalancette (changes requested):

rmw_zenoh_cpp/src/zenohd/main.cpp
nkoenig: The value that `running` provides is the predicate check in the condition variable's `wait`. This would prevent spurious wakeups. I could remove it if you think the check isn't necessary.

clalancette: Oh right.  Good call.  That's fine, we can leave it in.

Yadunund (approved):
The changes look good to me! Shall wait for @clalancette to return from vacation sometime next week for a final review.

nkoenig:
> The changes look good to me! Shall wait for @clalancette to return from vacation sometime next week for a final review.

Up to you @Yadunund . It's not super urgent for me.

clalancette (approved):

PR #266, opened 2024-08-22
Skip liveliness parts for qos elements with default values

Yadunund:
This PR is a follow-up to https://github.com/ros2/rmw_zenoh/pull/234. 

Instead of convering the qos value to a string and appending to the liveliness token's keyexpression, we only do so when the qos value deviates from the `QoS::default_qos()`. This reduces the number of bytes we need to pack into the liveliness tokens.

Prior to this PR, the publisher that spins when running `ros2 topic pub /chatter std_msgs/msg/String '{data: hello}' --qos-reliability reliable --qos-durability transient_local  --qos-depth 42  --qos-history keep_last`, the liveliness token would look like

```
@/liveliness/@ros2_lv/0/169ba0d08c2b76ba4f7e8f99cb450fe/0/4/MP/%/%/_ros2cli_42493/%chatter/std_msgs::msg::dds_::String_/RIHS01_df668c740482bbd48fb39d76a70dfd4bd59db1288021743503259e948f6b1a18/1:1:1,42:9223372036,854775807:9223372036,854775807:1,9223372036,854775807
```

Now the token looks like
```
@/liveliness/@ros2_lv/0/5af83680cf352c29b0a98954d16a5470/0/4/MP/%/%/_ros2cli_43208/%chatter/std_msgs::msg::dds_::String_/RIHS01_df668c740482bbd48fb39d76a70dfd4bd59db1288021743503259e948f6b1a18/::,:,:,:,,
```

The diff better highlights the differences

```diff
- @/liveliness/@ros2_lv/0/169ba0d08c2b76ba4f7e8f99cb450fe/0/4/MP/%/%/_ros2cli_42493/%chatter/std_msgs::msg::dds_::String_/RIHS01_df668c740482bbd48fb39d76a70dfd4bd59db1288021743503259e948f6b1a18/1:1:1,42:9223372036,854775807:9223372036,854775807:1,9223372036,854775807
+ @/liveliness/@ros2_lv/0/169ba0d08c2b76ba4f7e8f99cb450fe/0/4/MP/%/%/_ros2cli_42493/%chatter/std_msgs::msg::dds_::String_/RIHS01_df668c740482bbd48fb39d76a70dfd4bd59db1288021743503259e948f6b1a18/::,:,:,:,,

```

I've tested this PR against the `rcl` tests and no new errors.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
@@ -178,12 +178,11 @@ std::optional<T> str_to_size_t(const std::string & str, const T default_value)
   }
   errno = 0;
   char * endptr;
-  // TODO(Yadunund): strtoul and strtol both return long int and not size_t so we
-  // should consider fixing this implementation if this function is moved to
-  // a header file and will be used by other parts of the codebase.
-  size_t num = std::is_signed<decltype(default_value)>::value ?
-    strtol(str.c_str(), &endptr, 10) :
-    strtoul(str.c_str(), &endptr, 10);
+  // TODO(Yadunund): strtoul returns an unsigned long, not size_t.
+  // Depending on the architecture and platform, these may not be the same size.
+  // Further, if the incoming str is a signed integer, storing it in a size_t is incorrect.
+  // We should fix this piece of code to deal with both of those situations.
+  size_t num = strtoul(str.c_str(), &endptr, 10);
   if (endptr == str.c_str()) {
     // No values were converted, this is an error
     RMW_SET_ERROR_MSG("no valid numbers available");
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "logging_macros.hpp"
+#include "qos.hpp"
 
 #include "rcpputils/scope_exit.hpp"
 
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -232,19 +264,26 @@ std::optional<rmw_qos_profile_t> keyexpr_to_qos(const std::string & keyexpr)
...
+      liveliness_parts[0]);
   } catch (const std::exception & e) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("Error setting QoS values from strings: %s", e.what());
     return std::nullopt;
   }
 
   // Helper function to convert string to size_t.
   auto str_to_size_t =
-    [](const std::string & str) -> std::optional<size_t>
+    [](const std::string & str, const std::size_t default_value) -> std::optional<size_t>
```
clalancette: I'm concerned about the `size_t` here.  In particular, it is not always the case that `sizeof(size_t)` == `sizeof(uint64_t)`.  Yet the deadline, lifespan, and liveliness changes down below assume that it is the case (for the `sec` and `nsec`).

What I'm going to suggest instead is that we make this a "real" function, and templatize `default_value`.  Then all of those will get code generated properly.

Something like:

```
template<typename T>
static std::optional<T> str_to_size_t(const std::string & str, const T default_value)
{
  if (str.empty()) {
    return default_value;
  }
  errno = 0;
  char * endptr;
  size_t num = strtoul(str.c_str(), &endptr, 10);
  if (endptr == str.c_str()) {
    // No values were converted, this is an error
    RMW_SET_ERROR_MSG("no valid numbers available");
    return std::nullopt;
  } else if (*endptr != '\0') {
    // There was junk after the number
    RMW_SET_ERROR_MSG("non-numeric values");
    return std::nullopt;
  } else if (errno != 0) {
    // Some other error occurred, which may include overflow or underflow
    RMW_SET_ERROR_MSG(
      "an undefined error occurred while getting the number, this may be an overflow");
    return std::nullopt;
  }
  return num;
}
```

clalancette (changes requested):
Overall, this looks great, and is a great improvement.  I have one thing I think we should fix inline, but then this should be good to go.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Great suggestion; I've modified accordingly https://github.com/ros2/rmw_zenoh/pull/266/commits/aa5697d9c0c5bb8e8da7c4a743048fe7324734dc

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -227,6 +230,35 @@ std::optional<rmw_qos_profile_t> keyexpr_to_qos(const std::string & keyexpr);
 ///=============================================================================
 /// Convert a Zenoh id to a string.
 std::string zid_to_str(const z_id_t & id);
+
+///=============================================================================
+// Helper function to convert string to size_t.
+// The function is templated to enable conversion to size_t or std::size_t.
+template<typename T>
+std::optional<T> str_to_size_t(const std::string & str, const T default_value)
```
clalancette: This whole function doesn't need to live in the header file; we can just keep it as an anonymous namespace function within `liveliness_utils.cpp`.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -227,6 +230,35 @@ std::optional<rmw_qos_profile_t> keyexpr_to_qos(const std::string & keyexpr);
...
+{
+  if (str.empty()) {
+    return default_value;
+  }
+  errno = 0;
+  char * endptr;
+  size_t num = strtoul(str.c_str(), &endptr, 10);
```
clalancette: Thinking about it more today, I realized two problems here:

1.  `strtoul` is documented as returning an `unsigned long`, not a `size_t`.  This is an existing bug.
2.  If T ended up being a signed integer, we'd likely do the incorrect thing here.  We could fix this either by making the template type only support unsigned (any signed integers would fail to compile), or we could do some other template magic to choose `strtoul` or `strtol` as appropriate.

I would say neither of those are blockers, but if you decide not to fix them I'd appreciate a TODO comment explaining what should be improved here.

clalancette (changes requested):

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
Yadunund: Thanks for the feedback! I've left 1) as a TODO but addressed 2) with an `std::is_signed` check. See https://github.com/ros2/rmw_zenoh/pull/266/commits/bd3722f376cca97466ceb9f0f90ff035db60ec9e

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -20,15 +20,14 @@
 #include <stdexcept>
 #include <string>
 #include <unordered_map>
+#include <type_traits>
```
ahcorde, line 22: alphabetize

ahcorde (changes requested):
Just a minor nit, otherwise LGTM

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
Yadunund: Nice catch https://github.com/ros2/rmw_zenoh/pull/266/commits/8159b0bcfad2d1f466840764e5f769c49812a383

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -167,48 +166,113 @@ std::vector<std::string> split_keyexpr(
...
+    return default_value;
+  }
+  errno = 0;
+  char * endptr;
+  // TODO(Yadunund): strtoul and strtol both return long int and not size_t so we
+  // should consider fixing this implementation if this function is moved to
+  // a header file and will be used by other parts of the codebase.
+  size_t num = std::is_signed<decltype(default_value)>::value ?
+    strtol(str.c_str(), &endptr, 10) :
+    strtoul(str.c_str(), &endptr, 10);
```
clalancette: So this is a bit trickier, because `strtol` returns a `long` (signed), and `strtoul` returns an `unsigned long` (unsigned).  Thus, it isn't totally correct to assign them both to a `size_t`, particularly the signed long.

I'm going to suggest that for now, we just leave this as it was and add the TODO comment:

```suggestion
  // TODO(Yadunund): strtoul returns an unsigned long, not size_t.
  // Depending on the architecture and platform, these may not be the same size.
  // Further, if the incoming str is a signed integer, storing it in a size_t is incorrect.
  // We should fix this piece of code to deal with both of those situations.
  size_t num = strtoul(str.c_str(), &endptr, 10);
```

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -23,6 +23,9 @@
 #include <string>
 #include <vector>
 
+#include "logging_macros.hpp"
+
+#include "rmw/error_handling.h"
```
clalancette: Both of these can be moved back into `liveliness_utils.cpp`.

clalancette (changes requested):
Two more nits, then this will be good to go from my perspective.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
Yadunund: oops my bad https://github.com/ros2/rmw_zenoh/pull/266/commits/dd4498391084a774d79f2271b68d87154ec55f4e

clalancette (approved):
This looks great, thanks for iterating.

PR #269, opened 2024-08-26
Querying subs to query newly discovered publishers with publication caches

### LINKED ISSUE #263: Subscribers sometimes not getting messages from topics using durability transient_local
# Version
commit b78d6a15ca47681ddd11613c085fe41420fa9168
Date:   Wed Aug 7 13:51:57 2024 +0200

# Platform
Ubuntu 22, Docker

# ROS Version
Iron

# Description
Hi, I tried to run our codebase tests with zenoh-rmw, and noticed that many of those relying on latched (durability tranisent_local) topics experience seldom failures: the subscribers sometimes fail to get the data from "recently spawned" publishers.

I created a small package where I manage to reproduce the issue (happens roughly 50% of the times on my setup, never reproducible if switching to Cyclone DDS):

https://github.com/nirwester/zenoh_pub_bug

The instructions to execute it are in the README.

Let me know if I can provide additional information.

Yadunund:
Fix #263 

This PR:
* Renames `Entity::keyexpr()` to `Entity::liveliness_keyexpr()` to disambiguate between keyexpressions used for liveliness tokens and keyexpressions used for topic/service names.
* Adds `topic_keyexpr` string to `TopicInfo` stored within `Entity`. This way, entities constructed within `GraphCache` are aware of they topic keyexpressions.
* Adds an API to register callbacks to trigger when a Publisher with a PublicationCache is detected in the graph.
* Updates `rmw_create_subscription` to register a callback as described above that will call `ze_querying_subscriber_get` over the keyexpression. This requires the zenoh session ID as the "selector" to uniquely identify which PublicationCache to query.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -341,7 +341,7 @@ void GraphCache::parse_put(
   if (ignore_from_current_session && is_entity_local(*entity)) {
     RMW_ZENOH_LOG_DEBUG_NAMED(
       "rmw_zenoh_cpp",
-      "Ignoring parse_put for %s from the same session.\n", entity->keyexpr().c_str());
+      "Ignoring parse_put for %s from the same session.\n", entity->liveliness_keyexpr().c_str());
     return;
   }
 
@@ -559,7 +559,7 @@ void GraphCache::parse_del(
   if (ignore_from_current_session && is_entity_local(*entity)) {
     RMW_ZENOH_LOG_DEBUG_NAMED(
       "rmw_zenoh_cpp",
-      "Ignoring parse_del for %s from the same session.\n", entity->keyexpr().c_str());
+      "Ignoring parse_del for %s from the same session.\n", entity->liveliness_keyexpr().c_str());
     return;
   }
   // Lock the graph mutex before accessing the graph.
@@ -1315,4 +1315,20 @@ std::unique_ptr<rmw_zenoh_event_status_t> GraphCache::take_event_status(
   status_to_take.current_count_change = 0;
   return result;
 }
+
+///=============================================================================
+void GraphCache::set_querying_subscriber_callback(
+  const std::string & keyexpr,
+  QueryingSubscriberCallback cb)
+{
+  auto cb_it = querying_subs_cbs_.find(keyexpr);
+  if (cb_it == querying_subs_cbs_.end()) {
+    std::vector<QueryingSubscriberCallback> cbs = {};
+    cbs.push_back(std::move(cb));
+    querying_subs_cbs_[keyexpr] = std::move(cbs);
+    return;
+  }
+  cb_it->second.push_back(std::move(cb));
+}
+
 }  // namespace rmw_zenoh_cpp
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -406,7 +406,10 @@ void GraphCache::parse_put(
 
   // If the newly added entity is a publisher with transient_local qos durability,
   // we trigger any registered querying subscriber callbacks.
-  if (entity->topic_info().has_value() && entity->type() == liveliness::EntityType::Publisher) {
+  if (entity->type() == liveliness::EntityType::Publisher &&
+    entity->topic_info().has_value() &&
+    entity->topic_info()->qos_.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL)
+  {
     auto sub_cbs_it = querying_subs_cbs_.find(entity->topic_info()->topic_keyexpr_);
     if (sub_cbs_it != querying_subs_cbs_.end()) {
       for (const auto & cb : sub_cbs_it->second) {
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -403,6 +403,17 @@ void GraphCache::parse_put(
   // Otherwise, the entity represents a node that already exists in the graph.
   // Update topic info if required below.
   update_topic_maps_for_put(node_it->second, entity);
+
+  // If the newly added entity is a publisher with transient_local qos durability,
+  // we trigger any registered querying subscriber callbacks.
+  if (entity->topic_info().has_value() && entity->type() == liveliness::EntityType::Publisher) {
+    auto sub_cbs_it = querying_subs_cbs_.find(entity->topic_info()->topic_keyexpr_);
+    if (sub_cbs_it != querying_subs_cbs_.end()) {
+      for (const auto & cb : sub_cbs_it->second) {
+        cb(entity->zid());
+      }
+    }
+  }
 }
 
 ///=============================================================================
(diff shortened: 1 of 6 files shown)


Yadunund:
This PR will likely remain in draft until we migrate to Zenoh 1.0.0. https://github.com/ros2/rmw_zenoh/pull/269/commits/f4dc8ea1a7b93d74527ba5f9a12e9e8c7542c488 set `query_accept_replies` to `ZCU_REPLY_KEYEXPR_ANY` but this option is passed at QueryingSubscriber creation and is used only for the [initial query](https://github.com/eclipse-zenoh/zenoh/blob/e66745ebdf70f01bf01ba60f69bc902b917b2d24/zenoh-ext/src/querying_subscriber.rs#L237C9-L253C54). It's not used for subsequent calls to `ze_querying_subscriber_get` which is what we need to fix the original problem.
The ability to pass this option to the `ze_querying_subscriber_get()` call that is triggered in the callback will only land in `1.0.0`.

Yadunund:
I've temporarily merged changes from https://github.com/eclipse-zenoh/zenoh-c/pull/620 and I can confirm this PR now fixes the original issue. Once upstream `eclipse-zenoh/zenoh-c` is updated, we can re-update the vendored `zenoh-c` commit hash before merging. In the meanwhile @clalancette do you mind reviewing the changes?

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -47,8 +47,27 @@ NodeInfo::NodeInfo(
...
 }
 
+namespace
+{
+// Helper function to create a copy of a string after removing any
+// leading or trailing slashes.
+std::string strip_slashes(const std::string & str)
+{
+  std::string ret = str;
+  std::size_t start = 0;
```
ahcorde, line 59: #include `<cstddef>`

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -354,7 +379,7 @@ Entity::Entity(
   for (std::size_t i = 0; i < KEYEXPR_INDEX_MAX + 1; ++i) {
     bool last = false;
     if (!keyexpr_parts[i].empty()) {
-      this->keyexpr_ += std::move(keyexpr_parts[i]);
+      this->liveliness_keyexpr_ += std::move(keyexpr_parts[i]);
```
ahcorde, line 431: include `<utility>`

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -50,9 +50,11 @@ struct TopicInfo
   std::string name_;
   std::string type_;
   std::string type_hash_;
+  std::string topic_keyexpr_;
   rmw_qos_profile_t qos_;
 
   TopicInfo(
+    std::size_t domain_id,
```
ahcorde, line 57: include `<cstddef>`

ahcorde (changes requested):

Yadunund:
The API in zenoh-c has been fixed and i've updated the commit hash accordingly. This PR is ready for review.

Yadunund:
Hmm i've been unable to successfully run the nav2 demo with this PR. More specifically, the costmap does not show up in RViz and i'm unable to localize the robot. Digging into whether the problem is caused by the changes in the code here or perhaps a regression somewhere in zenoh-c between the old commit and the new one.

Yadunund:
I've narrowed the problem to this line that this PR added 
https://github.com/ros2/rmw_zenoh/blob/4188ea1157077bff1383b78629177dd1374da09d/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L623


If I comment out the line, running `ros2 launch nav2_bringup tb3_simulation.launch.py headless:=False` is able to successfully start up

image

Else I'm seeing

image


But the whole point of the bug fix is to include the `queryable_prefix` option to the publication cache. Perhaps some other options are not set correctly?

cc @JEnoch @imstevenpmwork @YuanYuYuan

JEnoch:
@Yadunund this PR fixes this issue: https://github.com/ros2/rmw_zenoh/pull/274

TL;DR: the QueryingSubscriber performs an initial query that must now be adapted following the addition of `zid` as a prefix in PublicationCache's queryable.
Detailed explanation in the PR.

Yadunund:
> @Yadunund this PR fixes this issue: #274
> 
> TL;DR: the QueryingSubscriber performs an initial query that must now be adapted following the addition of `zid` as a prefix in PublicationCache's queryable. Detailed explanation in the PR.

Thanks for the fix! I've merged it into this PR. This is now ready for another round of review.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2849,11 +2850,28 @@ rmw_create_service(
...
+      node->context->actual_domain_id,
+      rmw_service->service_name,
+      std::move(service_type),
+      type_hash_c_str,
+      service_data->adapted_qos_profile}
+  );
+  if (service_data->entity == nullptr) {
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to generate keyexpr for liveliness token for the service.");
```
ahcorde: Add `rmw_service->service_name` in the trace error?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -614,11 +570,31 @@ rmw_create_publisher(
...
+      node->context->actual_domain_id,
+      rmw_publisher->topic_name,
+      publisher_data->type_support->get_name(),
+      type_hash_c_str,
+      publisher_data->adapted_qos_profile}
+  );
+  if (publisher_data->entity == nullptr) {
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to generate keyexpr for liveliness token for the publisher.");
```
ahcorde: add `rmw_publisher->topic_name` in the error message?

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -1315,4 +1329,20 @@ std::unique_ptr<rmw_zenoh_event_status_t> GraphCache::take_event_status(
...
+  QueryingSubscriberCallback cb)
+{
+  auto cb_it = querying_subs_cbs_.find(keyexpr);
+  if (cb_it == querying_subs_cbs_.end()) {
+    std::vector<QueryingSubscriberCallback> cbs = {};
+    cbs.push_back(std::move(cb));
+    querying_subs_cbs_[keyexpr] = std::move(cbs);
+    return;
+  }
+  cb_it->second.push_back(std::move(cb));
```
clalancette: I think we can tighten this up:

```suggestion
  auto cb_it = querying_subs_cbs_.find(keyexpr);
  if (cb_it == querying_subs_cbs_.end()) {
    querying_subs_cbs_[keyexpr] = std::move(std::vector<QueryingSubscriberCallback>{});
    cb_it = querying_subs_cbs_.find(keyexpr);
  }
  cb_it->second.push_back(std::move(cb));
```

clalancette (changes requested):
I've left one minor thing to improve.  Once that is done, and @ahcorde 's comments are addressed, I think this will be good to go.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/269/commits/aef564483916cb723f56a22edb1d935ec89afce0

ahcorde (approved):

clalancette (approved):
Thanks for all of the changes here!  Looks good to me.

PR #274, opened 2024-08-28
Adapt the QueryingSubscriber's initial query to the new queryable_prefix

JEnoch:
This PR fixes the issue reported with Nav2 in https://github.com/ros2/rmw_zenoh/pull/269#issuecomment-2314220213 .

As the PublicationCache has changed by adding the `zid` as a prefix to it's queryable, all the QueryingSubscriber queries must be adapted to take this prefix in account:
- on PublicationCache's LivelinessToken discovery, setting the discovered `zid` for call to `ze_querying_subscriber_get`.  
  That's already done.
- on QueryingSubscriber creation since it performs an initial query to all possibly existing PublicationCaches that have not been discovered yet, using `"*"` in key expression in place of `zid`.  
  That was missing and fixed by this PR
  
Also, as this initial query is now using a `"*"`, the query target can't any longer be `COMPLETE`. No PublicationCache can be complete with such a query.

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -1468,6 +1468,8 @@ rmw_create_subscription(
     // This allows us to selectively query certain PublicationCaches when defining the
     // set_querying_subscriber_callback below.
     sub_options.query_accept_replies = ZCU_REPLY_KEYEXPR_ANY;
+    // As this initial query is now using a "*", the query target is not COMPLETE.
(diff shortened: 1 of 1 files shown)


Yadunund (approved):
@JEnoch thanks a lot for fixing the issue and providing an explanation. I pushed https://github.com/ros2/rmw_zenoh/pull/274/commits/57d1efdf1a663e63cb102f8df052a851d8c277dd to make the `query_target` explicit so that we don't accidentally assign an incorrect value in the future.

PR #278, opened 2024-09-09
Thread-safe access to PublisherData

Yadunund:
This PR replaces `class rmw_publisher_data_t` with `class PublisherData` which provides thread-safe access to publisher functions. An instance of `PublisherData` is created via the `NodeData` class. `rmw_publisher->data` is now set to `rmw_node_t`. The assumption here is that the node outlives the publisher similar to the assumption that the context outlives the node. From this `rmw_node_t` raw_ptr, we are able to obtain it's `NodeData` and then lookup the `PublisherData` for the publisher.

One nice thing this PR introduces is the mechanism where shutting down or destroying a parent entity, will shutdown the child entities. Eg. If `rclcpp::shutdown()` is invoked, it will first `shutdown` all the `PublisherData`, then `NodeData`, and finally the context. We also check for shutdown in the destructor. This is helpful when someone calls `node.reset` there by destroying an `rclcpp::Node`; this will now shutdown all child entities (only `PublisherData` so far) and then shutdown the node before clearing memory. We also check for `is_shutdown` before creating a publisher, publishing a message, etc. These principles will be extended to other entities in subsequent PRs.


rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -430,69 +418,29 @@ rmw_create_publisher(
...
-      RMW_TRY_DESTRUCTOR_FROM_WITHIN_FAILURE(
-        publisher_data->type_support->~MessageTypeSupport(),
-        rmw_zenoh_cpp::MessageTypeSupport);
-    });
-
-  publisher_data->context = node->context;
-  rmw_publisher->data = publisher_data;
+  // Store type erased node in rmw_publisher->data so that the
+  // PublisherData can be safely accessed.
+  rmw_publisher->data = reinterpret_cast<void *>(const_cast<rmw_node_t *>(node));
```
Yadunund, line 441: I had to rely on a `reinterpret_cast<const_cast<>>` to be able to store the `const rmw_node_t * node` available to `rmw_create_publisher()` within `rmw_publisher->data` as the `data` field is `void *` and not `const void *`. 

This is a bit messy. Another alternative I considered was storing the `rmw_context_impl_s *` ptr here and then including an `unordered_map` in this class to obtain the `NodeData` for a given `rmw_publisher_t` but this involves extra bookkeeping that we need to do when we create and destroy a publisher. Open to other suggestions!

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -657,4 +664,17 @@ std::string demangle_name(const std::string & input)
...
 }
 }  // namespace liveliness
+
+void
+generate_random_gid(uint8_t gid[RMW_GID_STORAGE_SIZE])
+{
+  std::random_device dev;
+  std::mt19937 rng(dev());
+  std::uniform_int_distribution<std::mt19937::result_type> dist(
+    std::numeric_limits<unsigned char>::min(), std::numeric_limits<unsigned char>::max());
```
ahcorde, line 675: include `<limits>`

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -149,8 +149,17 @@ rmw_ret_t rmw_context_impl_s::Data::subscribe_to_ros_graph()
...
-    return RMW_RET_OK;
+    return ret;
+  }
+
+  // Shutdown all the nodes in this context.
+  for (auto node_it = nodes_.begin(); node_it != nodes_.end(); ++node_it) {
+    ret = node_it->second->shutdown();
+    if (ret != RMW_RET_OK) {
+      return ret;
+    }
```
clalancette: I'm not convinced that we should get out early in case of a failure.  Definitely we should log it, but then I think we should go on and shutdown the rest of the nodes anyway.  I feel like that is slightly safer.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -0,0 +1,456 @@
...
+  }
+
+  return RMW_RET_OK;
+}
+
+///=============================================================================
+std::shared_ptr<const liveliness::Entity> PublisherData::entity() const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  return entity_;
```
clalancette: I'm not fond of this method (and the ones that follow).  They hold the lock briefly, but then the caller gets a pointer to the entity, which it can access unlocked.  Is there a follow-up plan to making these thread-safe?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -329,22 +328,6 @@ rmw_fini_publisher_allocation(
   return RMW_RET_UNSUPPORTED;
 }
 
-namespace
-{
-void
-generate_random_gid(uint8_t gid[RMW_GID_STORAGE_SIZE])
```
clalancette, line 335: Because we are removing this, we can also probably remove the `#include <random>` at the top of the file.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: It's a bit ugly, but I think it is OK.  At least, I think it is the best we can do given our constraints.

clalancette (changes requested):

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
Yadunund: The `Entity` is read-only (no methods to change it's state) so it should be ok?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/278/commits/3c1e1eb0e6aba8e7fbf4befa73dca7bda5c9eede

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Good point! done https://github.com/ros2/rmw_zenoh/pull/278/commits/3c1e1eb0e6aba8e7fbf4befa73dca7bda5c9eede

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
Yadunund: The alternative would be to add APIs that expose methods of `Entity` and `EventsManager`. eg `PublisherData::guid()` which would return `entity_->guid()`.  Similarly `PublisherData::add_new_event()` which would invoke `events_manager_->add_new_event()`

rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp
```diff
@@ -0,0 +1,116 @@
...
+    std::optional<zc_owned_shm_manager_t> & shm_manager);
+
+  // Get the entity from this PublisherData.
+  std::shared_ptr<const liveliness::Entity> entity() const;
+
+  // Get the next sequence number.
+  size_t get_next_sequence_number();
+
+  // Get the GID of this PublisherData.
+  const uint8_t * gid() const;
```
ahcorde: include `<cstdint>`

rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp
```diff
@@ -0,0 +1,116 @@
...
+  // Publish a ROS message.
+  rmw_ret_t publish(
+    const void * ros_message,
+    std::optional<zc_owned_shm_manager_t> & shm_manager);
+
+  // Get the entity from this PublisherData.
+  std::shared_ptr<const liveliness::Entity> entity() const;
+
+  // Get the next sequence number.
+  size_t get_next_sequence_number();
```
ahcorde: include `<cstddef>`

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3775,13 +3506,19 @@ rmw_get_gid_for_publisher(const rmw_publisher_t * publisher, rmw_gid_t * gid)
...
+    static_cast<rmw_context_impl_s *>(node->context->impl);
+  RMW_CHECK_ARGUMENT_FOR_NULL(context_impl, RMW_RET_INVALID_ARGUMENT);
+  auto node_data = context_impl->get_node_data(node);
+  RMW_CHECK_ARGUMENT_FOR_NULL(node_data, RMW_RET_INVALID_ARGUMENT);
+  auto pub_data = node_data->get_pub_data(publisher);
   RMW_CHECK_ARGUMENT_FOR_NULL(pub_data, RMW_RET_INVALID_ARGUMENT);
 
   gid->implementation_identifier = rmw_zenoh_cpp::rmw_zenoh_identifier;
-  memcpy(gid->data, pub_data->pub_gid, RMW_GID_STORAGE_SIZE);
+  memcpy(gid->data, pub_data->gid(), RMW_GID_STORAGE_SIZE);
```
ahcorde: include `<cstring>`

ahcorde (changes requested):

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
Yadunund: I've updated the implementation to not return`Entity::SharedPtr`. Instead, the required methods from `Entity` are exposed. This also improves the interaction with `GraphCache` since the `Entity` used to add/remove callbacks is technically a different one from the `Entity` created within `GraphCache` (from liveliness token). However, their `guids` are the same and the hashing was based on this `guid`. the `GraphCache` API now adds/removes callbacks with the `guids` which makes things cleaner.

I didn't do this with `EventsManager` for two reasons
- We need to cast `rmw_event->data` to a common type for both publisher and subscription datas. I can't think of a good way to do this given that `rmw_event->data` is `void *` type erased without relying on inheritance.
- The `EventManager` locks an [internal mutex](https://github.com/ros2/rmw_zenoh/blob/824886a4ea9fd8a032139fbb31d0181df1731407/rmw_zenoh_cpp/src/detail/event.hpp#L142) when accessing other data members so having an `EventsManager::SharedPtr` and using it across different threads should be safe.


Commit:
https://github.com/ros2/rmw_zenoh/pull/278/commits/071242996ee235ae5aed872274fed44dcf7f2808

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -289,13 +290,13 @@ class GraphCache final
...
   using GraphEventCallbacks = std::unordered_map<rmw_zenoh_event_type_t, GraphCacheEventCallback>;
-  // Map entity (based on uuid) to a map of event callbacks.
+  // Map an entity's guid to a map of event callbacks.
   // Note: Since we use unordered_map, we will only store a single callback for an
   // entity string. So we do not support the case where a node create a duplicate
   // pub/sub with the exact same topic, type & QoS but registers a different callback
   // for the same event type. We could switch to a multimap here but removing the callback
   // will be impossible right now since entities do not have unique IDs.
-  using GraphEventCallbackMap = std::unordered_map<liveliness::ConstEntityPtr, GraphEventCallbacks>;
+  using GraphEventCallbackMap = std::unordered_map<std::size_t, GraphEventCallbacks>;
```
ahcorde, line 300: include `<cstddef>`

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -230,8 +233,13 @@ std::optional<rmw_qos_profile_t> keyexpr_to_qos(const std::string & keyexpr);
 /// Convert a Zenoh id to a string.
 std::string zid_to_str(const z_id_t & id);
 }  // namespace liveliness
+
+///=============================================================================
+// Helper function to generate a randon GID.
+void generate_random_gid(uint8_t gid[RMW_GID_STORAGE_SIZE]);
```
ahcorde, line 241: include `<cstdint>`

ahcorde: and also `<cstddef>`, `std::size_t` is used in the header

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -230,8 +233,13 @@ std::optional<rmw_qos_profile_t> keyexpr_to_qos(const std::string & keyexpr);
...
 }  // namespace rmw_zenoh_cpp
 
+
```
ahcorde: remove

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -149,8 +149,22 @@ rmw_ret_t rmw_context_impl_s::Data::subscribe_to_ros_graph()
...
+    return ret;
+  }
+
+  // Shutdown all the nodes in this context.
+  for (auto node_it = nodes_.begin(); node_it != nodes_.end(); ++node_it) {
+    ret = node_it->second->shutdown();
+    if (ret != RMW_RET_OK) {
+      RMW_ZENOH_LOG_ERROR_NAMED(
+        "rmw_zenoh_cpp",
+        "Unable to shutdown node with id %zu. rmw_ret_t code: %zu",
```
ahcorde:
```suggestion
        "Unable to shutdown node with id %zu. rmw_ret_t code: %zu.",
```

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+///=============================================================================
+rmw_qos_profile_t PublisherData::adapted_qos_profile() const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  return entity_->topic_info()->qos_;
+}
+
+
+///=============================================================================
```
ahcorde:
```suggestion
}

///=============================================================================
```

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -0,0 +1,454 @@
...
+  {
+    shmbuf = zc_shm_alloc(
+      &shm_manager.value(),
+      max_data_length);
+    if (!z_check(shmbuf.value())) {
+      zc_shm_gc(&shm_manager.value());
+      shmbuf = zc_shm_alloc(&shm_manager.value(), max_data_length);
+      if (!z_check(shmbuf.value())) {
+        // TODO(Yadunund): Should we revert to regular allocation and not return an error?
+        RMW_SET_ERROR_MSG("Failed to allocate a SHM buffer, even after GCing");
```
ahcorde:
```suggestion
        RMW_SET_ERROR_MSG("Failed to allocate a SHM buffer, even after GCing.");
```

ahcorde (changes requested):
just some mirmo things, but overall LGTM

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
clalancette: In general, I like what you've done with this change.  I have one more suggestion of something I think we should fix, then I think this will be good.  I'll make a separate comment about that.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -0,0 +1,454 @@
...
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  return entity_->topic_info().value();
+}
+
+///=============================================================================
+const uint8_t * PublisherData::gid() const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  return gid_;
```
clalancette: Because this is returning a raw pointer, this is probably not thread-safe.

That said, from what I can tell there is only one caller of this API, from `rmw_get_gid_for_publisher`.  And there, what we are doing is copying the data into an out parameter.

That suggests to me that this function should be something like:

```
const void PublisherData::copy_gid(rmw_gid_t * gid) const
{
  std::lock_guard<std::mutex> lock(mutex_);
  memcpy(gid->data, gid_, RMW_GID_STORAGE_SIZE);
}
```

(and then update the caller down below as needed).  That way we are doing the copy under the lock, which should be safe.

clalancette (changes requested):

rmw_zenoh_cpp/src/detail/graph_cache.hpp
Yadunund: Done https://github.com/ros2/rmw_zenoh/pull/278/commits/cd27cb5164f9ab1b147edfc4592c8240c40818b8

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
Yadunund: Good idea! Also had to do something similar for copying the GID into the attachment map https://github.com/ros2/rmw_zenoh/pull/278/commits/4edbee2a038869d71529243748f0534a2258bae5

clalancette (approved):

PR #281, opened 2024-09-12
TODO to reenable tests on rclcpp for content filtering

ahcorde:


rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1693,6 +1693,8 @@ rmw_subscription_set_content_filter(
   rmw_subscription_t * subscription,
   const rmw_subscription_content_filter_options_t * options)
 {
+  // Reenable test on rclcpp when this feature is implemented
```
Yadunund: Nit
```suggestion
  // Re-enable test in rclcpp when this feature is implemented
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1706,6 +1708,8 @@ rmw_subscription_get_content_filter(
   rcutils_allocator_t * allocator,
   rmw_subscription_content_filter_options_t * options)
 {
+  // Reenable test on rclcpp when this feature is implemented
```
```suggestion
  // Re-enable test in rclcpp when this feature is implemented
```

Yadunund (changes requested):
Fine with the comments in the code but i'd also prefer having an an issue ticket in this repo that keeps track of all the features we need to develop and tests we need to re-enable.

Yadunund (approved):

Issue #170, opened 2024-05-08
bug: unsoundness of `rmw_shutdown`


YuanYuYuan:
We notice that [ApexAI performance test](https://gitlab.com/ApexAI/performance_test) with _rmw_zenoh_ keep failing with the error.

```bash
thread '<unnamed>' panicked at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/thread/local.rs:262:26:
cannot access a Thread Local Storage value during or after destruction: AccessError
stack backtrace:
   0:     0x7e09df3dfab6 - std::backtrace_rs::backtrace::libunwind::trace::h92ba5ac126ecc5ad
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1:     0x7e09df3dfab6 - std::backtrace_rs::backtrace::trace_unsynchronized::hfaa118b6e3a8f31a
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7e09df3dfab6 - std::sys_common::backtrace::_print_fmt::h2075f185adf13bbd
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x7e09df3dfab6 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h9c4bd387f9f3f544
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7e09df40ab20 - core::fmt::rt::Argument::fmt::h018ecebdf268b5bb
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/fmt/rt.rs:142:9
   5:     0x7e09df40ab20 - core::fmt::write::h938c332fdab924eb
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/fmt/mod.rs:1120:17
   6:     0x7e09df3dcc5f - std::io::Write::write_fmt::h4a694b02e44e6363
                               at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/io/mod.rs:1846:15
[...]

PR #283, opened 2024-09-20
Document issue with shutdown

Yadunund:
Address #170.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ In this scenario, the `Context` will be shutdown inside the `Context`'s destruct
 Since the ordering of global/static objects is not defined, this often leads to the above panic.
 
 The recommendation is to ensure the `Context` is shutdown before a program terminates.
-For example, when using `rclcpp`, ensure `rclcpp::shutdown()` is invoked the  program exits.
+One way to ensure this is to call `rclcpp::shutdown()` when the program exits.
+Note that composable nodes should *never* call `rclcpp::shutdown()`, as the composable node container will automatically do this.
 
 For more details, see https://github.com/ros2/rmw_zenoh/issues/170.
--- a/README.md
+++ b/README.md
@@ -131,3 +131,26 @@ For instance:
 - `RUST_LOG=zenoh=debug` activates all the debug logs.
 
 For more information on the `RUST_LOG` syntax, see https://docs.rs/env_logger/latest/env_logger/#enabling-logging.
+
+### Known Issues
+
+### Crash when program terminates
+
+When a program terminates, global and static objects are destructed in the reverse order of their
+construction. The `Thread Local Storage` is one such entity which the `tokio` runtime in Zenoh
+uses. If the Zenoh session is closed after this entity is cleared, it causes a panic like seen below.
+
+```
+thread '<unnamed>' panicked at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/thread/local.rs:262:26:
+cannot access a Thread Local Storage value during or after destruction: AccessError
+```
+
+This can happen with `rmw_zenoh` if a ROS 2 node's `Context` is not shutdown explicitly before the
+program terminates. In this scenario, the `Context` will be shutdown inside the `Context`'s destructor
+which then closes the Zenoh session. Since the ordering of global/static objects is not clear, this
+often leads to the above panic.
+
+The recommendation is to ensure the `Context` is shutdown before a program terminates.
+For example, when using `rclcpp`, ensure `rclcpp::shutdown()` is invoked the  program exits.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,3 +131,26 @@ For instance:
...
+cannot access a Thread Local Storage value during or after destruction: AccessError
+```
+
+This can happen with `rmw_zenoh` if a ROS 2 node's `Context` is not shutdown explicitly before the
+program terminates. In this scenario, the `Context` will be shutdown inside the `Context`'s destructor
+which then closes the Zenoh session. Since the ordering of global/static objects is not clear, this
+often leads to the above panic.
+
+The recommendation is to ensure the `Context` is shutdown before a program terminates.
+For example, when using `rclcpp`, ensure `rclcpp::shutdown()` is invoked the  program exits.
```
clalancette: I think we need to be careful with this advice.

In particular, if you are writing composable nodes, you basically never want to call `rclcpp::shutdown`.  That's because `rclcpp::shutdown` is a global operation, and will shut down your node and all other nodes that happen to be composed with your node at the same time.  That's probably not what is intended.

The composable container *should* call `rclcpp::shutdown` when it is going down, but that is a different thing.

Overall, I'm not saying that you are wrong, but there is subtle distinction here.

Yadunund: Wouldn't composable nodes be created as shared libraries without a `main` function and hence lack the opportunity to invoke `rclcpp::init()` and `rclcpp::shutdown()`? Only executables that implement compile-time composition would include these statements and the advise above should be valid in this scenario?

The default `main` impl for composable nodes includes an `rclcpp::shutdown()` https://github.com/ros2/rclcpp/blob/a78d0cbd33b8fe0b4db25c04f7e10017bfca6061/rclcpp_components/src/node_main.cpp.in#L78

Yadunund: Happy to massage the wording in any other way!

clalancette: > Wouldn't composable nodes be created as shared libraries without a `main` function and hence lack the opportunity to invoke `rclcpp::init()` and `rclcpp::shutdown()`?

I mean, they shouldn't call it, but that doesn't mean they won't.  `rclcpp::shutdown`, in particular, is often (erroneously) called when a node is done with its work; even our demos are guilty of this: https://github.com/ros2/demos/blob/0f2afe53be38b71c01d43f0900e18187d2e36082/demo_nodes_cpp/src/services/add_two_ints_client_async.cpp#L65

I'm going to suggest some wording to point out that composable nodes shouldn't generally do this.

README.md
```diff
@@ -131,3 +131,26 @@ For instance:
...
 
 For more information on the `RUST_LOG` syntax, see https://docs.rs/env_logger/latest/env_logger/#enabling-logging.
+
+### Known Issues
+
+### Crash when program terminates
+
+When a program terminates, global and static objects are destructed in the reverse order of their
+construction. The `Thread Local Storage` is one such entity which the `tokio` runtime in Zenoh
+uses. If the Zenoh session is closed after this entity is cleared, it causes a panic like seen below.
```
clalancette:
```suggestion
When a program terminates, global and static objects are destructed in the reverse order of their
construction.
The `Thread Local Storage` is one such entity which the `tokio` runtime in Zenoh uses.
If the Zenoh session is closed after this entity is cleared, it causes a panic like seen below.
```

README.md
```diff
@@ -131,3 +131,26 @@ For instance:
...
+
+```
+thread '<unnamed>' panicked at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/thread/local.rs:262:26:
+cannot access a Thread Local Storage value during or after destruction: AccessError
+```
+
+This can happen with `rmw_zenoh` if a ROS 2 node's `Context` is not shutdown explicitly before the
+program terminates. In this scenario, the `Context` will be shutdown inside the `Context`'s destructor
+which then closes the Zenoh session. Since the ordering of global/static objects is not clear, this
+often leads to the above panic.
```
clalancette:
```suggestion
This can happen with `rmw_zenoh` if the ROS 2 `Context` is not shutdown explicitly before the
program terminates.
In this scenario, the `Context` will be shutdown inside the `Context`'s destructor which then closes the Zenoh session.
Since the ordering of global/static objects is not defined, this often leads to the above panic.
```

README.md
clalancette:
```suggestion
One way to ensure this is to call `rclcpp::shutdown()` when the program exits.
Note that composable nodes should *never* call `rclcpp::shutdown()`, as the composable node container will automatically do this.
```

clalancette (changes requested):

clalancette (approved):

PR #288, opened 2024-10-02
Thread-safe access to SubscriptionData

Yadunund:
This PR introduces SubscriptionData class that manages the lifetime of Zenoh artifacts. It makes access to class members thread-safe where possible. Some z_closure callbacks still work with type erased raw ptrs but this will be addressed when we migrate to zenoh-cpp.

It seems like it is not feasible to store rmw_node_t * as type erased void * in rmw_subscription_t->data  like we did with rmw_publisher_t->data .

The problem is that because rmw_subscriptions_t in rmw_wait contains an array of the type erased subscriptions handles (ie, rmw_subscription_t->data ), there is no way to get the SubscriptionData::SharedPtr here https://github.com/ros2/rmw_zenoh/blob/cf8e8a2eccdf2d4bcce6858d35ed1eac14ce0a0a/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L3220-L3221. ie we cannot traverse the heirarchy from rmw_subscription->data->rmw_node_t->context->impl->get_node_data()->get_sub_data().

The only option we have is storing a rawptr to SubscriptionData in rmw_subscription_t->data but leads to the risk of it becoming a dangling pointer when the SubscriptionData::SharedPtr is deleted.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
@@ -237,12 +237,12 @@ std::string zid_to_str(const z_id_t & id);
 }  // namespace liveliness
 
 ///=============================================================================
-/// Helper function to generate a randon GID.
+/// Helper function to generate a random GID.
 void generate_random_gid(uint8_t gid[RMW_GID_STORAGE_SIZE]);
 
 ///=============================================================================
 /// Generate a hash for a given GID.
-size_t hash_gid(const uint8_t * gid);
+size_t hash_gid(const uint8_t gid[RMW_GID_STORAGE_SIZE]);
 }  // namespace rmw_zenoh_cpp
 
 ///=============================================================================
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
@@ -237,8 +237,12 @@ std::string zid_to_str(const z_id_t & id);
 }  // namespace liveliness
 
 ///=============================================================================
-// Helper function to generate a randon GID.
+/// Helper function to generate a randon GID.
 void generate_random_gid(uint8_t gid[RMW_GID_STORAGE_SIZE]);
+
+///=============================================================================
+/// Generate a hash for a given GID.
+size_t hash_gid(const uint8_t * gid);
 }  // namespace rmw_zenoh_cpp
 
 ///=============================================================================
--- a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
@@ -291,6 +291,7 @@ std::shared_ptr<SubscriptionData> SubscriptionData::make(
       return nullptr;
     }
   }
+  sub_data->graph_cache_ = graph_cache;
 
   auto undeclare_z_sub = rcpputils::make_scope_exit(
     [data = sub_data]() {
@@ -612,4 +613,12 @@ void SubscriptionData::set_on_new_message_callback(
   std::lock_guard<std::mutex> lock(mutex_);
   data_callback_mgr_.set_callback(user_data, callback);
 }
+
+//==============================================================================
+std::shared_ptr<GraphCache> SubscriptionData::graph_cache() const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  return graph_cache_;
+}
+
 }  // namespace rmw_zenoh_cpp
--- a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
@@ -128,8 +128,6 @@ std::shared_ptr<SubscriptionData> SubscriptionData::make(
   const rosidl_message_type_support_t * type_support,
   const rmw_qos_profile_t * qos_profile)
 {
-  auto sub_data = std::shared_ptr<SubscriptionData>(new SubscriptionData{});
-  sub_data->rmw_node_ = node;
   rmw_qos_profile_t adapted_qos_profile = *qos_profile;
   rmw_ret_t ret = QoS::get().best_available_qos(
     node, topic_name.c_str(), &adapted_qos_profile, rmw_get_publishers_info_by_topic);
@@ -140,9 +138,8 @@ std::shared_ptr<SubscriptionData> SubscriptionData::make(
   rcutils_allocator_t * allocator = &node->context->options.allocator;
 
   const rosidl_type_hash_t * type_hash = type_support->get_type_hash_func(type_support);
-  sub_data->type_support_impl_ = type_support->data;
   auto callbacks = static_cast<const message_type_support_callbacks_t *>(type_support->data);
-  sub_data->type_support_ = std::make_unique<MessageTypeSupport>(callbacks);
+  auto message_type_support = std::make_unique<MessageTypeSupport>(callbacks);
 
   // Convert the type hash to a string so that it can be included in
   // the keyexpr.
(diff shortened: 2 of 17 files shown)


rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -237,8 +237,12 @@ std::string zid_to_str(const z_id_t & id);
 }  // namespace liveliness
 
 ///=============================================================================
-// Helper function to generate a randon GID.
+/// Helper function to generate a randon GID.
```
clalancette:
```suggestion
/// Helper function to generate a random GID.
```

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -237,8 +237,12 @@ std::string zid_to_str(const z_id_t & id);
...
 void generate_random_gid(uint8_t gid[RMW_GID_STORAGE_SIZE]);
+
+///=============================================================================
+/// Generate a hash for a given GID.
+size_t hash_gid(const uint8_t * gid);
```
clalancette: I know you are just copying this over, but I'm going to suggest that we change this signature to:

```suggestion
size_t hash_gid(const uint8_t gid[RMW_GID_STORAGE_SIZE]);
```

It won't make much of a practical difference, but it will be clear to the reader exactly what the size of gid is.

rmw_zenoh_cpp/src/rmw_event.cpp
```diff
@@ -69,12 +69,14 @@ rmw_publisher_event_init(
   rmw_event->event_type = event_type;
 
   // Register the event with graph cache.
+  std::weak_ptr<rmw_zenoh_cpp::PublisherData> data_wp = pub_data;
```
clalancette, line 72: Is this a separate bugfix?

rmw_zenoh_cpp/src/rmw_event.cpp
```diff
@@ -131,16 +128,18 @@ rmw_subscription_event_init(
...
-  context_impl->graph_cache()->set_qos_event_callback(
-    sub_data->entity->guid(),
+  // std::weak_ptr<rmw_zenoh_cpp::SubscriptionData> data_wp = sub_data;
+  sub_data->graph_cache()->set_qos_event_callback(
+    sub_data->guid(),
     zenoh_event_type,
     [sub_data,
     zenoh_event_type](std::unique_ptr<rmw_zenoh_cpp::rmw_zenoh_event_status_t> zenoh_event)
     {
+      // auto sub_data = data_wp.lock();
```
clalancette:
```suggestion
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -928,65 +951,36 @@ rmw_create_subscription(
...
-    });
-
-  sub_data->context = node->context;
-
+  // Store type erased node in rmw_subscription->data so that the
+  // Subscription can be safely accessed.
+  // TODO(Yadunund): We cannot store the rmw_node_t * here since this type erased
+  // subscription handle will be returned in the rmw_subscriptions_t in rmw_wait
+  // from which we cannot obtain SubscriptionData.
+  // rmw_subscription->data = reinterpret_cast<void *>(const_cast<rmw_node_t *>(node));
```
```suggestion
  // TODO(Yadunund): We cannot store the rmw_node_t * here since this type erased
  // subscription handle will be returned in the rmw_subscriptions_t in rmw_wait
  // from which we cannot obtain SubscriptionData.
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -3063,11 +2757,8 @@ check_and_attach_condition(
 
   if (subscriptions) {
     for (size_t i = 0; i < subscriptions->subscriber_count; ++i) {
-      auto sub_data =
-        static_cast<rmw_zenoh_cpp::rmw_subscription_data_t *>(subscriptions->subscribers[i]);
-      if (sub_data == nullptr) {
-        continue;
-      }
+      rmw_zenoh_cpp::SubscriptionData * sub_data =
+        static_cast<rmw_zenoh_cpp::SubscriptionData *>(subscriptions->subscribers[i]);
```
clalancette, line 2758: I think we still want to check whether `sub_data` is NULL before doing anything:

```suggestion
      rmw_zenoh_cpp::SubscriptionData * sub_data =
        static_cast<rmw_zenoh_cpp::SubscriptionData *>(subscriptions->subscribers[i]);
      if (sub_data == nullptr) {
        continue;
      }
```

clalancette (changes requested):
This generally looks good.  I've left some things to improved, and it needs to be rebased.  Once those are done, I'll run some more extensive tests on it.

rmw_zenoh_cpp/src/rmw_event.cpp
Yadunund: The approach with capturing the `weak_ptr` ensures that if the event callback executes for some reason after the last reference to `PublisherData::SharedPtr` goes out of scope, we would safely exit the callback. This is what I want to do with some of the Zenoh callbacks as well once we switch to `zenoh-cpp`.
I would have liked to also do it within `rmw_subscription_event_init` but given that we're storing the `raw_ptr` to `SubscriptionData` in `rmw_subscription_t->data` and hence not able to retrieve the shared_ptr to the same, we are stuck with potential to dereference a dangling pointer if the scenario above. But this scenario should never occur given the implementation of `rcl*`

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -0,0 +1,618 @@
...
+      "Unable to obtain SubscriptionData from data for %s.",
+      z_loan(keystr)
+    );
+    return;
+  }
+
+  uint8_t pub_gid[RMW_GID_STORAGE_SIZE];
+  if (!get_gid_from_attachment(&sample->attachment, pub_gid)) {
+    // We failed to get the GID from the attachment.  While this isn't fatal,
+    // it is unusual and so we should report it.
```
ahcorde:
```suggestion
    // We failed to get the GID from the attachment. While this isn't fatal,
    // it is unusual and so we should report it.
```

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -0,0 +1,618 @@
...
+    RMW_SET_ERROR_MSG("Failed to allocate type_hash_c_str.");
+    return nullptr;
+  }
+  auto free_type_hash_c_str = rcpputils::make_scope_exit(
+    [&allocator, &type_hash_c_str]() {
+      allocator->deallocate(type_hash_c_str, allocator->state);
+    });
+
+  // Everything above succeeded and is setup properly.  Now declare a subscriber
+  // with Zenoh; after this, callbacks may come in at any time.
```
ahcorde:
```suggestion
  // Everything above succeeded and is setup properly. Now declare a subscriber
  // with Zenoh; after this, callbacks may come in at any time.
```

ahcorde:
[ros2.repos](https://gist.github.com/ahcorde/be917be95bf72bc1107d398d42fb264e)

# rmw_zenoh/yadu/raii-sub
* rmw_zenoh commit: https://github.com/ros2/rmw_zenoh/commit/f8ea0c330168bfa8df733e8af8558a6cd208b163

| rcl      | None       | None   |
| rcl_action   | test_graph__rmw_zenoh_cpp | segfault |
| rclcpp   | test_wait_set        | Failed (Check logs)      |
| rclcpp_action   | test_server       | Failed (Check logs)      |

------

There are some new segfaults 
 - `rcl_action` -> `test_graph__rmw_zenoh_cpp`
 - `rclcpp_Action` -> test_server

clalancette:
This generally looks good to me, and passes my tests locally.  We should fix the failing `rcl_action` and `rclcpp_action` tests before merging, though.

Yadunund:
@ahcorde thanks a lot for running those tests ahead of time and reporting the issues. I've fixed them in https://github.com/ros2/rmw_zenoh/pull/288/commits/2bd7e13631a8db2546523fa12522c9026f6cab71.

clalancette (approved):

ahcorde (approved):

Issue #286, opened 2024-09-26
Test failures with rmw_zenoh


Yadunund:
This ticket will track the status of test failures with `rmw_zenoh`.

Below is the template to be used with an example. Let's copy paste this each time we make changes to the `ros2.repos` file from 
https://github.com/Yadunund/ros2/blob/yadu/rmw_zenoh_testing/ros2.repos

## Test setup
* Date: 26-09-2024
* ros2.repos file:  https://github.com/Yadunund/ros2/blob/c425fa9d0ced5bfca6e530141fc1f1629be7e720/ros2.repos

## rmw_zenoh/rolling
* rmw_zenoh commit: b3e4a6378063d4f73def3069a1f23c9e41fd4d89

| ROS Package | Failing test name | Failing test output |
|-------------|-------------------|----------------------------|
| rclcpp           |  33 - test_intra_process_manager |                        [gist](https://gist.github.com/Yadunund/0453d07786bb8b6d0b9b2155f63f241d) |
|                      |                                |                                                |
|                      |                                |                                                |

## rmw_zenoh/dev/1.0.0
* rmw_zenoh commit: 6a48919727226ebad853373e8a1eec0ee22c3be4

| ROS Package | Failing test name | Failing test output |
|-------------|-------------------|----------------------------|
|                      |                                |                                                |
|                      |                                |                                                |
|                      |                                |                                                |

Additional tips
* Individual tests can be run with a command like `colcon test --event-handlers console_direct+ --packages-select rcl --ctest-args -R test_graph`

ahcorde:
## Test setup
* Date: 30-09-2024
* ros2.repos file:  https://gist.github.com/ahcorde/d5b6d86762badd54cfc5021b01705dc8


## rmw_zenoh/rolling
* rmw_zenoh commit: 67ed661cb9af460b0c4e0aa0b5f6e2e163b6a82c

| ROS Package | Failing test name | Failing test output |
|-------------|-------------------|----------------------------|
|      rcl     | None  |                         |
|      rcl_action     | None  |                         |
|        rcl_lifecyle              |                 None               |                                                |
|        rcl_yaml_param_parser              |               None                 |                                                |
| rclcpp| test_intra_process_manager | description below |
| rclcpp| test_node_interfaces__node_graph | description below |
| rclcpp| test_publisher | description below |
| rclcpp| test_qos | description below |
| rclcpp| test_events_executor | Timeout |
| rclcpp| test_wait_set | description below |
[...]

PR #289, opened 2024-10-02
Initialize PublsiherData only after all members are valid

Yadunund:
Fix regression in `rcl_lifecycle` as per https://github.com/ros2/rmw_zenoh/issues/286#issuecomment-2389519210

> Note: we already do the same for NodeData


ahcorde (approved):

clalancette (commented):
After it is already merged, but this looks good to me.  Thanks @Yadunund !

Issue #290, opened 2024-10-03
get_entities_info_by_topic is not setting gid


ahcorde:
get_entities_info_by_topic is not setting gid. It's setting `name`, `namespace`, `endpoint type`, `qos` and `topic_type_hash`. 

We need to set the gid too wit `rmw_topic_endpoint_info_set_gid`

There is a [TODO](https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/graph_cache.cpp#L1211) in the code, this will fix `test_node_interfaces__node_graph` in rclcpp

Yadunund:
Summarizing the state of things before describing how to attack this problem
- Every node/pub/sub/client/service entity publishes a [liveliness token](https://github.com/ros2/rmw_zenoh/blob/8ba1c916fa1563d40eccb9a3b2137d4b78860927/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp#L170-L174) which is received by all [contexts](https://github.com/ros2/rmw_zenoh/blob/8ba1c916fa1563d40eccb9a3b2137d4b78860927/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp#L136) in the ROS graph. The subscription callback of these tokens populate the `GraphCache`.
- Within the `GraphCache`, various `Entities` are reconstructed purely from liveliness tokens using [this API](https://github.com/ros2/rmw_zenoh/blob/8ba1c916fa1563d40eccb9a3b2137d4b78860927/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp#L137-L138.)
- The liveliness token only includes [these components](https://github.com/ros2/rmw_zenoh/blob/8ba1c916fa1563d40eccb9a3b2137d4b78860927/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L97-L112). GID is not included.
- We uniquely identify `Entities` in the `GraphCache` using a `guid` that is just the [hash of the liveliness token](https://github.com/ros2/rmw_zenoh/blob/8ba1c916fa1563d40eccb9a3b2137d4b78860927/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L444)
[...]

Issue #148, opened 2024-04-04
Align on Transport GID and Liveliness GUID


Yadunund:
At present, entities like pub/subs are assigned an `rmw_gid_t gid` value and this is attached to the payload. However, GraphCache relies on a `std::size guid` to unique identify entities. The `size_t` is a `std::hash` of the liveliness token which is now unique.

It would be best if we could combine the two somehow.
1) Either derive the 16byte GID from the liveliness token or
2) Append the 16byte GID to the liveliness token.

Personally I'd prefer 1) to since the liveliness tokens are already several bytes.

PR #291, opened 2024-10-04
Align on Transport GID and Liveliness GUID

ahcorde:
Related with https://github.com/ros2/rmw_zenoh/issues/290 and https://github.com/ros2/rmw_zenoh/issues/148

This PR makes `test_node_interfaces__node_graph` from rclcpp to pass.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -1208,7 +1208,28 @@ rmw_ret_t GraphCache::get_entities_info_by_topic(
           }
         }
 
-        // TODO(Yadunund): Set gid.
+        liveliness::ConstEntityPtr entity;
+        if (entity_type == EntityType::Publisher)
+        {
+          entity = (*topic_data->pubs_.begin());
+        } else {
+          entity = (*topic_data->subs_.begin());
+        }
+
+        if (entity)
+        {
+          uint8_t gid[RMW_GID_STORAGE_SIZE];
+          entity->get_gid(gid);
+
+          ret = rmw_topic_endpoint_info_set_gid(
+            &endpoint_info,
+            gid,
+            RMW_GID_STORAGE_SIZE
+          );
+          if (RMW_RET_OK != ret) {
+            return ret;
+          }
+        }
       }
     }
   }
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -1209,15 +1209,13 @@ rmw_ret_t GraphCache::get_entities_info_by_topic(
         }
 
         liveliness::ConstEntityPtr entity;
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -1208,7 +1208,26 @@ rmw_ret_t GraphCache::get_entities_info_by_topic(
...
+        liveliness::ConstEntityPtr entity;
+        if (entity_type == EntityType::Publisher) {
+          entity = (*topic_data->pubs_.begin());
+        } else {
+          entity = (*topic_data->subs_.begin());
+        }
+
+        if (entity) {
+          uint8_t gid[RMW_GID_STORAGE_SIZE];
+          entity->get_gid(gid);
```
Yadunund, line 1220: Should we iterate though all the `Entity` ptrs in the `EnittySet`?
```
const auto & entity_set = entity_type == EntityType::Publisher? topic_data->pubs_ : topic_data->subs_;
for (auto entity_it = entity_set.begin(); entity_it != entity_set.end(); ++entity_it) {
   uint8_t gid[RMW_GID_STORAGE_SIZE];
   entity->get_gid(gid);
    ret = rmw_topic_endpoint_info_set_gid(
        &endpoint_info,
        gid,
        RMW_GID_STORAGE_SIZE
     );
     if (RMW_RET_OK != ret) {
        return ret;
          }
}

```

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -153,6 +153,12 @@ class Entity
   // This is named guid and not gid to remain distinct as it is not of type rmw_gid_t.
   std::size_t guid() const;
 
+  // Generate the 16byte GID from the keyexpr guid
+  // The size of guid is std::size_t. It can vary depending on the platform and architecture.
+  // On most 32-bit systems, it is usually 32 bits, while on 64-bit systems,
+  // it is typically 64 bits. It fit in the RMW_GID_STORAGE_SIZE.
+  void get_gid(uint8_t gid[RMW_GID_STORAGE_SIZE]) const;
```
Yadunund, line 160: Nit: To be consistent with naming of other getters.
```suggestion
  void gid(uint8_t gid[RMW_GID_STORAGE_SIZE]) const;
```

Yadunund (changes requested):

rmw_zenoh_cpp/src/detail/graph_cache.cpp
ahcorde: If we iterate over the list, then it will override the gid value (if there is more than one).

clalancette:
I'm a little later here, but please see https://github.com/ros2/rmw_zenoh/issues/290#issuecomment-2397193935 for some of my thoughts.

clalancette:
Closing in favor of #298 .

PR #293, opened 2024-10-09
Make ClientData thread-safe

Yadunund:
Make access to `ClientData` thread-safe.

~~I've not fully tested this PR yet but I know `colcon test --event-handlers console_direct+ --packages-select rclcpp_action --ctest-args -R test_server` fails. It has to do with how i'm handling cases where the client is shutdown when there are still queries in flight.~~


ahcorde (changes requested):
fix conflicts

Yadunund:
@ahcorde could you help check if the expected tests are passing here?

ahcorde:
[ros2.repos](https://gist.github.com/ahcorde/6f970ad7270f43c85d43c3444adf27f9)

# rmw_zenoh/yadu/raii-service
* rmw_zenoh commit: https://github.com/ros2/rmw_zenoh/commit/aaa01c187ab90c01d9dd55b89818f731846c1dc1

| rcl | None | None |
| rcl_action | None | None |
| rcl_yaml_param_parser   | None        | None      |
| test_cli_remap   | test_cli_remapping        | TBD |
| rclpy   | None | |

ahcorde (approved):

clalancette:
Because of #298, this now needs a rebase and conflicts resolved.  Once that is done, I'll do another review pass over it.

Yadunund:
@clalancette i've resolved the conflicts!

rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -1182,15 +1181,15 @@ rmw_ret_t GraphCache::get_entities_info_by_topic(
 
 ///=============================================================================
 rmw_ret_t GraphCache::service_server_is_available(
-  const char * service_name,
-  const char * service_type,
+  const liveliness::TopicInfo & client_topic_info,
```
clalancette, line 1184: Technically, we should `#include "liveliness_utils.hpp"` for this class.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1405,245 +1411,69 @@ rmw_create_client(
...
   }
 
-  rcutils_allocator_t * allocator = &node->context->options.allocator;
-
-  // Validate service name
-  int validation_result;
-
-  if (rmw_validate_full_topic_name(service_name, &validation_result, nullptr) != RMW_RET_OK) {
-    RMW_SET_ERROR_MSG("rmw_validate_full_topic_name failed");
+  // Get the RMW the type support.
```
```suggestion
  // Get the service type support.
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1405,245 +1411,69 @@ rmw_create_client(
...
+      rmw_client,
+      context_impl->session(),
+      context_impl->get_next_entity_id(),
+      service_name,
+      type_support,
+      qos_profile))
+  {
+    // Error already handled.
     return nullptr;
   }
```
clalancette, line 1451: If a step below fails, do we need to remove this client from the node?  In other words, do we need a `rcpputils::scope_exit` block here?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1405,245 +1411,69 @@ rmw_create_client(
...
+  // TODO(Yadunund): We cannot store the rmw_node_t * here since this type erased
+  // Client handle will be returned in the rmw_clients_t in rmw_wait
+  // from which we cannot obtain ClientData.
+  rmw_client->data = static_cast<void *>(node_data->get_client_data(rmw_client).get());
   rmw_client->implementation_identifier = rmw_zenoh_cpp::rmw_zenoh_identifier;
   rmw_client->service_name = rcutils_strdup(service_name, *allocator);
   RMW_CHECK_FOR_NULL_WITH_MSG(
     rmw_client->service_name,
-    "failed to allocate client name",
+    "failed to allocate service name",
```
```suggestion
    "failed to allocate service client name",
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2866,29 +2522,16 @@ rmw_service_server_is_available(
...
+  rmw_zenoh_cpp::ClientData * client_data =
+    static_cast<rmw_zenoh_cpp::ClientData *>(client->data);
   if (client_data == nullptr) {
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "Unable to retreive client_data from client for service %s", client->service_name);
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  std::string service_type = client_data->request_type_support->get_name();
-  size_t suffix_substring_position = service_type.find("Request_");
```
clalancette, line 2874: I notice that we are no longer checking whether the service_type has `Request_` in it.  Do we need to add that back in?

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -0,0 +1,572 @@
...
+#include "rmw/get_topic_endpoint_info.h"
+#include "rmw/impl/cpp/macros.hpp"
+
+namespace rmw_zenoh_cpp
+{
+// A static mutex whose lifetime will extend beyond that of the internal mutex_.
+// This is needed as the decrement_queries_in_flight callback that me be executed
+// after ClientData is deallocated will try to lock an invalid mutex and thus lead
+// to UB. This only happens in rclcpp_action/test_client.
+static std::mutex num_in_flight_mutex;
```
clalancette: I'm not sure I understand how this fixes the problem, and I'm not sure it is required anymore.  If `client_data` has gone out of scope, then calling `client_data->decrement_queries_in_flight()` is already UB, never mind accessing the `std::mutex` within it (it will usually happen to work, but that's besides the point).

But we really shouldn't get into this situation anymore anyway, because `NodeData::delete_client_data` will only be deleting the structure if there is nothing in flight.  (The potential memory leak is worse here, in that `client_data_drop` isn't cleaning up anymore if this is the "last" thing in flight, but that is not UB).

In other words, I don't see why we can't embed this mutex in the `ClientData` structure.  But there may be something I'm missing.

clalancette (changes requested):

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: Ah nope! The shared_ptr we created within `create_client_data` will go out of scope and the destructor will clean everything up. The `free_rmw_client` block will take care of deallocating the `rmw_client_t` object.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
Yadunund: I don't think I implemented this correctly.

My approach here was to avoid UBs only since memory leaks should never happen given the use of shared_ptrs.

If `client_data` is deallocated by resetting an `rclcpp::Node` ptr (as done in one of the tests), dereferencing `client_data` within `client_data_drop` like `client_data-> decrement_queries_in_flight()` would be UB. In this scenario, even though we do not remove the `client_data` from `NodeData::clients_` when `NodeData::delete_client_data()` is called, it will get removed when the `NodeData`  is deleted by the reset call. 

Ideally we have a static map of ClientData rawptr address to num_in_flight. Then `client_data_drop` would lookup this map and decrement the count. We would increment the count in the same map.

But I don't know if it worth the effort. Please let me know what you think.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Yadunund: So this check was moved inside `ClientData::make()` but there was typo there anyways which I have fixed https://github.com/ros2/rmw_zenoh/pull/293/commits/4c9b1069629bfd2daea1f3e14ef5376dfca3f356. It would not have caused any problems since the substring we extract will be the same but now it's semantically correct.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
Yadunund: Alright I think I have a better solution implemented in https://github.com/ros2/rmw_zenoh/pull/293/commits/37e2e0372cec18665ef2fe1307a95646735711bc

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -0,0 +1,613 @@
...
+  if (ret != RMW_RET_OK) {
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Error destructing client /%s.",
+      entity_->topic_info().value().name_.c_str()
+    );
+  }
+  std::lock_guard<std::mutex> lock(num_in_flight_mutex);
+  num_in_flight_map.erase(this);
+  deleted_clients.insert(this);
```
clalancette: Importantly, I think we only need to add this to the `deleted_clients` map if there is anything in flight.  I think that should keep us from "leaking" memory in the `deleted_clients` map for every deleted client.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+  // See the comment about the "num_in_flight" class variable in the ClientData class for
+  // why we need to do this.
+  // Note: This called could lead to UB since the ClientData * could have been deallocated
+  // if the query reply is received after NodeData was destructed. ie even though we do not
+  // erase the ClientData from NodeData's clients_ map when rmw_destroy_client is called,
+  // the clients_ maps would get erased when the NodeData's destructor is invoked.
+  // This is an edge case that should be resolved once we switch to zenoh-cpp and capture
+  // weaK_ptr<ClientData> in this callback instead.
+  auto num_in_flight_it = num_in_flight_map.find(client_data);
```
clalancette: This needs to take the `num_in_flight_mutex`.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -0,0 +1,613 @@
...
+      "Unable to create liveliness token for the client.");
+    return nullptr;
+  }
+
+  free_ros_keyexpr.cancel();
+  free_token.cancel();
+
+  // Erase from deleted_clients set if the memory address for the client data is reused.
+  num_in_flight_map.erase(client_data.get());
+  deleted_clients.erase(client_data.get());
```
clalancette: There are a few different problems with this that I see.

First, this needs to take the `num_in_flight_mutex` for mutual exclusion.

Second, I don't think that we ever need to erase from the `num_in_flight_map`.  That only exists for the lifetime of a `ClientData` object.  In point of fact, we may actually want to add this object to the map right here (with a `num_in_flight` of 0).

Finally, I don't think that erasing this from the `deleted_clients` is the right thing to do.  If we take my other suggestion on only adding to the `deleted_clients` map if queries are in-flight, then removing this here means that those outstanding queries may do the wrong thing during `client_data_handler`.  That is, they will see that they are *not* in the `deleted_clients` map, and go on to operate on a different object.  This is actually tricky to fix, though I have a somewhat hacky idea.  That idea is to continually allocate `ClientData` structures (as in line 236) until we find an address that is *not* in `deleted_clients`.  Once we find one, we deallocate all of the other ones.  That way we'll be guaranteed to get an address that does not have queries currently in flight.

clalancette (changes requested):
Overall, I like the new direction here.  I have some specific concerns on some of the data handling inline.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
Yadunund: > That idea is to continually allocate ClientData structures (as in line 236) until we find an address that is not in deleted_clients. Once we find one, we deallocate all of the other ones. That way we'll be guaranteed to get an address that does not have queries currently in flight.

That should resolve it but wondering we need to go through this hassle now especially since we will undo all these changes once switch to zenoh-cpp and will rely on lambda captures?

clalancette: Personally, I think it is worth it.  It's not a whole lot of code, and we never know what is going to happen in the future.  The good news is that it is not too hard; see #304, which targets this PR and fixes all of these issues (I think).

Yadunund:
nav2 demo crashes when the following steps are performed
1. ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False
2. Localize the robot and wait for startup to complete
3. Click `Reset` from the nav2 panel in RViz

```
[component_container_isolated-8] malloc(): unsorted double linked list corrupted
[component_container_isolated-8] Magick: abort due to signal 6 (SIGABRT) "Abort"...
[ERROR] [component_container_isolated-8]: process has died [pid 2257204, exit code -6, cmd '/opt/ros/rolling/lib/rclcpp_components/component_container_isolated --ros-args --log-level info --ros-args -r __node:=nav2_container --params-file /tmp/launch_params_xwa4_atv --params-file /tmp/launch_params_3tgew0_b -r /tf:=tf -r /tf_static:=tf_static'].
```

Yadunund:
It seems like https://github.com/ros2/rmw_zenoh/pull/293/commits/082e64c42e5ee09c0f17b667001e306b4ff42e20 introduced this behavior as I can't reproduce the crash if I drop that commit.

ahcorde (changes requested):
conflicts

clalancette:
> conflicts

I've now rebased it onto the latest.

clalancette:
Besides rebasing this, I also did a local test where I rebased this onto https://github.com/ros2/rmw_zenoh/pull/307 , and also reverted 19b84e98771b0ec07c16959e1d746c51fdab4fc6 .  But I am running into shutdown issues in `client_data_handler` where it is deadlocking, so I'll need to look into that further.

clalancette:
All right, I've just rebased this and also pushed another commit to vastly simplify our handling of in-flight data.  I think this should work a lot more reliably now.  With that, I'm going to approve and merge this.  Thanks for the long slog on this one.

clalancette (approved):

PR #276, opened 2024-09-05
Zenoh 1.0.0 Porting

YuanYuYuan:
### Notes

- Remove some unused dependencies.
- Use `z_view_keyexpr_t` rather `z_owned_keyexpr_t` if possible. This keeps the reference only and doesn't need to be dropped.
- The attachment type has been removed. Use the plain z_bytes instead.
- Attachment is nothing but z_bytes. Use an iterator to get value by key.
- `z_check` is now internal. We check zenoh entities' health by the creation functions' return.

PR #294, opened 2024-10-10
Add tracing instrumentation using tracetools

christophebedard:
With this instrumentation, `rmw_zenoh_cpp` is supported out-of-the-box by tools that process ROS 2 trace data, such as Eclipse Trace Compass.

It's pretty straightforward except for a few things:

1. Getting the `rmw_publisher_t *` for the `rmw_publish` tracepoint and `rmw_service_t *` for the `rmw_send_response` tracepoint
    - Add them to `PublisherData` and `ServiceData`, respectively.
2. Getting the source timestamp for the `rmw_publish` and `rmw_send_response` tracepoints
    - Getting the source timestamp of a message being published is needed for tracking messages from publisher to subscription or replies from service to client.
    - Since https://github.com/ros2/ros2_tracing/pull/74, we're getting the source timestamp for a published message from `rmw`. This means we don't need to instrument the underlying middleware itself, which we used to have to do.
    - Remove `create_map_and_set_sequence_num()`, since it's already pretty simple and it makes extracting the source timestamp easier.
3. Getting the/a subscription `rmw` GID for the `rmw_subscription_init` tracepoint
    - See the comment before the tracepoint.

I'm open to suggestions.

---

action-ros-ci-repos-override: https://gist.githubusercontent.com/christophebedard/18561011c03603652ad49701412e5f41/raw/9e9d0bc9bbf0c36b7d5419bc5797ef17aa90d6fa/ros2.repos

### DIFF
--- a/rmw_zenoh_cpp/src/detail/zenoh_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/zenoh_utils.cpp
@@ -27,23 +27,6 @@
 namespace rmw_zenoh_cpp
 {
 ///=============================================================================
-zenoh::Bytes create_map_and_set_sequence_num(
-  int64_t sequence_number,
-  std::array<uint8_t, RMW_GID_STORAGE_SIZE> gid,
-  int64_t * source_timestamp)
-{
-  auto now = std::chrono::system_clock::now().time_since_epoch();
-  auto now_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(now);
-  int64_t timestamp = now_ns.count();
-  if (nullptr != source_timestamp) {
-    *source_timestamp = timestamp;
-  }
-
-  rmw_zenoh_cpp::AttachmentData data(sequence_number, timestamp, gid);
-  return data.serialize_to_zbytes();
-}
-
-///=============================================================================
 ZenohQuery::ZenohQuery(
   const zenoh::Query & query,
   std::chrono::nanoseconds::rep received_timestamp)
@@ -87,4 +70,11 @@ std::chrono::nanoseconds::rep ZenohReply::get_received_timestamp() const
 {
   return received_timestamp_;
 }
+
+int64_t get_system_time_in_ns()
+{
+  auto now = std::chrono::system_clock::now().time_since_epoch();
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
+}
+
 }  // namespace rmw_zenoh_cpp
--- a/rmw_zenoh_cpp/src/detail/zenoh_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/zenoh_utils.cpp
@@ -28,13 +28,18 @@ namespace rmw_zenoh_cpp
 {
 ///=============================================================================
 zenoh::Bytes create_map_and_set_sequence_num(
-  int64_t sequence_number, std::array<uint8_t, RMW_GID_STORAGE_SIZE> gid)
+  int64_t sequence_number,
+  std::array<uint8_t, RMW_GID_STORAGE_SIZE> gid,
+  int64_t * source_timestamp)
 {
   auto now = std::chrono::system_clock::now().time_since_epoch();
   auto now_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(now);
-  int64_t source_timestamp = now_ns.count();
+  int64_t timestamp = now_ns.count();
+  if (nullptr != source_timestamp) {
+    *source_timestamp = timestamp;
+  }
 
-  rmw_zenoh_cpp::AttachmentData data(sequence_number, source_timestamp, gid);
+  rmw_zenoh_cpp::AttachmentData data(sequence_number, timestamp, gid);
   return data.serialize_to_zbytes();
 }
 
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -56,6 +56,8 @@
 #include "rmw/validate_namespace.h"
 #include "rmw/validate_node_name.h"
 
+#include "tracetools/tracetools.h"
+
 namespace
 {
 //==============================================================================
@@ -454,6 +456,14 @@ rmw_create_publisher(
   free_topic_name.cancel();
   free_rmw_publisher.cancel();
 
+  if (TRACETOOLS_TRACEPOINT_ENABLED(rmw_publisher_init)) {
+    rmw_gid_t gid{};
+    // Trigger tracepoint even if we cannot get the GID
+    rmw_ret_t gid_ret = rmw_get_gid_for_publisher(rmw_publisher, &gid);
+    static_cast<void>(gid_ret);
+    TRACETOOLS_DO_TRACEPOINT(
+      rmw_publisher_init, static_cast<const void *>(rmw_publisher), gid.data);
+  }
   return rmw_publisher;
 }
 
@@ -990,6 +1000,11 @@ rmw_create_subscription(
   free_topic_name.cancel();
   free_rmw_subscription.cancel();
 
+  // rmw does not require GIDs for subscriptions, and GIDs in rmw_zenoh are not based on any ID of
+  // the underlying zenoh objects, so there is no need to collect a GID here
+  rmw_gid_t gid{};
+  TRACETOOLS_TRACEPOINT(
+    rmw_subscription_init, static_cast<const void *>(rmw_subscription), gid.data);
   return rmw_subscription;
 }
 
@@ -1134,7 +1149,18 @@ rmw_take(
     static_cast<rmw_zenoh_cpp::SubscriptionData *>(subscription->data);
   RMW_CHECK_ARGUMENT_FOR_NULL(sub_data, RMW_RET_INVALID_ARGUMENT);
 
-  return sub_data->take_one_message(ros_message, nullptr, taken);
+  if (!TRACETOOLS_TRACEPOINT_ENABLED(rmw_take)) {
+    return sub_data->take_one_message(ros_message, nullptr, taken);
+  }
+  rmw_message_info_t message_info{};
+  rmw_ret_t ret = sub_data->take_one_message(ros_message, &message_info, taken);
+  TRACETOOLS_DO_TRACEPOINT(
+    rmw_take,
+    static_cast<const void *>(subscription),
+    static_cast<const void *>(ros_message),
+    message_info.source_timestamp,
+    *taken);
+  return ret;
 }
 
 //==============================================================================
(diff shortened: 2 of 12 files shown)


christophebedard:
Oh, looks like this repo aims to support Iron, Jazzy, and Rolling on the `rolling` branch. I will deal with that after a first round of review.

christophebedard:
@Yadunund @clalancette any interest in having this?

Yadunund:
@christophebedard we'd love to have this! We're sort of freezing the `rolling` branch at the moment to focus effortts on getting the #276 merged before we add new features. So let's wait for that to get in before merging this PR . By then Iron would have ben sunsetted so we won't have to worry about failing Iron CI job.

Yadunund:
@christophebedard do you mind resolving the conflicts after rebasing to the latest `rolling` branch?

christophebedard:
I rebased, but now the tracing tests in `test_tracetools` sometimes time out. It seems to happen during `rclcpp::shutdown()`, which might be related to #329.

christophebedard:
I rebased & resolved the new merge conflicts.

I also added client/service instrumentation, see https://github.com/ros2/ros2_tracing/pull/145.

> I rebased, but now the tracing tests in `test_tracetools` sometimes time out. It seems to happen during `rclcpp::shutdown()`, which might be related to #329.

I ran the tests quite a few times and didn't get any failures. Not sure what changed in the code, other than the instrumentation.

rmw_zenoh_cpp/src/detail/zenoh_utils.cpp
```diff
@@ -42,13 +42,18 @@ ZenohSession::~ZenohSession()
 
 ///=============================================================================
 zenoh::Bytes create_map_and_set_sequence_num(
-  int64_t sequence_number, std::array<uint8_t, RMW_GID_STORAGE_SIZE> gid)
+  int64_t sequence_number,
+  std::array<uint8_t, RMW_GID_STORAGE_SIZE> gid,
+  int64_t * source_timestamp)
```
clalancette: I see why you've done this, but I'm honestly not a fan of this change to the API.  It changes all of the arguments from being input-only, to one of them being an out parameter.

I'm going to propose an alternate solution here as a PR targeting this one.

clalancette: See https://github.com/christophebedard/rmw_zenoh/pull/1

christophebedard: yeah that's great, thank you!

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1561,7 +1606,15 @@ rmw_take_response(
...
-  return client_data->take_response(request_header, ros_response, taken);
+  rmw_ret_t ret = client_data->take_response(request_header, ros_response, taken);
+  TRACETOOLS_TRACEPOINT(
+    rmw_take_response,
+    static_cast<const void *>(client),
+    static_cast<const void *>(ros_response),
+    (nullptr != request_header ? request_header->request_id.sequence_number : 0LL),
+    (nullptr != request_header ? request_header->source_timestamp : 0LL),
+    *taken);
+  return ret;
```
clalancette: Huh, interesting.  I think there is an existing bug here in that we really can't allow `request_header` to be nullptr, otherwise it will crash in `client_data->take_response`:

```suggestion
  RMW_CHECK_ARGUMENT_FOR_NULL(request_header, RMW_RET_INVALID_ARGUMENT);

  rmw_ret_t ret = client_data->take_response(request_header, ros_response, taken);
  TRACETOOLS_TRACEPOINT(
    rmw_take_response,
    static_cast<const void *>(client),
    static_cast<const void *>(ros_response),
    request_header->request_id.sequence_number),
    request_header->source_timestamp),
    *taken);
  return ret;
```

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1467,6 +1507,11 @@ rmw_create_client(
   free_rmw_client.cancel();
   free_service_name.cancel();
 
+  if (TRACETOOLS_TRACEPOINT_ENABLED(rmw_client_init)) {
+    auto client_data = static_cast<rmw_zenoh_cpp::ClientData *>(rmw_client->data);
```
clalancette: We don't really need to re-cast this from the `rmw_client->data` pointer; we can actually much more nicely use the shared_ptr.  What I'd suggest here is that above, before we do the `static_cast<void *>` to put this into `rmw_client->data`, we actually do:

```
ClientDataPtr client_data = node_data->get_client_data(rmw_client);
```

Then we can do `rmw_client->data = static_cast<void *>(client_data.get());` in line 1495, and use the `client_data->copy_gid()` in the next line below.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -990,6 +1000,11 @@ rmw_create_subscription(
   free_topic_name.cancel();
   free_rmw_subscription.cancel();
 
+  // rmw does not require GIDs for subscriptions, and GIDs in rmw_zenoh are not based on any ID of
+  // the underlying zenoh objects, so there is no need to collect a GID here
+  rmw_gid_t gid{};
```
clalancette, line 1005: I will note that if you have tracing disabled (like I do locally), then this is a compile warning since `gid` goes unused.  Maybe at least put a `static_cast<void>(gid)` below this?

clalancette (changes requested):
I've left 3 more minor things to fix, then I'm happy with it.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
christophebedard: changed. The `rmw_take_response` doc indeed says that `RMW_RET_INVALID_ARGUMENT` should be returned if it's `nullptr`: https://github.com/ros2/rmw/blob/7e4b0d5f8fb7f7f16fe80419c50c488ca0ab7df6/rmw/include/rmw/rmw.h#L2106

This should be tested by https://github.com/ros2/rmw_implementation/blob/7c17c783109a736ff5c136f178f4d14d6fe1211f/test_rmw_implementation/test/test_client.cpp#L316-L317. Not sure why it wasn't caught.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
christophebedard: changed!

rmw_zenoh_cpp/src/rmw_zenoh.cpp
christophebedard: added!

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1452,7 +1493,8 @@ rmw_create_client(
   // TODO(Yadunund): We cannot store the rmw_node_t * here since this type erased
   // Client handle will be returned in the rmw_clients_t in rmw_wait
   // from which we cannot obtain ClientData.
-  rmw_client->data = static_cast<void *>(node_data->get_client_data(rmw_client).get());
+  ClientDataPtr client_data = node_data->get_client_data(rmw_client);
```
```suggestion
  rmw_zenoh_cpp::ClientDataPtr client_data = node_data->get_client_data(rmw_client);
```

clalancette (changes requested):
One more fix, then I'm happy with this.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
christophebedard: thanks. `rmw_zenoh_cpp` was failing to build on my (22.04) system due to something else and I couldn't figure out why.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1467,6 +1509,10 @@ rmw_create_client(
   free_rmw_client.cancel();
   free_service_name.cancel();
 
+  if (TRACETOOLS_TRACEPOINT_ENABLED(rmw_client_init)) {
+    auto gid = client_data->copy_gid();
+    TRACETOOLS_DO_TRACEPOINT(rmw_client_init, static_cast<const void *>(rmw_client), gid.data());
+  }
```
clalancette: Sorry, one more thing.  Since we've simplified this enough now, can this be:

```
TRACETOOLS_TRACEPOINT(
  rmw_client_init,
  static_cast<const void *>(rmw_client),
  client_data->copy_gid().data());
```

(I ended up noticing it because `gid` ends up being unused when tracing is compiled out)

clalancette (changes requested):

rmw_zenoh_cpp/src/rmw_zenoh.cpp
christophebedard: changed!

`TRACETOOLS_TRACEPOINT_ENABLED()` should be defined to `false` when `TRACETOOLS_DISABLED`, not sure why it's not optimized out

clalancette: > `TRACETOOLS_TRACEPOINT_ENABLED()` should be defined to `false` when `TRACETOOLS_DISABLED`, not sure why it's not optimized out

Ah, that's because I was building with `-DCMAKE_BUILD_TYPE=Debug`.  Anyway, this is shorter now and accomplishes the same thing.

clalancette (approved):
This looks great.  Thanks for all of the iteration, and for being patient with us as we revamped other things.  I'm going to go ahead and merge now.

cybaol:
@christophebedard `TRACETOOLS_TRACEPOINT` seems break jazzy build (replicated on archlinux).

christophebedard:
Make sure to build the `tracetools` package (from the ros2/ros2_tracing repo) from source. If you rely on the binaries, rmw_zenoh will indeed not build.

christophebedard:
Oh, I guess this does break Jazzy builds. This will only work with Rolling (built from source for now).

The new service/client tracepoints would need to be excluded on Jazzy.

christophebedard:
So CI passed for this PR because the `build_and_test_binary_jazzy` job wasn't actually testing Jazzy: https://github.com/ros2/rmw_zenoh/actions/runs/12422154058/job/34683299857. It's my fault, I set `action-ros-ci-repos-override:` to use the `ros2_tracing` branch that enables tracing tests for `rmw_zenoh_cpp` (https://github.com/ros2/ros2_tracing/pull/140). The same job is indeed now failing on `rolling`: https://github.com/ros2/rmw_zenoh/actions/runs/12432407339/job/34711738348.

Since this repo doesn't use distro-specific branches yet, could we add distro-specific `#define`s`? Since I broke the build, I've opened a PR with a quick-and-dirty solution: #355.

PR #296, opened 2024-10-10
Rename guid in liveliness_utils to keyexpr_hash.

### LINKED ISSUE #148: Align on Transport GID and Liveliness GUID

### LINKED ISSUE #290: get_entities_info_by_topic is not setting gid

clalancette:
This more clearly reflect what it is, which is a
hash of the keyexpression that is used to index into various maps.

FYI @ahcorde and @Yadunund .  I think this more accurately reflects what this particular data is used for, and helps (at least in my mind) to clear up some confusion about the "real" GID that we need to return from [get_entities_info_by_topic](https://github.com/ros2/rmw_zenoh/blob/439d6dc6b0325811793533414dee89196cd72186/rmw_zenoh_cpp/src/detail/graph_cache.cpp#L1086-L1215).

Going forward, what I think we should do is the following:
1.  Close #148 , because there is really no direct relationship between those things.
2.  Close #291 (since it is attempting to implement #148).
3.  Fix #290 by getting the GID from https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp#L74 in this [line](https://github.com/ros2/rmw_zenoh/blob/439d6dc6b0325811793533414dee89196cd72186/rmw_zenoh_cpp/src/detail/graph_cache.cpp#L1208) (there is probably some support work that needs to be done to enable access to that structure).

What do you think?  Does that make sense?

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -319,7 +319,7 @@ void GraphCache::take_entities_with_events(const EntityEventMap & entities_with_
   for (const auto & [local_entity, event_set] : entities_with_events) {
     // Trigger callback set for this entity for the event type.
     GraphEventCallbackMap::const_iterator event_callbacks_it =
-      event_callbacks_.find(local_entity->guid());
+      event_callbacks_.find(local_entity->keyexpr_hash());
     if (event_callbacks_it != event_callbacks_.end()) {
       for (const rmw_zenoh_event_type_t & event_type : event_set) {
         GraphEventCallbacks::const_iterator callback_it =
@@ -1240,7 +1240,7 @@ rmw_ret_t GraphCache::service_server_is_available(
 
 ///=============================================================================
 void GraphCache::set_qos_event_callback(
-  std::size_t entity_guid,
+  std::size_t entity_keyexpr_hash,
   const rmw_zenoh_event_type_t & event_type,
   GraphCacheEventCallback callback)
 {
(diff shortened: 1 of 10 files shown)


ahcorde (approved):

ahcorde:
[ros2.repos](https://gist.github.com/ahcorde/921fb466b20115335d2ba1174b9e3f43)

# rmw_zenoh/yadu/raii-service
* rmw_zenoh commit: https://github.com/ros2/rmw_zenoh/commit/a43b86155511255d6ad2e2727a0a1e23ed2300f3

| rcl | None | None |
| rcl_action | None | None |

ahcorde:
@Yadunund are you fine with this change ?

clalancette:
> @Yadunund are you fine with this change ?

Let's hold off on this one; I'm working on another change (which will probably include some parts of this), that I think will solve this entire problem.

clalancette:
Closing in favor of #298

PR #298, opened 2024-10-15
Add the GID hash to get_entities_info_by_topic

### LINKED ISSUE #148: Align on Transport GID and Liveliness GUID

### LINKED ISSUE #290: get_entities_info_by_topic is not setting gid

clalancette:
This PR:
1. Fixes the failing rclcpp test `test_node_interfaces__node_graph`
2. Fixes a bug where if a node using rmw_zenoh had two publishers of the same type, topic_name, and QoS, it would be returned as a single one.
3. Resolve a TODO about combining GID and GUID.

To get there, we do the following:
1.  Import a simplified version of the xxHash library (https://github.com/Cyan4973/xxHash).  While xxHash is concerned with correctness and performance, we really only care about the correctness part (at the end of the day, we don't create *that* many entities in a ROS graph).  So this simplified version only imports the parts that are purely portable.  It won't be as fast, but for practical purposes it shouldn't matter.  If we find in the future that this is a performance bottleneck, we can take a dependency on the "real" xxHash and benefit from its per-platform optimizations.
2. Use the imported xxHash library to hash the arbitrarily-long liveliness string into a 128-bit (16-byte) GID.  This GID is useful for identifying different entities, and is guaranteed to be unique across the system.
3. Rename the liveliness_utils "GUID" to `keyexpr_hash`.  This `keyexpr_hash` is a size_t hash, using `std::hash` from the STL.  It is used for indexing into various maps, speeding up lookups (this is effectively a repeat of #296).
4. Removes `generate_random_gid` in favor of copying the GID we created above out of the `Entity`.  We no longer need random ones because we have guaranteed-unique ones in the system.
5. Removes `hash_gid` in favor of using the size_t hash from the entity.  Again, we don't need it since this is now stored in the `Entity`.
6. Reimplements `get_entities_info_by_topic`, simplifying it, fixing a bug in it, and adding in the GID.

This PR fixes #290 and fixes #148, while replacing/incorporating parts of #296 and #291 (which I'll close).

@ahcorde I'd appreciate it if you could run CI on this, to check that I haven't broken anything new.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
@@ -28,6 +28,7 @@
 
 #include "logging_macros.hpp"
 #include "qos.hpp"
+#include "simplified_xxhash3.hpp"
 
 #include "rcpputils/scope_exit.hpp"
 
@@ -441,6 +442,16 @@ Entity::Entity(
     // Append the delimiter unless it is the last component.
     this->liveliness_keyexpr_ += KEYEXPR_DELIMITER;
   }
+
+  // Here we hash the complete std::string that comprises the liveliness keyexpression
+  // into a GID that is associated with every entity in the system.  This is the GID that will be
+  // returned to the RMW layer as necessary.
+  simplified_XXH128_hash_t keyexpr_gid =
+    simplified_XXH3_128bits(this->liveliness_keyexpr_.c_str(), this->liveliness_keyexpr_.length());
+  memcpy(this->gid_, &keyexpr_gid.low64, sizeof(keyexpr_gid.low64));
+  memcpy(this->gid_ + sizeof(keyexpr_gid.low64), &keyexpr_gid.high64, sizeof(keyexpr_gid.high64));
+
+  // We also hash the liveliness keyexpression into a size_t that we use to index into our maps
   this->guid_ = std::hash<std::string>{}(this->liveliness_keyexpr_);
 }
 
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
@@ -452,7 +452,7 @@ Entity::Entity(
   memcpy(this->gid_ + sizeof(keyexpr_gid.low64), &keyexpr_gid.high64, sizeof(keyexpr_gid.high64));
 
   // We also hash the liveliness keyexpression into a size_t that we use to index into our maps
-  this->guid_ = std::hash<std::string>{}(this->liveliness_keyexpr_);
+  this->keyexpr_hash_ = std::hash<std::string>{}(this->liveliness_keyexpr_);
 }
 
 ///=============================================================================
@@ -596,9 +596,9 @@ std::string Entity::id() const
 }
 
 ///=============================================================================
-std::size_t Entity::guid() const
+std::size_t Entity::keyexpr_hash() const
 {
-  return this->guid_;
+  return this->keyexpr_hash_;
 }
 
 ///=============================================================================
@@ -643,9 +643,7 @@ std::string Entity::liveliness_keyexpr() const
 ///=============================================================================
 bool Entity::operator==(const Entity & other) const
 {
-  // TODO(Yadunund): If we decide to directly store the guid as a
-  // rmw_gid_t type, we should rely on rmw_compare_gids_equal() instead.
-  return other.guid() == guid_;
+  return other.keyexpr_hash() == keyexpr_hash_;
 }
 
 ///=============================================================================
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
@@ -680,16 +680,4 @@ std::string demangle_name(const std::string & input)
   return output;
 }
 }  // namespace liveliness
-
-///=============================================================================
-size_t hash_gid(const uint8_t gid[RMW_GID_STORAGE_SIZE])
-{
-  std::stringstream hash_str;
-  hash_str << std::hex;
-  size_t i = 0;
-  for (; i < (RMW_GID_STORAGE_SIZE - 1); i++) {
-    hash_str << static_cast<int>(gid[i]);
-  }
-  return std::hash<std::string>{}(hash_str.str());
-}
 }  // namespace rmw_zenoh_cpp
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
@@ -451,8 +451,8 @@ Entity::Entity(
   memcpy(this->gid_, &keyexpr_gid.low64, sizeof(keyexpr_gid.low64));
   memcpy(this->gid_ + sizeof(keyexpr_gid.low64), &keyexpr_gid.high64, sizeof(keyexpr_gid.high64));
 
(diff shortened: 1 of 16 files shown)


rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -441,7 +442,17 @@ Entity::Entity(
...
+
+  // Here we hash the complete std::string that comprises the liveliness keyexpression
+  // into a GID that is associated with every entity in the system.  This is the GID that will be
+  // returned to the RMW layer as necessary.
+  simplified_XXH128_hash_t keyexpr_gid =
+    simplified_XXH3_128bits(this->liveliness_keyexpr_.c_str(), this->liveliness_keyexpr_.length());
+  memcpy(this->gid_, &keyexpr_gid.low64, sizeof(keyexpr_gid.low64));
+  memcpy(this->gid_ + sizeof(keyexpr_gid.low64), &keyexpr_gid.high64, sizeof(keyexpr_gid.high64));
+
+  // We also hash the liveliness keyexpression into a size_t that we use to index into our maps
```
Yadunund:
```suggestion
  // We also hash the liveliness keyexpression into a size_t that we use to index into our maps.
```

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -441,7 +442,17 @@ Entity::Entity(
...
   }
-  this->guid_ = std::hash<std::string>{}(this->liveliness_keyexpr_);
+
+  // Here we hash the complete std::string that comprises the liveliness keyexpression
+  // into a GID that is associated with every entity in the system.  This is the GID that will be
+  // returned to the RMW layer as necessary.
+  simplified_XXH128_hash_t keyexpr_gid =
+    simplified_XXH3_128bits(this->liveliness_keyexpr_.c_str(), this->liveliness_keyexpr_.length());
+  memcpy(this->gid_, &keyexpr_gid.low64, sizeof(keyexpr_gid.low64));
+  memcpy(this->gid_ + sizeof(keyexpr_gid.low64), &keyexpr_gid.high64, sizeof(keyexpr_gid.high64));
```
Yadunund, line 452: I think we should move this logic (ie, getting a `std::size_t` hash from a `std::string`) into `hash_gid()` defined in `liveliness_utils.hpp` since it will also be needed within `ServiceData::take_request`. See my comment below.

rmw_zenoh_cpp/src/detail/rmw_service_data.cpp
```diff
@@ -344,12 +344,12 @@ rmw_ret_t ServiceData::take_request(
   request_header->received_timestamp = now_ns.count();
 
   // Add this query to the map, so that rmw_send_response can quickly look it up later.
-  const size_t hash = rmw_zenoh_cpp::hash_gid(request_header->request_id.writer_guid);
-  std::unordered_map<size_t, SequenceToQuery>::iterator it = sequence_to_query_map_.find(hash);
+  std::unordered_map<size_t, SequenceToQuery>::iterator it =
+    sequence_to_query_map_.find(entity_->keyexpr_hash());
```
Yadunund: I'm confused by this change and the related ones below.

The `sequence_to_query_map_` maps the client's GID hash to a map of sequence number from the same client to the `ZenohQuery`. Line 355 highlights highlights this. Also see [this line](https://github.com/ros2/rmw_zenoh/blob/1fb2b1e903c29c774525c84d9c9d5be8badb3cc8/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L1787-L1788) from `rmw_send_request` that appends the client's GID to the header.

But here we're always using the service entity's GID hash to lookup `sequence_to_query_map_`. imo we still need to call `rmw_zenoh_cpp::hash_gid(request_header->request_id.writer_guid)`. See my comment above on retaining this function.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -665,30 +680,4 @@ std::string demangle_name(const std::string & input)
...
-size_t hash_gid(const uint8_t gid[RMW_GID_STORAGE_SIZE])
-{
-  std::stringstream hash_str;
-  hash_str << std::hex;
-  size_t i = 0;
-  for (; i < (RMW_GID_STORAGE_SIZE - 1); i++) {
-    hash_str << static_cast<int>(gid[i]);
-  }
-  return std::hash<std::string>{}(hash_str.str());
-}
```
Yadunund: Retain function but move `simplified_XXH3` logic here.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -1606,6 +1604,8 @@ rmw_create_client(
     return nullptr;
   }
 
+  client_data->entity->copy_gid(client_data->client_gid);
+
```
Yadunund: Technically can get rid of `client_data->client_gid` like we did in `PublisherData`?

Yadunund (changes requested):
Apart from the issue with changes to `ServiceData:: sequence_to_query_map_ `, this PR looks good to me.

ahcorde:
[ros2.repos](https://gist.github.com/ahcorde/3e482ed9264bd16974aadcca3d57c998)

# rmw_zenoh/clalancette/fix-gid-hash
* rmw_zenoh commit: https://github.com/ros2/rmw_zenoh/commit/5503ad8bad4bf50c4c7b48afbedd08dd99cb7050

| rcl | None | None |
| rcl_action | None | None |
| rclcpp   | test_executors_timer_cancel_behavior        | Timeout (Check logs)      |

rmw_zenoh_cpp/src/detail/rmw_service_data.cpp
clalancette: You are right; commit eb2239213168d6ebcc897363f7fcf31c77a186bf is completely wrong (I wonder how it worked at all in my testing).  I've now reverted it in 728c072d9b488d274b95c3843b09b884a95e877a .

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
clalancette: This is now changed to use `hash_gid` in dbd7bfe141099c89ecf4f0b314144e5185dea3dc

rmw_zenoh_cpp/src/rmw_zenoh.cpp
clalancette: Good call; this is now removed in ff128e348d4acba6ab7b73ad7c7545cd0367818e

Yadunund (approved):
LGTM!

PR #299, opened 2024-10-16
Cleanup SubscriptionData creation and destruction.

clalancette:
In particular, make SubscriptionData::init/shutdown mirror images, in that they handle all of their internal data.  That means that rmw_destroy_subscription() really only calls delete_sub_data().

To achieve this, we had to split out SubscriptionData::init from construction, since we rely on shared pointers, but you can't use shared_from_this() in the constructor.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_subscription_data.hpp
@@ -42,7 +42,7 @@
 namespace rmw_zenoh_cpp
 {
 ///=============================================================================
-class SubscriptionData final
+class SubscriptionData final : public std::enable_shared_from_this<SubscriptionData>
 {
 public:
   struct Message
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/detail/rmw_subscription_data.hpp
```diff
@@ -132,6 +132,10 @@ class SubscriptionData final
     const void * type_support_impl,
     std::unique_ptr<MessageTypeSupport> type_support);
 
+  bool init();
+
+  bool initted_{false};
```
Yadunund: Can we
- call this `initialized_` instead
- move it's definition above `bool is_shutdown_`
- initialize the value to `false` in the constructor initializer list

Yadunund (changes requested):
Thanks for cleaning this up! Minor suggestion to consider.

rmw_zenoh_cpp/src/detail/rmw_subscription_data.hpp
clalancette: Done in efd8918.

Yadunund (approved):

PR #301, opened 2024-10-22
Updated error message when getting source timestamp

jmount1992:
The error message for when getting the source timestamp appeared to be a copy+paste error for the sequence number.


Yadunund (approved):
Thanks for the enhancement!

PR #304, opened 2024-10-29
Additional fixes for yadu/raii-client.

clalancette:


rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -138,15 +133,20 @@ void client_data_drop(void * data)
...
-  // weaK_ptr<ClientData> in this callback instead.
+  std::lock_guard<std::mutex> lock(num_in_flight_mutex);
   auto num_in_flight_it = num_in_flight_map.find(client_data);
-  if (num_in_flight_it != num_in_flight_map.end()) {
-    --num_in_flight_it->second;
+  if (num_in_flight_it == num_in_flight_map.end()) {
+    // This should never happen
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to find object in num_in_flight_map."
```
Yadunund:
```suggestion
      "Unable to find object in num_in_flight_map. Report this bug."
```

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -138,15 +133,20 @@ void client_data_drop(void * data)
...
+    );
+    return;
+  }
+
+  --num_in_flight_it->second;
+  if (num_in_flight_it->second == 0) {
```
Yadunund: We only need to remove the client from `deleted_clients` if it was deleted in the past. I guess the call to erase is harmless if the client was not deleted before but this makes the operation semantically clear.
```suggestion
  if (num_in_flight_it->second == 0 && deleted_clients.count(client) > 0) {
```

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -233,15 +233,21 @@ std::shared_ptr<ClientData> ClientData::make(
...
+  std::shared_ptr<ClientData> client_data;
+  do {
+    client_data = std::shared_ptr<ClientData>(
+      new ClientData{
+        node,
+        std::move(entity),
+        request_members,
+        response_members,
+        std::move(request_type_support),
+        std::move(response_type_support)
```
Yadunund: If the first pointer is a duplicate, creating the second instance will segfault since some of the data members have been moved into the first one.
```suggestion
        entity,
        request_members,
        response_members,
        request_type_support,
        response_type_support
```

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -233,15 +233,21 @@ std::shared_ptr<ClientData> ClientData::make(
...
+      });
+    duplicate_pointers.push_back(client_data);
```
Yadunund, line 249: Why do we store the instantiated `client_data` ptr in this vector?

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -536,9 +541,28 @@ ClientData::~ClientData()
...
+      // Since there is still something in flight, we need to just add
+      // it to the deleted_clients; it will be deleted when the last
+      // outstanding query finishes.
+      deleted_clients.insert(this);
+    }
+  } else {
+    // This should never happen
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Error finding client /%s in num_in_flight_map.",
```
Yadunund:
```suggestion
      "Error finding client /%s in num_in_flight_map. Report this bug.",
```

Yadunund (approved):
Thanks a lot for addressing the outstanding issues with this branch! Left some minor comments to consider.

Yadunund (changes requested):
Sorry I forgot we need to address the issue with `std::move` before we can merge. But apart from that, this generally looks good to me.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
clalancette: If we didn't do this, then we could conceivably get the same address over and over again.  Consider the case where we have the pointer 0x1 in `deleted_clients`.  The first call to `new ClientData` might return 0x1.  At that point, we'd detect that this is in `deleted_clients`, and loop around again.  But right before we make the second call to `new ClientData`, we drop the reference to the old one, at which point it may be freed.  At that point the allocator might decide to give us address 0x1 again, and we'd never make any progress here.

By holding all of the shared_ptr in a vector like this, we guarantee that the allocator will give us a new address.  And then at the end of the function, when `std::vector` goes out of scope, we'll free all of them.

Yadunund (approved):

PR #305, opened 2024-11-04
Fix the received_timestamp for client data.

clalancette:
These should be timestamped at the time they are received, not at the time that they are taken (that might be much later).

I found this while debugging some of the problems in #293 

@Yadunund This will cause merge conflicts with #293, but I thought it was important enough on its own that we should get it in now while I am still debugging that one.  Assuming this is approved, I'll merge this and then rebase #293 on top of this one.


ahcorde (approved):

Yadunund (approved):
Nice catch!

PR #310, opened 2024-11-13
Redo the per-client-data in-flight monitoring.

clalancette:
The previous approach with using a global map of
pointers to shared_ptr does not work for one simple reason; on shutdown, C++ does not define the order in which statics are destroyed.  Thus it is possible (and even likely in some cases) that on Ctrl-C, the global map would be deleted, and then we would try to access it, leading to crashes.  I was seeing some of these while running the rcl tests.

Instead, go back to the previous solution where we were storing num_in_flight completely within the ClientData structure. This requires doing a bit of gymnastics with the data types, and reintrodues the possible UB during shutdown if the structure was destructed but there is still queries in flight.  But it otherwise fixes all of the tests for me locally.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
@@ -43,10 +43,6 @@
 namespace
 {
 
-std::mutex client_data_ptr_to_shared_ptr_map_mutex;
-std::unordered_map<const rmw_zenoh_cpp::ClientData *,
-  std::shared_ptr<rmw_zenoh_cpp::ClientData>> client_data_ptr_to_shared_ptr_map;
-
 ///=============================================================================
 void client_data_handler(z_owned_reply_t * reply, void * data)
 {
@@ -59,16 +55,7 @@ void client_data_handler(z_owned_reply_t * reply, void * data)
     return;
   }
 
-  std::shared_ptr<rmw_zenoh_cpp::ClientData> client_data_shared_ptr{nullptr};
-  {
-    std::lock_guard<std::mutex> lk(client_data_ptr_to_shared_ptr_map_mutex);
-    if (client_data_ptr_to_shared_ptr_map.count(client_data) == 0) {
-      return;
-    }
-    client_data_shared_ptr = client_data_ptr_to_shared_ptr_map[client_data];
-  }
-
-  if (client_data_shared_ptr->is_shutdown()) {
+  if (client_data->is_shutdown()) {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -94,13 +81,28 @@ void client_data_handler(z_owned_reply_t * reply, void * data)
...
 }
 
+///=============================================================================
+void client_data_drop(void * data)
+{
+  auto client_data = static_cast<rmw_zenoh_cpp::ClientData *>(data);
+  if (client_data == nullptr) {
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp",
+      "Unable to obtain client_data_t "
```
Yadunund:
```suggestion
      "Unable to obtain client_data_t from data in client_data_drop. "
```

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -545,13 +542,41 @@ rmw_ret_t ClientData::shutdown()
...
+///=============================================================================
+void ClientData::decrement_in_flight_and_conditionally_remove()
+{
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  --num_in_flight_;
+
+  if (is_shutdown_ && num_in_flight_ == 0) {
+    rmw_context_impl_s * context_impl = static_cast<rmw_context_impl_s *>(rmw_node_->data);
+    std::shared_ptr<rmw_zenoh_cpp::NodeData> node_data = context_impl->get_node_data(rmw_node_);
+    node_data->delete_client_data(rmw_client_);
```
Yadunund: Could there be a possibility that the `context_impl` and `node_data` are nullptr? Eg. if for some reason the node is cleaned before we get the query reply. So we should check that these are not nullptr before proceeding with `get_node_data` and `delete_client_data`?
```suggestion
    rmw_context_impl_s * context_impl = static_cast<rmw_context_impl_s *>(rmw_node_->data);
    if (context_impl == nullptr) {
      return;
    }
    std::shared_ptr<rmw_zenoh_cpp::NodeData> node_data = context_impl->get_node_data(rmw_node_);
    if (node_data == nullptr) {
      return;
    }
    node_data->delete_client_data(rmw_client_);
```

Yadunund (changes requested):
Minor comments to consider!

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
clalancette: Good idea, fixed in [ec6c181](https://github.com/ros2/rmw_zenoh/pull/310/commits/ec6c181797d9d234f9577de6154ede4067d4b760)

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
clalancette: Yeah, really good call.  Fixed in [ec6c181](https://github.com/ros2/rmw_zenoh/pull/310/commits/ec6c181797d9d234f9577de6154ede4067d4b760)

Yadunund (approved):

PR #313, opened 2024-11-14
Don't shutdown contained entities.

clalancette:
It is tempting to think of the entities in the RMW graph as a hierarchy, where an rmw_context_t contains zero or more rmw_node_t, and rmw_node_t contains zero or more publishers, subscriptions, clients, and services.

However, the reality is that the upper layers (particularly rclcpp and rclpy) don't exactly view the entities like that. More specifically, each entity is considered standalone, that happens to have linkage to other entities in the graph. For example, a publisher is considered to be a standalone entity that happens to be linked to a particular node.

Because of this, it is not proper to shutdown all nodes within a context when the context is shutdown.  The node should continue to live on until it is shutdown.  And a similar argument goes for the node shutdown; it should not shutdown the publishers, subscriptions, clients, and services that are contained within it.

This manifested itself as a exception that was being thrown in some of the tests in test_communication.  Because it is using a loop with rclcpp::ok(), followed by a publisher->publish(), the test would sometimes throw an exception if Ctrl-C was hit between the rclcpp::ok() and the publish() call.  And that's because even though the context has been shut down, the publisher is an independent entity that should continue to exist until the next rclcpp::ok().

The fix here is simple; don't shut "contained" entities down during a context or node shutdown.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -233,19 +233,6 @@ public:
         return ret;
       }
 
-      // Shutdown all the nodes in this context.
-      for (auto node_it = nodes_.begin(); node_it != nodes_.end(); ++node_it) {
-        ret = node_it->second->shutdown();
-        if (ret != RMW_RET_OK) {
-          RMW_ZENOH_LOG_ERROR_NAMED(
-            "rmw_zenoh_cpp",
-            "Unable to shutdown node with id %zu. rmw_ret_t code: %zu.",
-            node_it->second->id(),
-            ret
-          );
-        }
-      }
-
       z_undeclare_subscriber(z_move(graph_subscriber_));
       if (shm_manager_.has_value()) {
         z_drop(z_move(shm_manager_.value()));
--- a/rmw_zenoh_cpp/src/detail/rmw_node_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_node_data.cpp
@@ -402,56 +402,6 @@ rmw_ret_t NodeData::shutdown()
     return ret;
   }
 
-  // Shutdown all the entities within this node.
-  for (auto pub_it = pubs_.begin(); pub_it != pubs_.end(); ++pub_it) {
-    ret = pub_it->second->shutdown();
-    if (ret != RMW_RET_OK) {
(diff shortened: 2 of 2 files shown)


Yadunund:
Hmm this contradicts the documentation for `rcl_node_fini` seen [here](https://github.com/ros2/rcl/blob/2615b0e8a5e3d9188e50412337e3b306991b05a6/rcl/include/rcl/node.h#L164-L170). Could it be that the test expecation needs to be fixed instead?

clalancette:
> Could it be that the test expecation needs to be fixed instead?

I don't think so; the pattern of `while (rclcpp::ok()) { pub->publish() }` is pretty well established.

> Hmm this contradicts the documentation for `rcl_node_fini` seen [here](https://github.com/ros2/rcl/blob/2615b0e8a5e3d9188e50412337e3b306991b05a6/rcl/include/rcl/node.h#L164-L170).

I read the documentation for `rcl_node_fini` as saying "it is the users responsibility to shut down all entities before shutting down the node".

Further, I'll point out that none of the rest of the current RMWs clean up entities when calling `rmw_destroy_node`:
* [rmw_cyclonedds_cpp](https://github.com/ros2/rmw_cyclonedds/blob/0c2904d725dade83cc0b1ffde931b46f9d28269f/rmw_cyclonedds_cpp/src/rmw_node.cpp#L1746-L1766) - removes this node from the graph, frees memory associated with the node, then decrements the references to the context
* [rmw_fastrtps_cpp](https://github.com/ros2/rmw_fastrtps/blob/9d2150ffc1b549d0adc71971b7f350e325b61340/rmw_fastrtps_cpp/src/rmw_node.cpp#L77-L116) - calls [__rmw_destroy_node](https://github.com/ros2/rmw_fastrtps/blob/9d2150ffc1b549d0adc71971b7f350e325b61340/rmw_fastrtps_shared_cpp/src/rmw_node.cpp#L116-L133), which removes this node from the graph and frees memory associated with the node, then decrements the references to the context
* [rmw_connextdds](https://github.com/ros2/rmw_connextdds/blob/92f0925e90094519351cfe7db05d8f282bfe85d3/rmw_connextdds_common/src/common/rmw_node.cpp#L177-L213) - removes this node from the graph and frees the memory associated with the node, then decrements the references to the context

So I think that we should do the same as all of the rest, and not actually destroy all of the associated entities here.

Yadunund (approved):
Thanks for clarifying!

PR #327, opened 2024-12-04
Port zenoh_c to zenoh_cpp

ahcorde:
For visibility reason I created a branch called `dev/1.0.0` in this repository, otherwise I can't open the PR 

The goal of this PR is to port the `zenoh_c` API to `zenoh_cpp`.

I added a hack in the zenoh_cpp API to be able to make the transition easier. Edit the following file:

```bash
nano +83 install/opt/zenoh_cpp_vendor/include/zenoh/api/base.hxx
```

And add this change:

```cpp
public:
    OwnedType _0;
protected:
```
------

 - zenohd is ported to `zenoh_cpp`

PR #333, opened 2024-12-06
refactor: wrap the zenoh session with a shared pointer

### LINKED ISSUE #44: Issues caused by not shutting down the contained entity
Several failures have been found since https://github.com/ros2/rmw_zenoh/pull/313 was introduced. 

1. iRobot test begins hanging
2. github.com/ZettaScaleLabs/ros2-simple-performance begins hanging

The workaround is using this branch https://github.com/ZettaScaleLabs/rmw_zenoh/tree/dev/1.0.0-revert-not-shutdown-contained-entities with the buggy commit reverted

YuanYuYuan:
Closes https://github.com/ZettaScaleLabs/rmw_zenoh/issues/44.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
@@ -269,6 +269,7 @@ bool ClientData::init(std::shared_ptr<ZenohSession> session)
       z_drop(z_move(this->token_));
     });
 
+  sess_ = session;
   initialized_ = true;
 
   free_ros_keyexpr.cancel();
@@ -409,7 +410,6 @@ rmw_ret_t ClientData::send_request(
   if (context_impl == nullptr) {
     return RMW_RET_INVALID_ARGUMENT;
   }
-  sess_ = context_impl->session();
 
   size_t max_data_length = (
     request_type_support_->get_estimated_serialized_size(
--- a/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
@@ -105,7 +105,7 @@ namespace rmw_zenoh_cpp
 {
 ///=============================================================================
 std::shared_ptr<ClientData> ClientData::make(
-  const z_loaned_session_t * session,
+  std::shared_ptr<ZenohSession> session,
   const rmw_node_t * const node,
   const rmw_client_t * client,
   liveliness::NodeInfo node_info,
@@ -167,7 +167,7 @@ std::shared_ptr<ClientData> ClientData::make(
 
   std::size_t domain_id = node_info.domain_id_;
   auto entity = liveliness::Entity::make(
-    z_info_zid(session),
+    z_info_zid(session->loan()),
     std::to_string(node_id),
     std::to_string(service_id),
     liveliness::EntityType::Client,
@@ -192,6 +192,7 @@ std::shared_ptr<ClientData> ClientData::make(
       node,
       client,
       entity,
+      session,
       request_members,
       response_members,
       request_type_support,
@@ -211,6 +212,7 @@ ClientData::ClientData(
   const rmw_node_t * rmw_node,
   const rmw_client_t * rmw_client,
   std::shared_ptr<liveliness::Entity> entity,
+  std::shared_ptr<ZenohSession> sess,
   const void * request_type_support_impl,
   const void * response_type_support_impl,
   std::shared_ptr<RequestTypeSupport> request_type_support,
@@ -218,6 +220,7 @@ ClientData::ClientData(
 : rmw_node_(rmw_node),
   rmw_client_(rmw_client),
   entity_(std::move(entity)),
+  sess_(std::move(sess)),
   request_type_support_impl_(request_type_support_impl),
   response_type_support_impl_(response_type_support_impl),
   request_type_support_(request_type_support),
@@ -232,7 +235,7 @@ ClientData::ClientData(
 }
 
 ///=============================================================================
-bool ClientData::init(const z_loaned_session_t * session)
+bool ClientData::init(std::shared_ptr<ZenohSession> session)
 {
   if (z_keyexpr_from_str(
     &this->keyexpr_,
@@ -250,7 +253,7 @@ bool ClientData::init(const z_loaned_session_t * session)
   z_view_keyexpr_t liveliness_ke;
   z_view_keyexpr_from_str(&liveliness_ke, liveliness_keyexpr.c_str());
   if (z_liveliness_declare_token(
-      session,
+      session->loan(),
       &this->token_,
       z_loan(liveliness_ke),
       NULL
@@ -406,6 +409,7 @@ rmw_ret_t ClientData::send_request(
   if (context_impl == nullptr) {
     return RMW_RET_INVALID_ARGUMENT;
   }
+  sess_ = context_impl->session();
 
   size_t max_data_length = (
     request_type_support_->get_estimated_serialized_size(
@@ -470,7 +474,7 @@ rmw_ret_t ClientData::send_request(
   z_owned_closure_reply_t zn_closure_reply;
   z_closure(&zn_closure_reply, client_data_handler, client_data_drop, this);
   z_get(
-    context_impl->session(),
+    sess_->loan(),
     z_loan(keyexpr_), "",
     z_move(zn_closure_reply),
     &opts);
@@ -535,6 +539,7 @@ void ClientData::_shutdown()
     z_drop(z_move(keyexpr_));
   }
 
+  sess_.reset();
   is_shutdown_ = true;
 }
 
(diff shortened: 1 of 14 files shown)


rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
```diff
@@ -406,6 +409,7 @@ rmw_ret_t ClientData::send_request(
   if (context_impl == nullptr) {
     return RMW_RET_INVALID_ARGUMENT;
   }
+  sess_ = context_impl->session();
```
clalancette: This doesn't seem correct to me.  If there are multiple `send_request` in-flight at the same time, then we'll only take *one* additional reference for the whole class, instead of one per in-flight request.  Instead, I think we should probably take a reference per-outstanding-request.  To do that while using the current `zenoh-c`, I think we'll have to manually do some work here.  In particular, I have something like this in mind:

```
struct ClientDataWrapper
{
    ClientDataWrapper(std::shared_ptr<Session> sess, std::shared_ptr<ClientData> cd)
    : session(sess), client_data(cd)
    {
    }
    std::shared_ptr<Session> session;
    std::shared_ptr<ClientData> client_data;
};
```

And then right here we would do:

```
ClientDataWrapper * wrapper = new ClientDataWrapper(context_impl->session(), shared_from_this());
```

And then we'd pass the `wrapper` into the `z_closure` as the user_data:

```
  z_closure(&zn_closure_reply, client_data_handler, client_data_drop, wrapper);
```

And then in `client_data_handler`, we'd use the wrapper class:

```
void client_data_handler(z_loaned_reply_t * reply, void * data)
{
  auto wrapper = static_cast<ClientDataWrapper *>(data);
  if (wrapper == nullptr) {
    RMW_ZENOH_LOG_ERROR_NAMED(
      "rmw_zenoh_cpp",
      "Unable to obtain client_data_t from data in client_data_handler."
    );
    return;
  }
  if (wrapper->client_data->is_shutdown()) {
    return;
  }
  if (!z_reply_is_ok(reply)) {
    const z_loaned_reply_err_t * err = z_reply_err(reply);
    const z_loaned_bytes_t * err_payload = z_reply_err_payload(err);
    z_owned_string_t err_str;
    z_bytes_to_string(err_payload, &err_str);
    RMW_ZENOH_LOG_ERROR_NAMED(
      "rmw_zenoh_cpp",
      "z_reply_is_ok returned False for keyexpr %s. Reason: %.*s",
      wrapper->client_data->topic_info().topic_keyexpr_.c_str(),
      static_cast<int>(z_string_len(z_loan(err_str))),
      z_string_data(z_loan(err_str)));
    z_drop(z_move(err_str));
    return;
  }
  std::chrono::nanoseconds::rep received_timestamp =
    std::chrono::system_clock::now().time_since_epoch().count();
  wrapper->client_data->add_new_reply(
    std::make_unique<rmw_zenoh_cpp::ZenohReply>(reply, received_timestamp));
}
```

And then finally, when `client_data_drop` is called, we'd free the memory:

```
void client_data_drop(void * data)
{
  auto wrapper = static_cast<ClientDataWrapper *>(data);
  if (wrapper == nullptr) {
    RMW_ZENOH_LOG_ERROR_NAMED(
      "rmw_zenoh_cpp",
      "Unable to obtain client_data_t from data in client_data_drop."
    );
    delete wrapper;
  }
}
```

What do you think?

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -438,11 +441,16 @@ rmw_ret_t PublisherData::shutdown()
 
   // Unregister this publisher from the ROS graph.
   z_liveliness_undeclare_token(z_move(token_));
+  // if (pub_cache_.has_value() && !z_session_is_closed(z_loan(sess_))) {
   if (pub_cache_.has_value()) {
-    z_drop(z_move(pub_cache_.value()));
+    ze_undeclare_publication_cache(z_move(pub_cache_.value()));
+    // if (z_session_is_closed(z_loan(sess_))) {
+    //   z_drop(z_move(pub_cache_.value()));
+    // }
```
clalancette:
```suggestion
```

clalancette (changes requested):
I generally like the idea here, but I think we need to do some more work particularly in the `ClientData` class.  I've left a detailed proposal in there.

ahcorde (changes requested):
These changes are already included in the zenoh_c to zenoh_cpp PR https://github.com/ros2/rmw_zenoh/pull/327/files#diff-124e1086db3e5f3edb76619bd65943a900a963ef671e014c817d3bf27d803908

clalancette:
> These changes are already included in the zenoh_c to zenoh_cpp PR https://github.com/ros2/rmw_zenoh/pull/327/files#diff-124e1086db3e5f3edb76619bd65943a900a963ef671e014c817d3bf27d803908

I do think we should get this one in first, and then go for the switch to zenoh-cpp.  That will allow us to concentrate on one thing at a time, and I think this one is largely ready to go (with my one big comment up above still to be addressed).

YuanYuYuan:
> > These changes are already included in the zenoh_c to zenoh_cpp PR https://github.com/ros2/rmw_zenoh/pull/327/files#diff-124e1086db3e5f3edb76619bd65943a900a963ef671e014c817d3bf27d803908
> 
> I do think we should get this one in first, and then go for the switch to zenoh-cpp. That will allow us to concentrate on one thing at a time, and I think this one is largely ready to go (with my one big comment up above still to be addressed).

Hi @clalancette, I'm trying to understand the underlying issue with the client and the request-in-flight. But I'm wondering why we must clone a session for each request. It seems to me either the callback or the `client_data_drop` doesn't utilize the session.

clalancette:
> Hi @clalancette, I'm trying to understand the underlying issue with the client and the request-in-flight. But I'm wondering why we must clone a session for each request. It seems to me either the callback or the `client_data_drop` doesn't utilize the session.

Doesn't `client_data_drop` implicitly use the session, by the fact that it being called by the internals of Zenoh?  If that is not the case, then indeed we do not need to hold onto this here.  But in that case, I think we would want to hold onto the session for the lifetime of `ClientData` (i.e. take the reference in the constructor).

YuanYuYuan:
> Doesn't `client_data_drop` implicitly use the session, by the fact that it being called by the internals of Zenoh? If that is not the case, then indeed we do not need to hold onto this here. But in that case, I think we would want to hold onto the session for the lifetime of `ClientData` (i.e. take the reference in the constructor).

1. The `client_data_drop` is part of the closure. It doesn't really depend on a zenoh session but a zenoh runtime which is global and internal.
2. But whether the callback is triggered or not makes a difference. I ran some quick tests and discovered that in the recent zenoh version, a get request sent to a remote queryable won't trigger the callback if the get session is closed between the time the queryable receives the request and sends back the reply. The question is do we want to make sure each reply is handled even after the session is closed? If the answer is yes, then we need to clone the session for every request. Otherwise, we don't.
3. We recently introduced the `Querier`, which works like a `session.put` vs `publisher`. This means we can keep the `Querier` rather than a `session` as we do for the other rmw data in the future.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
YuanYuYuan: Thanks. Addressed in https://github.com/ros2/rmw_zenoh/pull/333/commits/0c27cabbec1d3646791f2998b3d4e17074a81cb5

clalancette:
> * But whether the callback is triggered or not makes a difference. I ran some quick tests and discovered that in the recent zenoh version, a get request sent to a remote queryable won't trigger the callback if the get session is closed between the time the queryable receives the request and sends back the reply. The question is do we want to make sure each reply is handled even after the session is closed? If the answer is yes, then we need to clone the session for every request. Otherwise, we don't.

No, I don't think we need to handle the reply after the session is closed.  In our case, the session being closed is the result of the context being shutdown, and after that point nothing in the higher layers can handle anything new anyway.

OK, so in that case, I'm going to withdraw my previous comment about fixing it (though I will open a separate PR doing that for the `ClientData`, which does need to be kept alive).  I'll put a different comment about possibly moving the Session reference count elsewhere.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
clalancette: As discussed in https://github.com/ros2/rmw_zenoh/pull/333#issuecomment-2528971056 , we don't necessarily need this.  So resolving this particular conversation.

clalancette: I don't think we need this, right?  Because we are holding onto the session while this `ClientData` is alive, this is an unnecessary additional reference count, and I think it is enough to just keep this in the constructor.

clalancette (changes requested):
One more change to make, then I think this is good to go.

rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
YuanYuYuan: Thanks. It's misplaced. Resolved in https://github.com/ros2/rmw_zenoh/pull/333/commits/416f4303c1ef9ed739bb6adcec961b760829ad23

YuanYuYuan:
Rebase it against the latest rolling.

clalancette (approved):

PR #336, opened 2024-12-10
Another fix for ClientData.

clalancette:
In this fix, what we do is to manually allocate a struct that wraps the ClientData.  Inside of that struct, we take a `shared_ptr` reference to `ClientData`.  And then in `client_data_drop`, we manually delete the struct.

This ensures that the `ClientData` object is not destroyed while there are still requests in-flight.  This also allows us to get rid of the tricky tracking of the `num_in_flight`; we don't need it anymore, and the `rmw_node_data_t` can just drop the reference directly.

Finally, I'll note that this is basically equivalent to do a lambda capture when zenoh-cpp.  Still, I think this is worthwhile getting in before #327 since it fixes UB and it is relatively easy to understand.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp
@@ -43,11 +43,21 @@
 namespace
 {
 
+struct ClientDataWrapper
+{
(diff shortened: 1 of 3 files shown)


Yadunund (approved):
This is a much cleaner approach with the existing implementation!

PR #340, opened 2024-12-16
Fix the high latency on the subscriber due to zenoh bytes conversion

YuanYuYuan:
Recently, higher latency has been observed on the zenoh-cpp porting. 
The reason is doing zenoh bytes conversion in `take_one_message` is way slower than doing so in the subscriber callback.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
@@ -287,7 +287,6 @@ bool SubscriptionData::init()
     zenoh::Subscriber<void> sub = context_impl->session()->declare_subscriber(
       sub_ke,
       [data_wp](const zenoh::Sample & sample) {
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -224,7 +225,7 @@ bool SubscriptionData::init()
 
         sub_data->add_new_message(
           std::make_unique<SubscriptionData::Message>(
-            sample.get_payload().clone(),
+            sample.get_payload().as_vector(),
```
clalancette, line 228: Hm, I have a serious concern about this, but I'll leave a comment in the other PR since it is already merged.

YuanYuYuan: Sure. Also answered in that PR :smile:.

PR #343, opened 2024-12-16
Don't copy contiguous bytes on reception

fuzzypixelz:
This uses the slices iterator API of zenoh-c to avoid unecessarily copying bytes into an owned slice, if and only if the bytes is made up of exactly one slice.


fuzzypixelz:
The aim of this pull request is to fix high latency for relatively large topics (e.g. 1 MB `tagus`) in the single-process iRobot benchmark with `--ipc off`. Results were obtained using the Mont Blanc topology on an **Ubuntu 24.04.1 LTS** machine with an **Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz**.

# dd82e84
| node | topic | size_b | mean_us | min_us | max_us |
|--------|--------|--------|--------|--------|--------|
| ponce | tagus | 1000000 | 3016 | 2512 | 5705 |
| geneva | tagus | 1000000 | 3513 | 2997 | 6457 |
| mandalay | tagus | 1000000 | 2678 | 2214 | 5233 |

# eed223a
| node | topic | size_b | mean_us | min_us | max_us |
|--------|--------|--------|--------|--------|--------|
| ponce | tagus | 1000000 | 1651 | 1363 | 2252 |
| geneva | tagus | 1000000 | 1875 | 1557 | 2563 |
| mandalay | tagus | 1000000 | 1390 | 1151 | 1867 |

# eed223a + 401016c
| node | topic | size_b | mean_us | min_us | max_us |
|--------|--------|--------|--------|--------|--------|
| ponce | tagus | 1000000 | 393 | 257 | 1301 |
| geneva | tagus | 1000000 | 657 | 438 | 1701 |
| mandalay | tagus | 1000000 | 108 | 68 | 650 |

clalancette:
Thanks for opening the PR.

We are just in the process of converting to zenoh-cpp as our API in #327.  I'm assuming a similar idea will work there, but until we merge that, I'm going to mark this as a draft PR, and we can revisit it once we've merged that in.

fuzzypixelz:
> Thanks for opening the PR.
> 
> We are just in the process of converting to zenoh-cpp as our API in #327. I'm assuming a similar idea will work there, but until we merge that, I'm going to mark this as a draft PR, and we can revisit it once we've merged that in.

The bytes slice iterator API exists in zenoh-cpp, however there is no `z_bytes_to_slice` equivalent so we would need to manually allocate a buffer in that case or just use `as_vector`. We could coordinate potential changes to the zenoh-cpp API if necessary.

clalancette:
All right, we've now merged in #327.  Please rebase this and make any necessary changes to adapt to zenoh-cpp, and then we can do a full review of it.

fuzzypixelz:
The following are updated benchmark results after switching from the C to the C++ Zenoh API.

# df6ec26
| node | topic | size_b | mean_us | min_us | max_us |
|--------|--------|--------|--------|--------|--------|
| ponce | tagus | 1000000 | 1117 | 642 | 3399 |
| geneva | tagus | 1000000 | 1433 | 868 | 3775 |
| mandalay | tagus | 1000000 | 755 | 351 | 2807 |

# bf7c354
| node | topic | size_b | mean_us | min_us | max_us |
|--------|--------|--------|--------|--------|--------|
| ponce | tagus | 1000000 | 629 | 402 | 2343 |
| geneva | tagus | 1000000 | 873 | 611 | 2674 |
| mandalay | tagus | 1000000 | 304 | 160 | 1943 |

fuzzypixelz:
> All right, we've now merged in #327. Please rebase this and make any necessary changes to adapt to zenoh-cpp, and then we can do a full review of it.

Should be ready for review now.

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -44,10 +44,10 @@ namespace rmw_zenoh_cpp
 {
 ///=============================================================================
 SubscriptionData::Message::Message(
-  std::vector<uint8_t> && p,
```
clalancette, line 47: Since we are removing the use of `std::vector` from this file, please also remove the `#include <vector>`.

rmw_zenoh_cpp/src/detail/rmw_subscription_data.hpp
```diff
@@ -51,13 +52,13 @@ class SubscriptionData final : public std::enable_shared_from_this<SubscriptionD
   struct Message
   {
     explicit Message(
-      std::vector<uint8_t> && p,
```
clalancette, line 54: Since we are removing the use of `std::vector` from this file, please also remove the `#include <vector>`.

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,88 @@
...
+namespace rmw_zenoh_cpp
+{
+///=============================================================================
+class Payload
+{
+public:
+  explicit Payload(const zenoh::Bytes & bytes)
+  {
+    auto slices = bytes.slice_iter();
+    auto slice = slices.next();
```
clalancette: Could we use real types here instead of `auto`?  I think it will be easier to read.

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,88 @@
...
+    if (!slice.has_value()) {
+      bytes_ = nullptr;
+    } else {
+      if (!slices.next().has_value()) {
+        bytes_ = Contiguous {slice.value(), bytes.clone()};
```
clalancette: So I'm looking at the definition of `zenoh::Slice` in https://github.com/eclipse-zenoh/zenoh-cpp/blob/e8eca99b4eaff0963e52aefd8405909c1552080d/include/zenoh/api/bytes.hxx#L44-L47 .  It looks to me like this is a shallow copy, and thus we are just copying the `data` pointer and `len` here.  Is this safe to do?  How are we guaranteeing that the data that is pointed to by the `data` pointer still exists later on when we try to access it?

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DETAIL__PAYLOAD_HPP_
+#define DETAIL__PAYLOAD_HPP_
+
+#include <zenoh.h>
```
clalancette:
```suggestion
#include <zenoh.hxx>
```

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <variant>
+#include <utility>
+#include <vector>
```
clalancette:
```suggestion
#include <utility>
#include <variant>
#include <vector>
```

clalancette (changes requested):
Besides what I put inline, I'd also prefer not to have the implementation in the header file.  That means that any downstream consumer has to compile the code all of the time.  My suggestion is to just put this whole thing in the `rmw_subscription_data.cpp` file, but if you want to keep it a separate file then I'd suggest keeping just the API in the .hpp file and putting the implementation in a .cpp file.

rmw_zenoh_cpp/src/detail/payload.hpp
fuzzypixelz: That is precisely why we keep a clone of `bytes`, ensuring that the `Bytes` reference count maintained by Zenoh wouldn't go to zero and invalidate the slice.

clalancette: Right, but what's the relationship between the pointer returned by `bytes.clone()`, and `slice.data`?  It looks to me like in the `Contiguous` case, we clone the bytes, but then return the data from the `slice.data` later anyway.

fuzzypixelz: These are roughly the definitions on the Rust side:

```rust
pub struct ZBytes(ZBuf);

pub struct ZBuf {
    slices: SingleOrVec<ZSlice>,
}

pub struct ZSlice {
    buf: Arc<dyn ZSliceBuffer>,
    start: usize,
    end: usize,
    #[cfg(feature = "shared-memory")]
    pub kind: ZSliceKind,
}
```

So `Bytes` is roughly a list of reference-counted buffers. We only need to maintain a non-zero the reference count of the `Arc` to ensure that `slice.data` stays valid, as it points directly to the buffer behind the `Arc`.

See:
- https://github.com/eclipse-zenoh/zenoh/blob/a6bc509b5997c922a56253455149858af715edfa/commons/zenoh-buffers/src/zslice.rs#L92C1-L98C2
- https://github.com/eclipse-zenoh/zenoh/blob/a6bc509b5997c922a56253455149858af715edfa/commons/zenoh-buffers/src/zbuf.rs#L32-L34
- https://github.com/eclipse-zenoh/zenoh/blob/a6bc509b5997c922a56253455149858af715edfa/zenoh/src/api/bytes.rs#L93

fuzzypixelz:
> Besides what I put inline, I'd also prefer not to have the implementation in the header file. That means that any downstream consumer has to compile the code all of the time. My suggestion is to just put this whole thing in the `rmw_subscription_data.cpp` file, but if you want to keep it a separate file then I'd suggest keeping just the API in the .hpp file and putting the implementation in a .cpp file.

I will keep a separate translation unit for `Payload` as it might be useful for queries as well.

rmw_zenoh_cpp/src/detail/payload.hpp
clalancette: Thanks for the explanation.

From what I can tell of your explanation, then, we need to hold onto the `zenoh::Bytes` object as the reference count, and then that keeps the `zenoh::Slice` buffer around until we drop the reference later.  Do I have that right?

Assuming that is the case, I'm going to suggest adding a comment to that effect, as it is very non-obvious in this code.  It might even be worth renaming the `bytes` member of the structure to `ref` as well, but I'll leave that up to you.

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,88 @@
+// Copyright 2024 Open Source Robotics Foundation, Inc.
```
Yadunund: Can we move this definition to the existing `detail/zenoh_utils.hpp` header and have the implementation defined in `detail/zenoh_utils.cpp`?

Yadunund (changes requested):

rmw_zenoh_cpp/src/detail/payload.hpp
fuzzypixelz: See https://github.com/ros2/rmw_zenoh/pull/343#issuecomment-2551797597.

rmw_zenoh_cpp/src/detail/payload.hpp
fuzzypixelz: Please see if the comment I added in e76073831c078012343e315cf3564be68fa02489 is satisfactory.

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,96 @@
...
+        bytes_ = Contiguous {slice.value(), bytes.clone()};
+      } else {
+        bytes_ = bytes.as_vector();
+      }
+    }
+  }
+
+  ~Payload() = default;
+
+  const uint8_t * data()
```
ahcorde: include `<cstdint>`

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,96 @@
...
+    // NOTE(fuzzypixelz): `zenoh::Bytes` is an list of reference-couted buffers. When the list of
+    // buffers contains exactly one element, it is not necessary to concatenate the list of buffers.
+    // In this case, we store a clone of the bytes object to maintain a non-zero reference-count on
+    // the buffer. This ensures that the slice into said buffer stays valid until we drop our copy
+    // of the bytes object (at the very least). This case corresponds to the `Contiguous`
+    // alternative of the `bytes_` variant and aims to optimize away a memcpy during "session-local"
+    // communication.
+
+    zenoh::Bytes::SliceIterator slices = bytes.slice_iter();
+    std::optional<zenoh::Slice> slice = slices.next();
```
ahcorde: include `<optional>`

rmw_zenoh_cpp/src/detail/payload.hpp
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  ~Payload() = default;
+
+  const uint8_t * data()
+  {
+    if (std::holds_alternative<Empty>(bytes_)) {
+      return nullptr;
+    } else if (std::holds_alternative<NonContiguous>(bytes_)) {
+      return std::get<NonContiguous>(bytes_).data();
```
ahcorde: include `<tuple>`

ahcorde (changes requested):

rmw_zenoh_cpp/src/detail/payload.hpp
fuzzypixelz: [This is `std::get(std::variant)` and is defined in `variant`.](https://en.cppreference.com/w/cpp/utility/variant/get)

rmw_zenoh_cpp/src/detail/payload.hpp
fuzzypixelz: Nevermind.

rmw_zenoh_cpp/src/detail/payload.hpp
clalancette: That comment looks great, though I think it got lost in the recent refactor.

YuanYuYuan:
A little suggestion: The `Payload` name seems to be too generic and we don't explain its behavior (which is a copy-on-write when making the bytes contiguous). Can we give it a proper name like `CowPayload` or `CowBytes`?

fuzzypixelz:
79fc5ce304fa4ef9a40ab497e71d8a12d91a2449 seems to have broken the Jazzy build.

clalancette (approved):
I'm happy with this now.  Thanks for the revisions, I'm going to go ahead and merge this.

Yadunund:
@Mergifyio backport humble jazzy

PR #344, opened 2024-12-17
Updates to the C++ code

clalancette:
This is a series of relatively small updates to the zenoh-cpp port.  In particular:

1.  Change the storage of the GID from a std::vector to a std::array, since it is always a fixed size.
2.  Remove the now completely unnecessary ClientData init method; we can do all of the work in the constructor.
3.  Do a bunch of really small refactorings.
4.  Fix a bad string print in rmw_service_data.cpp
5.  Rewrite how we connect to the router so that it only prints when we fail.

With this in place, I'm now happy with the state of #327.  @ahcorde FYI


ahcorde (approved):

ahcorde:
thank you for the updates :+1:

PR #348, opened 2024-12-19
Remove ZenohSession since it's not used now.

evshary:
Since we are using zenoh-cpp now, the class ZenohSession wrapper is not used anymore. I think it was missed while migrating.


YuanYuYuan:
Hi @evshary, we still need it since we want to `z_close` the session if it's the last holder. Otherwise, we need to customize the `std::shared_ptr` deleter.

ahcorde (approved):
The class is never instantiated, I think it's safe to remove this

clalancette (approved):
Yes, agreed, we don't need this anymore.

Issue #350, opened 2024-12-19
Missing <functional> include when building on 22.04


christophebedard:
I rebased #294 on `rolling` yesterday (Dec 18th, potentially after #347), and then `rmw_zenoh_cpp` started failing to build on my 22.04 system:

<details>

<summary>Click to expand</summary>

```
Starting >>> rmw_zenoh_cpp
--- stderr: rmw_zenoh_cpp                                      
In file included from /usr/include/c++/11/bits/move.h:57,
                 from /usr/include/c++/11/bits/stl_pair.h:59,
                 from /usr/include/c++/11/utility:70,
                 from /usr/include/c++/11/array:38,
                 from /home/christophe.bedard/ros2_ws/src/rmw_zenoh/rmw_zenoh_cpp/src/detail/attachment_helpers.cpp:15:
/usr/include/c++/11/type_traits: In instantiation of ‘struct std::__is_trivially_copy_constructible_impl<std::reference_wrapper<zenoh::ZShmMut>, true>’:
/usr/include/c++/11/type_traits:1254:12:   required from ‘struct std::is_trivially_copy_constructible<std::reference_wrapper<zenoh::ZShmMut> >’
/usr/include/c++/11/type_traits:3200:43:   required from ‘constexpr const bool std::is_trivially_copy_constructible_v<std::reference_wrapper<zenoh::ZShmMut> >’
/usr/include/c++/11/optional:662:11:   required from ‘class std::optional<std::reference_wrapper<zenoh::ZShmMut> >’
/home/christophe.bedard/ros2_ws/install/zenoh_cpp_vendor/opt/zenoh_cpp_vendor/include/zenoh/api/shm/buffer/zshm.hxx:67:83:   required from here
/usr/include/c++/11/type_traits:1249:25: error: invalid use of incomplete type ‘class std::reference_wrapper<zenoh::ZShmMut>’
 1249 |                         __is_trivially_constructible(_Tp, const _Tp&)>>
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/11/type_traits:48:11: note: declaration of ‘class std::reference_wrapper<zenoh::ZShmMut>’
   48 |     class reference_wrapper;
      |           ^~~~~~~~~~~~~~~~~
[...]

PR #351, opened 2024-12-19
Fixed build

ahcorde:
Related with this issue https://github.com/ros2/rmw_zenoh/issues/350

FYI @christophebedard

### DIFF
--- a/rmw_zenoh_cpp/src/detail/attachment_helpers.cpp
+++ b/rmw_zenoh_cpp/src/detail/attachment_helpers.cpp
@@ -15,6 +15,7 @@
 #include <array>
 #include <cstdlib>
 #include <cstring>
+#include <functional>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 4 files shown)


christophebedard:
I think this should go in zenoh instead of `rmw_zenoh_cpp`.

clalancette:
> I think this should go in zenoh instead of `rmw_zenoh_cpp`.

Yeah, agreed, it looks like probably zshm.hxx in `zenoh-cpp` needs to have this included.

I'm going to leave this open as a draft for now, but we'll probably close it assuming that upstream takes a patch.  @ahcorde do you mind submitting a patch to https://github.com/eclipse-zenoh/zenoh-cpp to add it to the `zshm.hxx` file?

ahcorde:
closing this one in favour of https://github.com/eclipse-zenoh/zenoh-cpp/pull/342

PR #354, opened 2024-12-20
Switch to using get_system_time_in_ns() function everywhere.

clalancette:
That way we have just a single place where this is defined, and everywhere can be sure they are getting the correct value in nanoseconds.


Yadunund (approved):

Yadunund:
@Mergifyio backport humble jazzy

PR #355, opened 2024-12-21
Do not include service/client tracepoints for Jazzy

christophebedard:
Follow-up to #294

As reported on that PR (https://github.com/ros2/rmw_zenoh/pull/294#issuecomment-2557968743), Jazzy builds are broken, because the service/client tracepoints do not exist on Jazzy: `rmw_client_init`, `rmw_send_request`, `rmw_take_request`, `rmw_send_response`, `rmw_take_response` (added in Rolling in https://github.com/ros2/ros2_tracing/pull/145).

I added a generated header with some distro-specific `#define`s based on `ros_environment`, and then used it to exclude the service/client tracepoints on Jazzy.

### DIFF
--- a/rmw_zenoh_cpp/CMakeLists.txt
+++ b/rmw_zenoh_cpp/CMakeLists.txt
@@ -72,6 +72,22 @@ target_link_libraries(rmw_zenoh_cpp
     zenohcxx::zenohc
 )
 
+# Get distro from ros_environment and create config header
+set(ROS_DISTRO "$ENV{ROS_DISTRO}")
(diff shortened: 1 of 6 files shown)


christophebedard:
This fixes Jazzy builds. Feel free to not take this (or request changes) if you prefer another solution; I just wanted to propose a PR to fix what I broke :grin:

Yadunund:
Thanks @christophebedard! We've decided to leave the `rolling` branch as is. We've branched off `jazzy` and `humble` branches with the commit before #294 was merged and will selectively backport changes in the future. So we can close out this PR.

PR #356, opened 2024-12-23
Remove jazzy CI from rolling branch

Yadunund:
Following #294, we've decided to create distro specific branches in this repo, eg. `jazzy` and `humble`.

This PR removes `jazzy` jobs from running on `rolling` branch.

PR #358, opened 2024-12-23
README: add '-b <DISTRO>' to git clone instruction

JEnoch:
Now that there are [separate branches for each distribution](https://github.com/ros2/rmw_zenoh/pull/355#issuecomment-2558928701) and the `rolling` branch can no longer be compiled on Jazzy, the build instructions need to be updated to guide users to clone the correct branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ See [zenoh_cpp_vendor/CMakeLists.txt](./zenoh_cpp_vendor/CMakeLists.txt) for mor
 
 ```bash
 mkdir ~/ws_rmw_zenoh/src -p && cd ~/ws_rmw_zenoh/src
-git clone https://github.com/ros2/rmw_zenoh.git
+git clone https://github.com/ros2/rmw_zenoh.git -b <DISTRO>
 cd ~/ws_rmw_zenoh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ See [zenoh_cpp_vendor/CMakeLists.txt](./zenoh_cpp_vendor/CMakeLists.txt) for mor
 
 ```bash
 mkdir ~/ws_rmw_zenoh/src -p && cd ~/ws_rmw_zenoh/src
-git clone https://github.com/ros2/rmw_zenoh.git
+git clone https://github.com/ros2/rmw_zenoh.git -b <DISTRO>
```
Timple, line 26: If you make it `-b $ROS_DISTRO` it's just as readable but also copy-pastable.

JEnoch: I agree! I was just complying with the other instructions below which are using `<DISTRO>` instead of `$ROS_DISTRO`.

@Yadunund any reason for not having used `$ROS_DISTRO` in the very first version of this README ?

Yadunund: Probably because we don't source the distro till a couple steps after so the `ROS_DISTRO` envar will not be set at this point.

Anyways, I have the `-b` flag added in this PR https://github.com/ros2/rmw_zenoh/pull/356 so we can close this PR out.

Issue #326, opened 2024-12-03
[Test Failure] rclcpp_action / test_client


YuanYuYuan:
## Info

- rmw_zenoh 1.0: https://github.com/ZettaScaleLabs/rmw_zenoh/commit/f5efd0216095313f2f859256904578e137eafd2c
- rmw_zenoh 0.11: https://github.com/ros2/rmw_zenoh/commit/8306a63313cd9e21883344deff9abc782636a464

## To reproduce 
1. Build the rclcpp_action
2. Run the specific command
```bash
/workspace/build/rclcpp_action/test_client --gtest_filter="TestClientAgainstServer.*"
```
## Backtrace
```txt
#0  0x00007f8c871ca5c4 in std::__shared_ptr<rmw_context_impl_s::Data, (__gnu_cxx::_Lock_policy)2>::get (this=0x2f6e6f697463615f)
    at /usr/include/c++/13/bits/shared_ptr_base.h:1666
#1  0x00007f8c871cb3b8 in std::__shared_ptr_access<rmw_context_impl_s::Data, (__gnu_cxx::_Lock_policy)2, false, false>::_M_get (this=0x2f6e6f697463615f)
    at /usr/include/c++/13/bits/shared_ptr_base.h:1363
#2  0x00007f8c871ca4fe in std::__shared_ptr_access<rmw_context_impl_s::Data, (__gnu_cxx::_Lock_policy)2, false, false>::operator-> (this=0x2f6e6f697463615f)
    at /usr/include/c++/13/bits/shared_ptr_base.h:1357
#3  0x00007f8c871c776d in rmw_context_impl_s::get_node_data (this=0x2f6e6f697463615f, node=0x5646b9d54300)
    at /workspace/src/rmw_zenoh/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp:538
#4  0x00007f8c871c2576 in rmw_zenoh_cpp::ClientData::decrement_in_flight_and_conditionally_remove (this=0x5646b9d4eae0)
    at /workspace/src/rmw_zenoh/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp:555
#5  0x00007f8c871c004f in (anonymous namespace)::client_data_drop (data=0x5646b9d4eae0)
    at /workspace/src/rmw_zenoh/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp:101
[...]

PR #368, opened 2024-12-31
feat: introduce the advanced publisher and subscriber

YuanYuYuan:
With this PR, we replace the `QueryingSubscriber` and `PublicationCache` (+ a normal publisher)
with the `AdvancedSubscriber` and `AdvancedPublisher`.

NOTE: A deadlock is found when undeclaring advanced subscribers. It will be fixed once https://github.com/eclipse-zenoh/zenoh/pull/1685 is merged and synced with zenoh-c.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.hpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.hpp
@@ -104,8 +104,6 @@ public:
   /// @brief Signature for a function that will be invoked by the GraphCache when a QoS
   ///   event is detected.
   using GraphCacheEventCallback = std::function<void (int32_t change)>;
-  /// Callback to be triggered when a publication cache is detected in the ROS Graph.
-  using QueryingSubscriberCallback = std::function<void (const std::string & queryable_prefix)>;
 
   /// @brief Constructor
   /// @param id The id of the zenoh session that is building the graph cache.
@@ -278,9 +276,6 @@ private:
   using GraphEventCallbackMap = std::unordered_map<std::size_t, GraphEventCallbacks>;
   // EventCallbackMap for each type of event we support in rmw_zenoh_cpp.
   GraphEventCallbackMap event_callbacks_;
-  // Map key expressions to a map of sub keyexpr_hash and QueryingSubscriberCallback.
-  std::unordered_map<std::string, std::unordered_map<std::size_t,
-    QueryingSubscriberCallback>> querying_subs_cbs_;
   std::mutex events_mutex_;
 
   // Mutex to lock before modifying the members above.
--- a/rmw_zenoh_cpp/src/detail/graph_cache.hpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.hpp
@@ -189,15 +189,6 @@ public:
   /// Returns true if the entity is a publisher or client. False otherwise.
   static bool is_entity_pub(const liveliness::Entity & entity);
 
-  void set_querying_subscriber_callback(
-    const std::string & sub_keyexpr,
-    const std::size_t sub_keyexpr_hash,
-    QueryingSubscriberCallback cb);
-
-  void remove_querying_subscriber_callback(
-    const std::string & sub_keyexpr,
-    const std::size_t sub_keyexpr_hash);
-
 private:
   // Helper function to convert an Entity into a GraphNode.
   // Note: this will update bookkeeping variables in GraphCache.
--- a/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
@@ -115,16 +115,19 @@ std::shared_ptr<PublisherData> PublisherData::make(
     adv_pub_opts.publisher_detection = true;
   }
 
-  if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL
-    || adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
+  if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL ||
+    adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE)
+  {
     adv_pub_opts.cache = AdvancedPublisherOptions::CacheOptions::create_default();
     adv_pub_opts.cache->max_samples = adapted_qos_profile.depth;
   }
 
   if (adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
     // Allow matching Subscribers to detect lost samples and ask for retransimission.
-    adv_pub_opts.sample_miss_detection = AdvancedPublisherOptions::SampleMissDetectionOptions::create_default();
-    // The period of publisher heartbeats in ms, used by ``AdvancedSubscriber`` with heartbeat-based recovery enabled for missed sample retransimission.
+    adv_pub_opts.sample_miss_detection =
+      AdvancedPublisherOptions::SampleMissDetectionOptions::create_default();
+    // The period of publisher heartbeats in ms, used by ``AdvancedSubscriber`` with
+    // heartbeat-based recovery enabled for missed sample retransimission.
     adv_pub_opts.sample_miss_detection->heartbeat_period_ms = 1000;
   }
 
--- a/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
@@ -107,25 +107,25 @@ std::shared_ptr<PublisherData> PublisherData::make(
     return nullptr;
   }
 
-  auto adv_pub_opts = zenoh::ext::SessionExt::AdvancedPublisherOptions::create_default();
+  using AdvancedPublisherOptions = zenoh::ext::SessionExt::AdvancedPublisherOptions;
+  auto adv_pub_opts = AdvancedPublisherOptions::create_default();
 
-  // Create a Publication Cache if durability is transient_local.
   if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL) {
-    // Retransmission can only be done if history is enabled on subscriber side.
+    // Allow this publisher to be detected through liveliness.
     adv_pub_opts.publisher_detection = true;
   }
 
   if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL
     || adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
-    adv_pub_opts.cache =
-      zenoh::ext::SessionExt::AdvancedPublisherOptions::CacheOptions::create_default();
+    adv_pub_opts.cache = AdvancedPublisherOptions::CacheOptions::create_default();
     adv_pub_opts.cache->max_samples = adapted_qos_profile.depth;
   }
 
   if (adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
-    // Allow this publisher to be detected through liveliness.
-    adv_pub_opts.sample_miss_detection = true;
-    // Heartbeat
+    // Allow matching Subscribers to detect lost samples and ask for retransimission.
+    adv_pub_opts.sample_miss_detection = AdvancedPublisherOptions::SampleMissDetectionOptions::create_default();
+    // The period of publisher heartbeats in ms, used by ``AdvancedSubscriber`` with heartbeat-based recovery enabled for missed sample retransimission.
+    adv_pub_opts.sample_miss_detection->heartbeat_period_ms = 1000;
   }
 
   zenoh::KeyExpr pub_ke(entity->topic_info()->topic_keyexpr_);
--- a/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
@@ -113,11 +113,19 @@ std::shared_ptr<PublisherData> PublisherData::make(
   if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL) {
     // Retransmission can only be done if history is enabled on subscriber side.
     adv_pub_opts.publisher_detection = true;
-    // Allow this publisher to be detected through liveliness.
-    adv_pub_opts.sample_miss_detection = true;
+  }
+
+  if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL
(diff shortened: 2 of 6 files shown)


YuanYuYuan:
Update: The deadlock fix has been merged. Mark this PR as ready for review.

YuanYuYuan:
## Test Failure

[ros2.repos](https://github.com/Yadunund/ros2/commit/87d14774e8609d8201d9f15a8421081c628bb1d6)

- rclcpp_action
  - 1 - test_client
- rclc_parameter
  - 1 - rclc_parameter_test
- test_rclcpp.log
  - 74 - test_n_nodes__rmw_zenoh_cpp (Failed) <--- Flaky

No new failure is found.

YuanYuYuan:
Update: bump up zenoh-cpp version to include the memory leak fix https://github.com/eclipse-zenoh/zenoh-cpp/pull/363.

YuanYuYuan:
Will rebase once #405 is merged.

zenoh_cpp_vendor/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set(ZENOHC_CARGO_FLAGS "--no-default-features$<SEMICOLON>--features=shared-memor
 # - https://github.com/eclipse-zenoh/zenoh/pull/1696 (Fix SHM Garbage Collection (GC) policy)
 ament_vendor(zenoh_c_vendor
   VCS_URL https://github.com/eclipse-zenoh/zenoh-c.git
-  VCS_VERSION 61d8fcc136ce4ed36d921a32244da4f3d81a6097
+  VCS_VERSION 85ca060fa4037239ca4102a3a61f96626cc6b434
```
Yadunund: Please replace the comment above and below to include changes in this commit hash.

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -189,15 +189,6 @@ class GraphCache final
   /// Returns true if the entity is a publisher or client. False otherwise.
   static bool is_entity_pub(const liveliness::Entity & entity);
 
-  void set_querying_subscriber_callback(
```
Yadunund, line 192: You'll also need to delete `querying_subs_cbs_` and update `graph_cache.cpp` to not iterate over `querying_subs_cbs_`

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -107,40 +107,36 @@ std::shared_ptr<PublisherData> PublisherData::make(
     return nullptr;
   }
 
-  zenoh::ZResult result;
-  std::optional<zenoh::ext::PublicationCache> pub_cache;
-  zenoh::KeyExpr pub_ke(entity->topic_info()->topic_keyexpr_);
+  auto adv_pub_opts = zenoh::ext::SessionExt::AdvancedPublisherOptions::create_default();
+  adv_pub_opts.publisher_detection = true;
+  adv_pub_opts.sample_miss_detection = true;
```
Yadunund: Looking at the [documentation for this option](https://zenoh-cpp.readthedocs.io/en/latest/ext.html#_CPPv4N5zenoh3ext10SessionExt24AdvancedPublisherOptions21sample_miss_detectionE), it sounds like this should be set `true` only if `adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE`? Could you clarify how this differs from setting `pub_opts.reliability = Z_RELIABILITY_RELIABLE`?

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -107,40 +107,36 @@ std::shared_ptr<PublisherData> PublisherData::make(
...
-      zenoh::ext::SessionExt::PublicationCacheOptions::create_default();
-
-    pub_cache_opts.history = adapted_qos_profile.depth;
-    pub_cache_opts.queryable_complete = true;
-
-    std::string queryable_prefix = entity->zid();
-    pub_cache_opts.queryable_prefix = zenoh::KeyExpr(queryable_prefix);
-
-    pub_cache = session->ext().declare_publication_cache(
-      pub_ke, std::move(pub_cache_opts), &result);
```
Yadunund: We previously needed to do this to fix https://github.com/ros2/rmw_zenoh/issues/263. Can you confirm that we no longer need to explicitly call `QueryingSubscriber::get()` when we discover an AdvancedPublisher in the graph?

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -173,151 +173,60 @@ bool SubscriptionData::init()
...
-        "/" +
-        sub_data->entity_->topic_info().value().topic_keyexpr_;
-        RMW_ZENOH_LOG_DEBUG_NAMED(
+    adv_sub_opts.query_timeout_ms = std::numeric_limits<uint64_t>::max();
+    adv_sub_opts.history =
+      zenoh::ext::SessionExt::AdvancedSubscriberOptions::HistoryOptions::create_default();
+    adv_sub_opts.history->detect_late_publishers = true;
+    adv_sub_opts.history->max_samples = entity_->topic_info()->qos_.depth;
+    adv_sub_opts.recovery = zenoh::ext::SessionExt::AdvancedSubscriberOptions::RecoveryOptions{};
+    adv_sub_opts.recovery->periodic_queries_period_ms = 1000;
```
Yadunund: Again looking at the [documentation](https://zenoh-cpp.readthedocs.io/en/latest/ext.html#_CPPv4N5zenoh3ext10SessionExt25AdvancedSubscriberOptions15RecoveryOptions26periodic_queries_period_msE), it sounds like recovery options should only be enabled if reliability is `RELIABLE`. Could you clarify on the behavior?

Yadunund (changes requested):
Please see feedback inline.
I have the following additional questions

1. It looks like this PR does not fix #326. But from my experience, it changes the behavior from deterministically failing to failing sometimes. ie, the test is now flaky. Do you observe the same?
2. With `AdvancedSubscribers` do we no longer need to call `Get()`  when we detect an `AdvancedPublisher` with publication cache? This was needed to fix https://github.com/ros2/rmw_zenoh/issues/263.

zenoh_cpp_vendor/CMakeLists.txt
YuanYuYuan: Thanks. This was out of sync. We don't need to change it in this PR.

rmw_zenoh_cpp/src/detail/graph_cache.hpp
YuanYuYuan: Done in 9cc776bd8cc7fddfea8dec0ad703a22d3a6c96d3

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
YuanYuYuan: The reliability option can still be set by the normal publisher option. See [here](https://github.com/ZettaScaleLabs/rmw_zenoh/blob/9cc776bd8cc7fddfea8dec0ad703a22d3a6c96d3/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp?plain=1#L128). The reliability in zenoh is based on TCP reliability per see. The sample_miss_detection here is a custom data retransmission feature cached on the publisher side in the sense of transient_local duriability.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
YuanYuYuan: We should no longer need this hack. Otherwise, it's a bug in zenoh.

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
YuanYuYuan: Similarly, answered in https://github.com/ros2/rmw_zenoh/pull/368#discussion_r1912826044.

YuanYuYuan:
> Please see feedback inline. I have the following additional questions
>
>     1. It looks like this PR does not fix [[Test Failure] rclcpp_action / test_client #326](https://github.com/ros2/rmw_zenoh/issues/326). But from my experience, it changes the behavior from deterministically failing to failing sometimes. ie, the test is now flaky. Do you observe the same?
>
>     2. With `AdvancedSubscribers` do we no longer need to call `Get()`  when we detect an `AdvancedPublisher` with publication cache? This was needed to fix [Subscribers sometimes not getting messages from topics using durability transient_local #263](https://github.com/ros2/rmw_zenoh/issues/263).
Here’s an improved version of your paragraph with clearer phrasing and a more professional tone:

1. This PR does not resolve the issue reported in #326. Based on the investigation, the root cause lies in the inappropriate synchronization assumptions in the test. We have also observed similar failures (slightly more often) when using the previous `QueryingSubscriber` implementation. To address this issue, we can either make the local publisher-to-subscriber action a blocking call or modify the test itself.

2. We expect the issue will be resolved. So no longer need the hack :smile:

YuanYuYuan:
Redid the test with the [repos](https://github.com/Yadunund/ros2/commit/87d14774e8609d8201d9f15a8421081c628bb1d6). No new failures were found.

- rclcpp_action/test_client
- rclc_parameter/rclc_parameter_test

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
Yadunund: Could we add some comments on why a value of `1000` is appropriate for `periodic_queries_period_ms`?

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -173,151 +173,62 @@ bool SubscriptionData::init()
...
-    // This allows us to selectively query certain PublicationCaches when defining the
-    // set_querying_subscriber_callback below.
-    sub_options.query_accept_replies = ZC_REPLY_KEYEXPR_ANY;
-    // As this initial query is now using a "*", the query target is not COMPLETE.
-    sub_options.query_target = Z_QUERY_TARGET_ALL;
-    // We set consolidation to none as we need to receive transient local messages
-    // from a number of publishers. Eg: To receive TF data published over /tf_static
-    // by various publishers.
-    sub_options.query_consolidation =
-      zenoh::QueryConsolidation(zenoh::ConsolidationMode::Z_CONSOLIDATION_MODE_NONE);
```
Yadunund, line 192: And it does not look like there is a way to set these parameters anymore. Can you confirm that the behavior with the AdvancedSubscriber will match the expectations of the old implementation?

YuanYuYuan: Yes. They're preset in the same way https://github.com/eclipse-zenoh/zenoh/blob/26d67de043efe327befd0bbfb29c76a5157f1baf/zenoh-ext/src/advanced_subscriber.rs?plain=1#L615.

YuanYuYuan:
Now we set the advanced pub/sub to map the ROS2 QoS according to the following table.

- AdvPub

|Option                          |ROS2 QoS                              |
|-                               |-                                     |
|publisher_detection             |Durability::TransientLocal            |
|cache                           |Durability::TransientLocal or Reliable|
|cache->max_samples              |Depth                                 |
|sample_miss_detection           |Reliable                              |
|sample_miss_detection->heartbeat|Reliable                              |
|Z_RELIABILITY_RELIABLE          |Reliable                              |
|Z_CONGESTION_CONTROL_BLOCK      |Reliable & History::KeepALL           |

- AdvSub

|Option                                          |ROS2 QoS                  |
|-                                               |-                         |
|subscriber_detection                            |Durability::TransientLocal|
|sample_miss_detection                           |Durability::TransientLocal|
|history->detect_late_publishers                 |Durability::TransientLocal|
|history->max_samples                            |Depth                     |
|recovery                                        |Reliable                  |
|recovery->last_sample_miss_detection = heartbeat|Reliable                  |

Tested against https://github.com/Yadunund/ros2/commit/87d14774e8609d8201d9f15a8421081c628bb1d6. No regression was found.

Yadunund:
We discussed this today and we agreed to hold merging this PR until we test the changes here on more complex applications to gauge whether the constants adopted are appropriate.

Yadunund:
We discussed this again today and agreed to convert this to a draft PR until https://github.com/eclipse-zenoh/zenoh/pull/1753 is merged and zenoh version here is updated. Along with this change, we will also be disabling recovery and sample missed options as it does not fit the typical ROS use case.

YuanYuYuan:
The recovery option has been removed in https://github.com/ros2/rmw_zenoh/pull/368/commits/0e4b99d7363bcc61e9fa6431dec5c3a2380547d3

JEnoch:
I created #457 to address the possible usage of `recovery` and `sample_miss_detection` options.
Those options are only required to benefit end-to-end reliability.
They are not required to implement `TRANSIENT_LOCAL` QoS which is the goal of this PR, as a replacement of the deprecated QueryingSubscriber and Publication Cache.

As soon as #456 (that includes https://github.com/eclipse-zenoh/zenoh/pull/1753) is merged, I think this PR is good to go.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -107,40 +107,40 @@ std::shared_ptr<PublisherData> PublisherData::make(
...
 
-    if (result != Z_OK) {
-      RMW_SET_ERROR_MSG("unable to create zenoh publisher cache");
-      return nullptr;
-    }
+  if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL ||
+    adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE)
+  {
+    adv_pub_opts.cache = AdvancedPublisherOptions::CacheOptions::create_default();
+    adv_pub_opts.cache->max_samples = adapted_qos_profile.depth;
```
Yadunund: Based on this logic, a PubCache is created even if durability is volatile as long as reliability is reliable. Isn't this a change in behavior? Why do we need to check for reliability here?

Yadunund (changes requested):
Tested with latest zenoh bump and I can confirm that rclcpp_action/test_client passes 🎉 

However, I'm concerned with one change which i've flagged below.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
YuanYuYuan: It was needed if **end-to-end** reliability is required. But since we have decided to exclude it from this PR, let's roll back to its original setting.

Addressed in [6068096](https://github.com/ros2/rmw_zenoh/pull/368/commits/6068096797b90595545d1d6a73ecae1bd9bb7dfd)

Yadunund: Sounds good, I pushed some minor changes in https://github.com/ros2/rmw_zenoh/pull/368/commits/b1b2dd8a05f8f10d00807f3cb9bb468ea18cd0c3

Yadunund (approved):
LGTM! Thanks for iterating on this PR to find the right fix! We can finally close out #326

Yadunund:

PR #335, opened 2024-12-10
Backward Compatibility with ROS 2 Humble

YuanYuYuan:
## Description

This is a draft that backward supports Humble #303. It's currently based on #333.

## Status
- Basic pub/sub works.
- The workflow should be changed to Humble.
- rclcpp test failures
```txt
30 - test_intra_process_manager (Failed)
53 - test_parameter (Failed)
56 - test_publisher (Failed)
61 - test_qos (Failed)
64 - test_qos_event (Failed)
91 - test_guard_condition (Failed)
92 - test_wait_set (Failed)
```

PR #388, opened 2025-01-02
Updates to compile against Humble.

clalancette:
There are 5 main changes in here:
1.  Humble does not support type hashes.  Rather than change all of the liveliness token handling (which will make backports in the future harder), we just remove the code that computes it, and instead set it to a constant string.
2.  The definition of RMW_GID_STORAGE_SIZE in Humble is wonky, and set to 24 for some ancient historical reasons.  We don't actually want to do that, so just hard-code 16.
3.  Some of the events in modern ROS 2 aren't supported in Humble, so remove the code for handling them.
4.  There are a few APIs dealing with dynamic messages that aren't supported in Humble.
5.  There is at least one QOS policy (best available) which doesn't exist in Humble.

With all of these changes in, I'm able to compile and run basic programs against Humble.

This takes many of the changes from #335, cleans them up, and rebases them onto the latest.  Thanks @YuanYuYuan for starting that!


Yadunund (approved):
Thanks a lot for working on this!
Do you think it makes sense to include a `#define RMW_GID_STORAGE_SIZE 16u` somewhere as opposed to replace with `16`?

clalancette:
> Do you think it makes sense to include a `#define RMW_GID_STORAGE_SIZE 16u` somewhere as opposed to replace with `16`?

So the problem with that is the RMW_GID_STORAGE_SIZE is defined on Humble, but it is defined to 24.  So this would be a "multiple definition" problem during compilation.  We could change the name (`RMW_ZENOH_GID_STORAGE_SIZE` or something), but I'm not sure it is worth it.  What do you feel like?

Yadunund:
Ah nvm then, I think the changes here are great.

PR #395, opened 2025-01-07
Documentation for Zenoh configuration

### LINKED ISSUE #230: Document restricting Zenoh sessions and router to a specific network interface
This is important for certain use cases, like sensors which use ethernet.  In that case, the user has to be sure to *not* run Zenoh on those specific interfaces.

JEnoch:
The goal of this PR is to add a documentation about the way to configure Zenoh for various use cases and deployments.

So far I plan address the following sections:

- [x] Default configuration
- [ ] Explanations of `connect` and `listen` configuration parts (will close #230)
- [ ] Case of a robot with multiple hosts/boards (e.g. Turtlebot3)
- [ ] Connection from/to another host (i.e. router-to-router)
- [ ] Direct connection of a remote Node (e.g. remote `rviz`)
- [ ] Automatic discovery and interconnections of Zenoh routers (i.e. routers scouting)
- [ ] Remote connection via a Zenoh router in the cloud
- [x] No router, UDP multicast scouting, single host (i.e. DDS-like, single host)
- [ ] No router, UDP multicast scouting, multiple hosts (i.e. DDS-like, multi hosts)
- [ ] Downsampling
- [ ] Access Control (by network interface)
- [ ] Performances tuning (configuration of batching, express, priorities...)

Any comment or suggestion of additional section is welcome !


JEnoch:
Also, I'm not sure yet how to re-organize the main [README.md](https://github.com/ros2/rmw_zenoh/blob/rolling/README.md#configuration) to avoid duplicate and to point to this new doc...

boxanm:
Hello @JEnoch ! Thanks for taking the time to give us, users, more examples of using `zenoh-rmw `in different scenarios. I'm particularly interested in multi-host examples (router-to-router). Do you plan to include these in the near future? I understand that you target the next ROS release for a polished `zenoh-rmw` version. Having these configs available earlier would definitely help earlier adopters with the testing efforts. Thanks!

JEnoch:
Sure, will do! I haven't had time to make progress on this recently, but I will continue next week.
In the meantime, you can have a look here for explanation of router-to-router config:
https://github.com/ZettaScaleLabs/roscon2024_workshop/blob/main/exercises/ex-2.md

boxanm:
Hello again @JEnoch !
I checked the ROScon exercise you mentioned, and I believe it suffers from the same issues as most other DDS or ROS 2 comm related tutorials out there - it only considers an extremely simplified setting, in this case, sending small text messages between two containers on the same machine. I believe this is something that should be taken into account when writing the example configuration, as most users operate more complicated systems counting dozens of nodes, and send fairly large data chunks such as point clouds and camera images. For example, why the `chatter` string passes without any issues, we didn't succeed in sending a point cloud message over Zenoh between two computers directly connected over ethernet. Do you have any example configuration file that would support such a setting? Thanks!

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/rmw-zenoh-binaries-for-rolling-jazzy-and-humble/41395/8

siddux:
Hi @boxanm did you find any suitable configuration (or more information) on complex multiple host communication? I've been able to set it up but the CPU load is extremely high, much worse than on any other RMW, and specially if I have a node or router crashing.

PR #403, opened 2025-01-09
fix: check the context validity before accessing the session

YuanYuYuan:
This PR addresses the flaky failure in rclcpp/test_wait_for_message. The test sometimes behaves as a use-after-free, say use the session to create a subscription after the context is shut down.


Yadunund (approved):

Yadunund:

PR #411, opened 2025-01-10
Added Zenoh security configuration package

ahcorde:
@yadu and I chatted offline about having a tool to generate zenoh security configuration files. This is based on this other package https://github.com/osrf/nexus/tree/main/nexus_network_configuration

For the router is required to generate a config file and then setup the environment variable `ZENOH_ROUTER_CONFIG_URI`.

But for the nodes we can use two different paths:
 - Generate config files as well and the setup `ZENOH_SESSION_CONFIG_URI`
 - use `--ros-args --enclave` which requires to include some modification in the code to read and setup the Zenoh Config data structure from the code.

```bash
ros2 run zenoh_security_configuration zenoh_security_configuration -o zenoh_config -p tls
```

This command should generate a file in `zenoh_config/router.json5`. For now it's only setting up the protocols

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,15 @@ The `ZENOH_ROUTER_CHECK_ATTEMPTS` environment variable can be used to configure 
 The behavior is explained in the table below.
 
 
-| ZENOH_ROUTER_CHECK_ATTEMPTS |                                                 Session behavior                                                 |
-|:---------------------------:|:----------------------------------------------------------------------------------------------------------------:|
-|            unset or 0           |                                                             Indefinitely waits for connection to a Zenoh router. |
-|             < 0            |                                                                                        Skips Zenoh router check. |
+| ZENOH_ROUTER_CHECK_ATTEMPTS |                                                 Session behavior                                                   |
+|:---------------------------:|:------------------------------------------------------------------------------------------------------------------:|
+|               0             |                                                             Indefinitely waits for connection to a Zenoh router.   |
+|             < 0             |                                                                                        Skips Zenoh router check.   |
 |             > 0             | Attempts to connect to a Zenoh router in `ZENOH_ROUTER_CHECK_ATTEMPTS` attempts with 1 second wait between checks. |
+|            unset            |                                                                    Equivalent to `1`: the check is made only once. |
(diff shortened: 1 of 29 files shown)


Yadunund (changes requested):
Thanks for getting started on this config generator. 

Here's my high-level feedback:

1. The executable should parse the standard `.policy.xml` file as defined in `sros2` to generate the necessary Zenoh configs. Eg. [add_two_ints.policy.xml](https://github.com/ros2/sros2/blob/rolling/sros2/test/policies/add_two_ints.policy.xml). For DDS systems, this file is used to generate this [permissions.xml](https://github.com/ros2/sros2/blob/rolling/sros2/test/policies/permissions/add_two_ints/permissions.xml) but in our case we will parse it and generate router & session configs. 

2. After parsing the same `.policy.xml` file, we should also generate access control lists for different endpoints. The keyexpressions can be simplified using wildcards. eg: `0/chatter/**` to allow topic `chatter` over ros_domain_id `0`. We should also allow all liveliness keypressions over `@ros2_lv/0/**` where `@ros2_lv` is the fixed admin space and 0 is the ros_domain_id specified in the `.policy.xml`

3. I don't think we should have a copy of configs in `templates/` here and another set in `rmw_zenoh_cpp/config`. Syncing these configs as we bump zenoh versions will become a chore. We could move `rmw_zenoh_cpp/config` to the root of this repo and then add the config files to the build dir when building `rmw_zenoh_cpp` while also being able to use it here. Or we could also create a `rmw_zenoh_common` package to simply install the configs such that both `rmw_zenoh_cpp` and this package can retrieve the configs using `ament` utilities. Or if C++ has good libraries to work with xml and json5, we could just have this executable live in `rmw_zenoh_cpp` similar to `rmw_zenohd`.

Lastly, should we consider writing this executable in Rust? 🦀 
We can look at how upstream Zenoh parses/creates the configs with Rust and we could rely on [colcon-cargo](https://github.com/colcon/colcon-cargo) to build. Will double check if we can release such a package on the buildfarm but alternatively we can use CMakeLists.txt to invoke the Rust build.

PR #412, opened 2025-01-10
Config security using enclave ROS argument

ahcorde:
As I mentioned in this other PR https://github.com/ros2/rmw_zenoh/pull/411#issue-2780261789 we should be able to configure the security features using `--ros-args --enclave <enclave name>`.

This PR drafted the required changes in the code to be able to set them, note that the Zenoh Config should be configure before the session is created.

What do you think about the changes?

Another question, Up to this point, we only supported one kind of rmw: DDS, we are using a package called `rmw_dds_common` to get the certificates, etc, in this package which is not DDS based. I know it's just a naming thing, but do we need to create a more generic package name and move some stuff there? @clalancette @Yadunund ?


rmw_zenoh_cpp/CMakeLists.txt
```diff
@@ -21,9 +21,15 @@ find_package(rcutils REQUIRED)
 find_package(rosidl_typesupport_fastrtps_c REQUIRED)
 find_package(rosidl_typesupport_fastrtps_cpp REQUIRED)
 find_package(rmw REQUIRED)
+find_package(rmw_dds_common REQUIRED)
 find_package(tracetools REQUIRED)
 find_package(zenoh_cpp_vendor REQUIRED)
 
+if(SECURITY)
+  find_package(OpenSSL REQUIRED)
```
Yadunund: I don't see OpenSSL being linked anywhere? Does this happen transitively via `rmw_dds_common::rmw_dds_common_library`?

rmw_zenoh_cpp/CMakeLists.txt
```diff
@@ -21,9 +21,15 @@ find_package(rcutils REQUIRED)
...
+  set(HAVE_SECURITY 1)
```
Yadunund, line 31: Let's not have any special build flags for security. We will overwrite the Zenoh config with auth params if `security_options` is present in `rmw_context_impl_s`.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -52,7 +53,8 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
   // Constructor.
   Data(
     std::size_t domain_id,
-    const std::string & enclave)
+    const std::string & enclave,
+    const rmw_security_options_t * security_options)
```
Yadunund:
```suggestion
    const rmw_security_options_t & security_options)
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -67,7 +69,37 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
     if (!config.has_value()) {
       throw std::runtime_error("Error configuring Zenoh session.");
     }
-
+#ifdef HAVE_SECURITY
+    std::unordered_map<std::string, std::string> security_files_paths;
+    if (rmw_dds_common::get_security_files(
+        true, "", security_options->security_root_path, security_files_paths))
```
Yadunund:
```suggestion
        true, "", security_options.security_root_path, security_files_paths))
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -67,7 +69,37 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
     if (!config.has_value()) {
       throw std::runtime_error("Error configuring Zenoh session.");
     }
-
+#ifdef HAVE_SECURITY
```
Yadunund, line 72: Why do we need this flag? My understanding is that `rmw_dds_common::get_security_files` will return false if the enclave ROS arg is not passed or is invalid so we can have this codeblock that modifies for all situations. Let me know if I misunderstood something.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -67,7 +69,37 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
...
+        "\t\t\t\t\"tls\" \n"
+        "\t\t\t], \n"
+        "\t\t\t\"tls\": { \n"
+        "\t\t\t\t\"enable_mtls\": true, \n"
+        "\t\t\t\t\"verify_name_on_connect\": false, \n"
+        "\t\t\t\t\"root_ca_certificate\": \"" + security_files_paths["IDENTITY_CA"] + "\",\n" +
+        "\t\t\t\t\"listen_private_key\": \"" + security_files_paths["PRIVATE_KEY"] + "\",\n" +
+        "\t\t\t\t\"listen_certificate\": \"" + security_files_paths["CERTIFICATE"] + "\",\n" +
+        "\t\t\t\t\"connect_private_key\": \"" + security_files_paths["PRIVATE_KEY"] + "\",\n" +
+        "\t\t\t\t\"connect_certificate\": \"" + security_files_paths["CERTIFICATE"] + "\",\n" +
```
Yadunund: Let's check that these keys exist in `security_files_paths` before accessing their values. If not present, return an error.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -432,9 +464,10 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
 ///=============================================================================
 rmw_context_impl_s::rmw_context_impl_s(
   const std::size_t domain_id,
-  const std::string & enclave)
+  const std::string & enclave,
+  const rmw_security_options_t * security_options)
```
Yadunund:
```suggestion
  const rmw_security_options_t & security_options)
```

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.hpp
```diff
@@ -40,7 +40,8 @@ struct rmw_context_impl_s final
   // check has not succeeded.
   rmw_context_impl_s(
     const std::size_t domain_id,
-    const std::string & enclave);
+    const std::string & enclave,
+    const rmw_security_options_t * security_options);
```
Yadunund:
```suggestion
    const rmw_security_options_t & security_options);
```

rmw_zenoh_cpp/src/rmw_init.cpp
```diff
@@ -110,7 +110,8 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     return RMW_RET_BAD_ALLOC,
     rmw_context_impl_t,
     context->actual_domain_id,
-    std::string(options->enclave)
+    std::string(options->enclave),
+    &context->options.security_options
```
Yadunund:
```suggestion
    context->options.security_options
```

Yadunund (changes requested):
Thanks for working on this! Please find my feedback below.

Also, my preference is to not merge this PR while it still depends on `rmw_dds_common`. We only need one function from that library and I don't think we need to add a dependency for that reason especially since `rmw_sercurity_common` is in the works. Semantically, it also does not make sense for this middleware to depend on a DDS package. I see two paths forward:
1. Get `rmw_sercurity_common` merged upstream, backport that merge to Jazzy atleast, and release binaries for the package on Rolling and Jazzy. Then merge #434 into this one.
2. Temporarily copy in the definition and implementation of [rmw_dds_common::get_security_files](https://github.com/ros2/rmw_dds_common/blob/rolling/rmw_dds_common/include/rmw_dds_common/security.hpp) into this repo and use the function without linking depending on `rmw_dds_common`.

Personally I think we should do 2) first (in this PR) and then when 1) is ready, delete the internal `get_security_files` and switch to the one in `rmw_security_common`.

PR #416, opened 2025-01-13
Remove `gid_hash_` from `AttachmentData`

fuzzypixelz:
This member is computed in the constructor which is in turn called in `rmw_publish`. This hash is not serialized as part of the attachment.

I noticed this when profiling the multi-process iRobot benchmark, where `hash_gid` was taking around 23% of the running time of `rmw_publish`.


ahcorde (approved):

Yadunund (approved):
Ran the usual CI tests and no regressions.

Yadunund:

PR #419, opened 2025-01-15
fix: resolve the memory leak in the subcription message

YuanYuYuan:
A memory leak was caught due to the memory leak in the subscription message. This PR addresses this by using the default destructor to drop the member `Payload` automatically.


ahcorde (approved):

ahcorde:

YuanYuYuan:
Testing against the [ros.repos](https://github.com/Yadunund/ros2/commit/87d14774e8609d8201d9f15a8421081c628bb1d6), no regression was found.

PR #422, opened 2025-01-15
Replace `hash_gid` with FNV-1a

fuzzypixelz:
`hash_gid` is called on Zenoh subscriber callbacks and shows up when profiling as a hot function. The use of `std::stringstream` seems to be relatively inefficient. A significant portion of `hash_gid`'s running time is spend allocating & deallocating the `std::stringstream`.

Screenshot 2025-01-15 at 17 39 31

### DIFF
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
@@ -289,7 +289,7 @@ struct hash<rmw_zenoh_cpp::Gid>
       hash ^= byte;
       hash *= FNV_PRIME_64;
     }
-    return hash;
+    return static_cast<std::size_t>(hash);
   }
 };
 }  // namespace std
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
@@ -275,8 +275,9 @@ struct hash<rmw_zenoh_cpp::Gid>
 {
   std::size_t operator()(const rmw_zenoh_cpp::Gid & gid) const noexcept
   {
-    // This function implemented FNV-1a 64-bit as the GID is small enough (e.g. as of commit db4d05e
-    // of ros2/rmw RMW_GID_STORAGE_SIZE is set to 16) and FNV is known to be efficient in such cases.
+    // This function implemented FNV-1a 64-bit as the GID is small enough
+    // (e.g. as of commit db4d05e of ros2/rmw RMW_GID_STORAGE_SIZE is set to 16)
+    // and FNV is known to be efficient in such cases.
     //
     // See https://github.com/ros2/rmw/blob/db4d05e/rmw/include/rmw/types.h#L44
     // and https://datatracker.ietf.org/doc/html/draft-eastlake-fnv-17.html
--- a/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
+++ b/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
@@ -234,8 +234,7 @@ std::string qos_to_keyexpr(const rmw_qos_profile_t & qos);
 std::optional<rmw_qos_profile_t> keyexpr_to_qos(const std::string & keyexpr);
 }  // namespace liveliness
 
-///=============================================================================
-size_t hash_gid(const std::array<uint8_t, RMW_GID_STORAGE_SIZE> gid);
+using Gid = std::array<uint8_t, RMW_GID_STORAGE_SIZE>;
 }  // namespace rmw_zenoh_cpp
 
 ///=============================================================================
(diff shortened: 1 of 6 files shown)


fuzzypixelz:
We seem to have many uses of `std::unordered_map` where the hashing is done manually instead of relying on the `Hash` template parameter. I'll revisit that before marking this pull request as ready for review.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -270,6 +269,29 @@ struct equal_to<rmw_zenoh_cpp::liveliness::ConstEntityPtr>
...
+  std::size_t operator()(const rmw_zenoh_cpp::Gid & gid) const noexcept
+  {
+    // This function implemented FNV-1a 64-bit as the GID is small enough
+    // (e.g. as of commit db4d05e of ros2/rmw RMW_GID_STORAGE_SIZE is set to 16)
+    // and FNV is known to be efficient in such cases.
+    //
+    // See https://github.com/ros2/rmw/blob/db4d05e/rmw/include/rmw/types.h#L44
+    // and https://datatracker.ietf.org/doc/html/draft-eastlake-fnv-17.html
+    static constexpr uint64_t FNV_OFFSET_BASIS_64 = 0x00000100000001b3;
+    static constexpr uint64_t FNV_PRIME_64 = 0xcbf29ce484222325;
```
clalancette, line 285: So we are already using xxhash3 in [liveliness_utils](https://github.com/ros2/rmw_zenoh/blob/bda2d7181d7e1ea2f7d143fabf7c4ffc6c670b20/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L438-L439) to take a string and hash it into a 128-bit quantity.  Do we need to hash again?  In other words, could we just use the 128-bit GID directly and use that as the key?

clalancette (commented):
I left a question inline.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
fuzzypixelz: I believe [`std::hash`](https://en.cppreference.com/w/cpp/utility/hash) requires a `size_t`.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -270,6 +269,29 @@ struct equal_to<rmw_zenoh_cpp::liveliness::ConstEntityPtr>
...
+
+    uint64_t hash = FNV_OFFSET_BASIS_64;
+    for (const uint8_t & byte : gid) {
+      hash ^= byte;
+      hash *= FNV_PRIME_64;
+    }
+    return hash;
```
Yadunund:
```suggestion
    return static_cast<std::size_t>(hash);
```

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
```diff
@@ -270,6 +269,29 @@ struct equal_to<rmw_zenoh_cpp::liveliness::ConstEntityPtr>
...
+{
+  std::size_t operator()(const rmw_zenoh_cpp::Gid & gid) const noexcept
+  {
+    // This function implemented FNV-1a 64-bit as the GID is small enough
+    // (e.g. as of commit db4d05e of ros2/rmw RMW_GID_STORAGE_SIZE is set to 16)
+    // and FNV is known to be efficient in such cases.
+    //
+    // See https://github.com/ros2/rmw/blob/db4d05e/rmw/include/rmw/types.h#L44
+    // and https://datatracker.ietf.org/doc/html/draft-eastlake-fnv-17.html
+    static constexpr uint64_t FNV_OFFSET_BASIS_64 = 0x00000100000001b3;
```
Yadunund, line 284: Include `<cstdint>` for `uint64_t`

Yadunund (changes requested):
This is looking good to me. A couple minor nitpicks to address after which this is ready to merge.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
Yadunund: We talked about this today and one optimization to make is to rely on FNA-1a to store the hash of the liveliness string as a `std::size_t` [here](https://github.com/ros2/rmw_zenoh/blob/bda2d7181d7e1ea2f7d143fabf7c4ffc6c670b20/rmw_zenoh_cpp/src/detail/liveliness_utils.cpp#L445) and send this size_t over any attachments as opposed to the 16B `Gid`.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
fuzzypixelz: Already there: https://github.com/ZettaScaleLabs/rmw_zenoh/blob/a372ede0834c69b0b649cf8f798b43a9f1c1b9c5/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp#L20.

rmw_zenoh_cpp/src/detail/liveliness_utils.hpp
fuzzypixelz: @Yadunund After taking a closer look at this, things are more complicated than they seem. FNV only  outperforms XXH3 for very small keys (see https://github.com/Cyan4973/xxHash/wiki/Performance-comparison#benchmarks-concentrating-on-small-data). So it's better to simply move to a 64-bit version of XXH3, as long as we agree that GID can be 8 bytes instead of 16. What do you think?

Yadunund:
@fuzzypixelz we started to discuss this PR in the context of #1028. Could you remind us why we marked this as draft again? Did it have to do with `XXH3` performance?

PR #435, opened 2025-01-24
Clear the error after rmw_serialized_message_resize()

clalancette:
It is actually an alias for rcutils_uint8_array_resize(), so the error message isn't useful.  Instead, reset the error and set a more useful one.


ahcorde (approved):

ahcorde:

Issue #439, opened 2025-01-30
tag 0.3.0 (current rolling binary) not working: Scouting delay elapsed before start conditions are met.


berndpfrommer:
The current apt package on Ubuntu 24.04 (noble) "does not work" (more below).
This is the package:
```
ros-rolling-rmw-zenoh-cpp/noble 0.3.0-1noble.20250102.230425 amd64
```

When I run the demo nodes (all on the same host), I get this warning:
```
ros2 run demo_nodes_cpp talker                                                                                                       
2025-01-30T12:57:34.365307Z  WARN ThreadId(06) zenoh::net::runtime::orchestrator: Scouting delay elapsed before start conditions are met.                               
[INFO] [1738241855.370982479] [talker]: Publishing: 'Hello World: 1'                                                                                                    
[INFO] [1738241856.370704297] [talker]: Publishing: 'Hello World: 2'                                                                   
```
And neither topic nor node list work (the listener also receives no messages):
```
ros2 topic list
2025-01-30T12:57:41.659755Z  WARN ThreadId(06) zenoh::net::runtime::orchestrator: Scouting delay elapsed before start conditions are met.
```

I was able to reproduce this problem by compiling the rmw_zenoh repo from source using tag 0.3.0.
I also noted that with tag 0.3.0 I no longer have to start an rmw_zenoh_hd daemon.

But if I upgrade to the latest commit e638f8c5ea79d614596ea4dfa89f07dc8065ad93 I have to start the ros2 daemon manually again - AND the "Scouting delay elapsed" message goes away, and everything works again.

What happened with 0.3.0? I don't understand why this was ever released and how you did not catch that in testing. Given the generally solid experience I had with rmw_zenoh, I still have nagging doubts that somehow my setup is very unusual or something is wrong on my side.

The other thing is: I don't like the ros2 daemon. It creates hidden state that I find confusing when switching between rmw versions (I went back and forth a couple times for testing). I'd much rather start the rmw_zenoh_hd manually, then I know exactly what process is doing the communication and I see its console log. Granted that's a matter of taste, but it may be worthwhile to poll people what they prefer.

clalancette:
> The other thing is: I don't like the ros2 daemon. It creates hidden state that I find confusing when switching between rmw versions (I went back and forth a couple times for testing).

See https://github.com/ros2/rmw_zenoh/issues/242

Yadunund:
Hi @berndpfrommer,

[...]

PR #440, opened 2025-01-30
Inform users that peers will not discover and communicate with one another until the router is started

Yadunund:
Address ambiguity from https://github.com/ros2/rmw_zenoh/issues/439

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
@@ -108,9 +108,17 @@ public:
           RMW_ZENOH_LOG_WARN_NAMED(
             "rmw_zenoh_cpp",
             "Unable to connect to a Zenoh router. "
-            "Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?");
+            "Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`? "
+            "You may also increase the number of attempts to check for a router "
+            "but setting the ZENOH_ROUTER_CHECK_ATTEMPTS environment variable.");
         }
         if (++connection_attempts >= configured_connection_attempts.value()) {
+          RMW_ZENOH_LOG_WARN_NAMED(
+            "rmw_zenoh_cpp",
+            "Unable to connect to a Zenoh router after %zu attempt(s). "
+            "Proceeding with initialization but other peers will not discover "
+            "or receive data from peers in this session until the router is started.",
+            configured_connection_attempts.value());
           break;
         }
         std::this_thread::sleep_for(sleep_time);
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -108,9 +108,17 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
           RMW_ZENOH_LOG_WARN_NAMED(
             "rmw_zenoh_cpp",
             "Unable to connect to a Zenoh router. "
-            "Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?");
+            "Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`? "
+            "You may also increase the number of attempts to check for a router "
+            "but setting the ZENOH_ROUTER_CHECK_ATTEMPTS environment variable.");
```
Timple: `but` -> `by`?

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
```diff
@@ -108,9 +108,17 @@ class rmw_context_impl_s::Data final : public std::enable_shared_from_this<Data>
           RMW_ZENOH_LOG_WARN_NAMED(
             "rmw_zenoh_cpp",
             "Unable to connect to a Zenoh router. "
-            "Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`?");
+            "Have you started a router with `ros2 run rmw_zenoh_cpp rmw_zenohd`? "
+            "You may also increase the number of attempts to check for a router "
```
Timple: I don't think this hint adds much. Playing with `ZENOH_ROUTER_CHECK_ATTEMPTS` is only useful if all checks are exhausted. And thus this hint would be better placed at the log below.

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Nice catch https://github.com/ros2/rmw_zenoh/pull/440/commits/dd555139cb4c0f2bf5044757cd3e7f6b0a2490fc

rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp
Yadunund: Makes sense https://github.com/ros2/rmw_zenoh/pull/440/commits/dd555139cb4c0f2bf5044757cd3e7f6b0a2490fc

Yadunund:

PR #454, opened 2025-02-07
Switch to debug log if topic_name not in topic_map

Yadunund:
For reasons described in the code comments.


ahcorde (approved):

Yadunund:
Might need to switch the other printouts to debug log as well (to avoid #455) so converting to debug for now.

Yadunund:
We discussed this issue today and the takeaway is that suppressing the logs for nested maps in GraphCache is still appropriate given the ordering behavior of liveliness tokens. The issue in #455 is likely because of some other underlying issue which is being masked by these printouts.

ahcorde (approved):

Yadunund:

PR #473, opened 2025-02-20
fix: use the default destructor that automatically drops the zenoh reply/query and hence sends the final signal

### LINKED ISSUE #53: Implement destructors for rmw_zenoh classes
Classes like `ZenohQuery` and `ZenohReply` should define the internal destructor to ensure the correct behaviour is triggered.

YuanYuYuan:
Closes https://github.com/ZettaScaleLabs/rmw_zenoh/issues/53.

The lack of destruction of `ZenohQuery` and `ZenohReply` leads to a memory leak. Furthermore, a drop of zenoh query/reply can trigger sending the final signal to the remote.

No regression was found on the usual test checks like system_tests, rcl, rclcpp, etc.


Yadunund (approved):

Yadunund:

PR #487, opened 2025-02-28
Use rustup dependency instead of cargo

Mallets:
Depends on https://github.com/eclipse-zenoh/zenoh/pull/1803 and https://github.com/ros/rosdistro/pull/44640.

Once they are merged, the [commit id](https://github.com/ros2/rmw_zenoh/blob/4fc36b75606911c2bbd2467667827f04edc7254f/zenoh_cpp_vendor/CMakeLists.txt#L29) of zenoh-c and zenoh-cpp should be updated.


Mallets:
I meant to open first the PR [here](https://github.com/ZettaScaleLabs/rmw_zenoh/pull/55).
Once the all the preparatory work is done, a PR will be opened upstream.

Issue #485, opened 2025-02-27
Align the default QoS with other RMWs


YuanYuYuan:
The following table summarizes the default QoS of the popular RMWs.

| **QoS Policy**                | **rmw_cyclonedds_cpp**            | **rmw_zenoh**             | **rmw_fastrtps_cpp**     |
| ---------------------------   | --------------------------------- | ------------------------- | ------------------------ |
| **History**                   | Keep Last (Depth = 1)             | Keep Last (Depth = 42)    | Keep Last (Depth = 1)    |
| **Reliability**               | Reliable                          | Reliable                  | Best Effort              |
| **Durability**                | Volatile                          | Transient Local           | Volatile                 |
| **Lifespan**                  | Infinite (if unspecified)         | Infinite                  | Infinite                 |
| **Deadline**                  | Infinite (if unspecified)         | Infinite                  | Infinite                 |
| **Liveliness**                | Automatic                         | Automatic                 | Automatic                |
| **Liveliness Lease Duration** | Infinite (if unspecified)         | Infinite                  | Infinite                 |

The question is why we use **Transient Local** by default. This different QoS setting causes the failure at [nav2_behavior_tree/test/plugins/condition/test_is_battery_low](https://github.com/ros-navigation/navigation2/blob/4694386872365851c840737515f89e3d409c1d9f/nav2_behavior_tree/test/plugins/condition/test_is_battery_low.cpp?plain=1#L111). Since navigation2 simply uses the default QoS in the test and expect the durability is volatile.

Mallets:
Default durability has been set to volatile in #521

PR #492, opened 2025-03-03
Make system default durability volatile

Yadunund:
This PR updates the system default durability to volatile from transient_local. This change is the better align with QoS settings in other RMWs https://github.com/ros2/rmw_zenoh/issues/485. While rmw_zenoh does not necessarily have to conform to other RMW settings but the durability, and particularly setting to transient_local by default, might lead to unintended behavior in tests and applications that users might have developed. So it makes sense to switch to volatile by default.

I've ran the usual CI tests and didn't see any new regressions. (rclpy/test_logging is flaky).

cc: @JEnoch @YuanYuYuan

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
@@ -16,6 +16,7 @@
 
 #include <fastcdr/FastBuffer.h>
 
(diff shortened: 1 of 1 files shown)


Yadunund:
@ahcorde could you run CI on all core repos to double check that there are no new regressions apart from those reported in #493

Issue #481, opened 2025-02-27
Tests failing in test_rmw_implementation


Yadunund:
Several tests in https://github.com/ros2/rmw_implementation/tree/rolling/test_rmw_implementation are failing.

- [x]  test_publisher__rmw_zenoh_cpp: 
- [x] test_subscription__rmw_zenoh_cpp
- [x] test_wait_set__rmw_zenoh_cpp
- [x] test_graph_api__rmw_zenoh_cpp
- [x] test_unique_identifiers__rmw_zenoh_cpp
- [x] test_service__rmw_zenoh_cpp
- [x] test_client__rmw_zenoh_cpp
- [x] test_event__rmw_zenoh_cpp

Details of one of many failures.

```
/home/yadunund/ros2_rolling_ci/src/ros2/rmw_implementation/test_rmw_implementation/test/test_subscription.cpp:144: Failure
53: Expected equality of these values:
53:   nullptr
53:     Which is: NULL
53:   sub
53:     Which is: 0x55622dc494a0
53: 
53: [  FAILED  ] TestSubscription.create_with_bad_arguments (4 ms)
```

Fixed by
- https://github.com/ros2/rmw_zenoh/pull/499
- https://github.com/ros2/rmw_zenoh/pull/497
- https://github.com/ros2/rmw_zenoh/pull/504
- https://github.com/ros2/rmw_zenoh/pull/508

Yadunund:
Some of these issues are related to the test expectations that expect creation of entities to fail if the qos_profile is rmw_qos_profile_unknown.

Some other test failures are related to expectations on matched count which requires similar fixes as https://github.com/ros2/rcl/pull/1164

Issue #493, opened 2025-03-03
Pass All Tests in ros2.repos


YuanYuYuan:
| name                    | list                                                               | reason                                                       |
| ----------------------- | ------------------------------------------------------------------ | ------------------------------------------------------------ |
| rosbag2_transport       | 1 - test_play__rmw_zenoh_cpp (Timeout)                             | Unknown Timeout                                              |
| rosbag2_transport       | 14 - test_burst__rmw_zenoh_cpp (Timeout)                           | Flaky                                                        |
| test_rmw_implementation | 4 - test_publisher__rmw_zenoh_cpp (Failed)                         | #418                                                         |
| test_rmw_implementation | 5 - test_subscription__rmw_zenoh_cpp (Failed)                      | #418                                                         |
| test_rmw_implementation | 9 - test_wait_set__rmw_zenoh_cpp (Failed)                          | #418                                                         |
| test_rmw_implementation | 10 - test_graph_api__rmw_zenoh_cpp (Failed)                        | #418                                                         |
| test_rmw_implementation | 11 - test_unique_identifiers__rmw_zenoh_cpp (Failed)               | #418                                                         |
| test_rmw_implementation | 12 - test_service__rmw_zenoh_cpp (Failed)                          | #418                                                         |
| test_rmw_implementation | 13 - test_client__rmw_zenoh_cpp (Failed)                           | #418                                                         |
| test_rmw_implementation | 16 - test_event__rmw_zenoh_cpp (Failed)                            | matched_status.current_count_change                          |
| image_transport         | 8 - image_transport-single_subscriber_publisher (Failed)           | https://github.com/ros-perception/image_common/pull/347                                    |
| message_filters         | 1 - message_filters-test_simple (Failed)                           | https://github.com/ros2/message_filters/pull/167                             |
| message_filters         | 2 - message_filters-msg_cache_unittest (Failed)                    | https://github.com/ros2/message_filters/pull/167                                   |
| message_filters         | 3 - message_filters-test_chain (Failed)                            |https://github.com/ros2/message_filters/pull/167                               |
| message_filters         | 4 - message_filters-test_time_sequencer (Failed)                   | https://github.com/ros2/message_filters/pull/167                                    |
| message_filters         | 5 - message_filters-test_subscriber (Failed)                       | https://github.com/ros2/message_filters/pull/167                                   |
[...]

Issue #391, opened 2025-01-03
Write pytest and ctest extensions that will help start the router in CI jobs


Yadunund:
The road to Tier-1 status includes passing all [nightly CI jobs](https://ci.ros2.org/view/nightly/) that run for all core ROS packages.

While we could setup the environment to run without the zenohd router and have nodes discover each other over multicast, it would mean we test with `rmw_zenoh` in a different configuration that the default behavior.

@cottsay recommended writing extensions for `pytest` and `ctest` that can start the zenohd router in a separate process. Also, multiple packages are tested in parallel on the same host in isolated environments and hence the extension would also need to generate router and sessions configs on the fly such that a new system port is assigned to the new router (and the session is notified) as having multiple zenohd routers start with the default config will lead to errors as they will try to bind to the same port.

Yadunund:
This is addressed by #583 and other changes to `ament_cmake_ros`.

PR #494, opened 2025-03-03
Fail creation of entities if qos contains unknown settings

Yadunund:
Partially addresses some of the failing tests in #481, Specifically those that expect creation of pub/sub/client/service to fail if the qos profile has unknown settings. The changes here make test_service__rmw_zenoh_cpp pass.

This works because QoS::get().best_available_qos() is invoked within all make() functions of these entities which will now return nullptr if qos_profile contains unknown settings.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/qos.cpp
+++ b/rmw_zenoh_cpp/src/detail/qos.cpp
@@ -74,7 +74,6 @@ bool QoS::is_supported(const rmw_qos_profile_t & qos_profile)
   return true;
 }
 
-
 ///=============================================================================
 rmw_ret_t QoS::best_available_qos(
   const rmw_node_t * node,
--- a/rmw_zenoh_cpp/src/detail/qos.cpp
+++ b/rmw_zenoh_cpp/src/detail/qos.cpp
@@ -61,6 +61,20 @@ const rmw_qos_profile_t & QoS::default_qos() const
 }
 
 ///=============================================================================
+bool QoS::is_supported(const rmw_qos_profile_t & qos_profile)
+{
+  if (qos_profile.history == RMW_QOS_POLICY_HISTORY_UNKNOWN ||
+    qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_UNKNOWN ||
+    qos_profile.durability == RMW_QOS_POLICY_DURABILITY_UNKNOWN ||
+    qos_profile.liveliness == RMW_QOS_POLICY_LIVELINESS_UNKNOWN) {
+      return false;
+    }
+
+  return true;
+}
+
+
+///=============================================================================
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/detail/qos.cpp
```diff
@@ -60,6 +60,21 @@ const rmw_qos_profile_t & QoS::default_qos() const
...
+    qos_profile.liveliness == RMW_QOS_POLICY_LIVELINESS_UNKNOWN)
+  {
+    return false;
+  }
+
+  return true;
+}
+
+
 ///=============================================================================
```
ahcorde:
```suggestion
}

///=============================================================================
```

ahcorde (changes requested):

rmw_zenoh_cpp/src/detail/qos.cpp
```diff
@@ -60,6 +60,21 @@ const rmw_qos_profile_t & QoS::default_qos() const
   return default_qos_;
 }
 
+///=============================================================================
+bool QoS::is_supported(const rmw_qos_profile_t & qos_profile)
+{
+  if (qos_profile.history == RMW_QOS_POLICY_HISTORY_UNKNOWN ||
```
Timple, line 66:
```c++
bool QoS::is_supported(const rmw_qos_profile_t & qos_profile)
{
  return (qos_profile.history != RMW_QOS_POLICY_HISTORY_UNKNOWN &&
    qos_profile.reliability != RMW_QOS_POLICY_RELIABILITY_UNKNOWN &&
    qos_profile.durability != RMW_QOS_POLICY_DURABILITY_UNKNOWN &&
    qos_profile.liveliness != RMW_QOS_POLICY_LIVELINESS_UNKNOWN)
}
```
This is a strange way of inverting a condition:
```c++
if (condition)
{
  return false;
}
return true;
```

Yadunund: I don't know if there is an objective "right way" here. I could argue that the current implementation is more readable or that I could add a debug log statement inside the braces 🤷🏼‍♂️ 
It's probably worth spending our efforts on outstanding items like #493, #391 to get `rmw_zenoh` ready for `kilted` instead :)

ahcorde:
I made this release to fix CI https://github.com/ros/rosdistro/pull/44667

ahcorde (approved):

ahcorde:
From source is compiling properly, merging

ahcorde:
https://github.com/Mergifyio backport jazzy

rmw_zenoh_cpp/src/detail/qos.cpp
Timple: I don't mind much though. I just triggered on the logic flip 😃

PR #498, opened 2025-03-05
Fix invalid argument

ahcorde:
Partially addresses some of the failing tests in https://github.com/ros2/rmw_zenoh/issues/481. Specifically those that expect creation of pub.


ahcorde:
already fixed here https://github.com/ros2/rmw_zenoh/pull/498

PR #500, opened 2025-03-05
Fixed windows warning

ahcorde:


rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -505,10 +505,10 @@ void SubscriptionData::add_new_message(
       msg->attachment.sequence_number() -
       last_known_pub_it->second);
     if (seq_increment > 1) {
-      const size_t num_msg_lost = seq_increment - 1;
+      const int64_t num_msg_lost = seq_increment - 1;
       events_mgr_->update_event_status(
         ZENOH_EVENT_MESSAGE_LOST,
-        num_msg_lost);
+        static_cast<int32_t>(num_msg_lost));
```
Yadunund: The truncation here from the static_cast can lead to UB if `num_msg_lost` is greater than `INT32_MAX`. It's unlikely that we'll ever have those many lost messages but it's probably better if we can avoid such patterns since we're fixing the implementation here.

How about

1. Switch to `int64_t` as the datatype for [current_count_change](https://github.com/ros2/rmw_zenoh/blob/bb2deedea45f45621ed2e6cc99874a8e011f4cdc/rmw_zenoh_cpp/src/detail/event.hpp#L120)
2. Pass `num_msg_lost` as is here

```suggestion
        num_msg_lost);
```

3. Update the implementation of [update_event_status](https://github.com/ros2/rmw_zenoh/blob/bb2deedea45f45621ed2e6cc99874a8e011f4cdc/rmw_zenoh_cpp/src/detail/event.cpp#L172) to clamp the value of `current_count_change` to [INT32_MIN, INT32_MAX]

Yadunund (changes requested):

rmw_zenoh_cpp/src/detail/event.hpp
```diff
@@ -60,7 +60,7 @@ struct rmw_zenoh_event_status_t
   size_t total_count;
   size_t total_count_change;
   size_t current_count;
-  int32_t current_count_change;
+  int64_t current_count_change;
```
Yadunund: We still want to keep this as `int32_t` and apply the clamps within update_event_status.

rmw_zenoh_cpp/src/rmw_event.cpp
```diff
@@ -243,7 +246,10 @@ rmw_take_event(
         ei->total_count = st.total_count;
         ei->total_count_change = st.total_count_change;
         ei->current_count = st.current_count;
-        ei->current_count_change = st.current_count_change;
+        ei->current_count_change = static_cast<int32_t>(std::clamp<int64_t>(
+          st.current_count_change,
+          std::numeric_limits<int32_t>::min(),
+          std::numeric_limits<int32_t>::max()));
```
Yadunund: Apply clamps in update_event_status

Yadunund (changes requested):

rmw_zenoh_cpp/src/rmw_event.cpp
ahcorde: clamping the value in the method call
https://github.com/ros2/rmw_zenoh/pull/500/commits/55e487ff78ae1a752677040eebac940ddafcb9d4

Yadunund (approved):
I pushed https://github.com/ros2/rmw_zenoh/pull/500/commits/4348b3baeb2900033a9a3f608a08ba3ae39108c9 to ensure we are explicitly cast the limits to `int64_t` which is the type defined for `seq_increment`.

Could you verify that the windows warning is gone before merging?

ahcorde:
> I pushed [4348b3b](https://github.com/ros2/rmw_zenoh/commit/4348b3baeb2900033a9a3f608a08ba3ae39108c9) to ensure we are explicitly cast the limits to `int64_t` which is the type defined for `seq_increment`.
> 
> Could you verify that the windows warning is gone before merging?

all good!

ahcorde:

PR #502, opened 2025-03-05
Added rmw_event_type_is_supported

ahcorde:
Related with this issue https://github.com/ros2/rmw/issues/394. Check if a event is supported by the `rmw`

### DIFF
--- a/rmw_zenoh_cpp/src/detail/event.hpp
+++ b/rmw_zenoh_cpp/src/detail/event.hpp
@@ -53,8 +53,6 @@ enum rmw_zenoh_event_type_t
 
 rmw_zenoh_event_type_t zenoh_event_from_rmw_event(rmw_event_type_t rmw_event_type);
 
-bool event_check_compatible(rmw_event_type_t rmw_event_type);
-
 ///=============================================================================
 /// A struct to store status changes which can be mapped to rmw event statuses.
 struct rmw_zenoh_event_status_t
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/src/rmw_event.cpp
```diff
@@ -175,6 +175,13 @@ rmw_event_set_callback(
   return RMW_RET_OK;
 }
 
+///=============================================================================
+bool
+rmw_event_type_is_supported(rmw_event_type_t rmw_event_type)
+{
+  return rmw_zenoh_cpp::event_check_compatible(rmw_event_type);
```
Yadunund: We don't need a new `event_check_compatible` method.
```suggestion
  return (rmw_zenoh_cpp::zenoh_event_from_rmw_event(rmw_event_type) != rmw_zenoh_cpp::ZENOH_EVENT_INVALID);
```

Yadunund (changes requested):

rmw_zenoh_cpp/src/detail/event.hpp
```diff
@@ -53,6 +53,8 @@ enum rmw_zenoh_event_type_t
 
 rmw_zenoh_event_type_t zenoh_event_from_rmw_event(rmw_event_type_t rmw_event_type);
 
+bool event_check_compatible(rmw_event_type_t rmw_event_type);
```
Yadunund: If we're keeping this function, we should update `rmw_event.cpp` to rely on it. Else we can remove this.

Yadunund (changes requested):

Yadunund (approved):

ahcorde:
https://github.com/Mergifyio backport jazzy

PR #505, opened 2025-03-06
Ignore local publications if specified

### LINKED ISSUE #496: Honor ignore_local_publications in subscription options
Add support to honor [ignore_local_publications](https://github.com/ros2/rmw/blob/95ac9726e5fea128d9e825775d3e5eccaf0a0713/rmw/include/rmw/types.h#L160-L168).

Required to get pass [test_rmw_implementation/test_subscripton](https://github.com/ros2/rmw_implementation/blob/b12c55d0b5e7d0cf1c51f41288e87b9dc1156539/test_rmw_implementation/test/test_subscription.cpp#L521) and address #481 

> Note: Neither rclcpp nor rclpy actually do anything with this option (even to manage intra-process comms). But since it's part of the tests, we should also support it.

Yadunund:
Fix #496 

This PR
- Updates `AttachmentData` to also store the zenoh session id (zid) of the entity creating the attachment.
- Updates publishers (and other entities) to append their `zid` to the attachment they fill in.
- Updates the `SubscriptionData` to accept `rmw_subscriptions_options_t`
-  Updates the sample cb in `SubscriptionData` to ignore any received samples with an attachment having zid matching the zid of the subscription if `ignore_local_publications` is set to `true`.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/attachment_helpers.cpp
+++ b/rmw_zenoh_cpp/src/detail/attachment_helpers.cpp
@@ -29,21 +29,27 @@
 namespace rmw_zenoh_cpp
 {
 
+///=============================================================================
 AttachmentData::AttachmentData(
   const int64_t sequence_number,
   const int64_t source_timestamp,
-  const std::array<uint8_t, RMW_GID_STORAGE_SIZE> source_gid)
+  const std::array<uint8_t, RMW_GID_STORAGE_SIZE> source_gid,
+  const std::string & zid)
 : sequence_number_(sequence_number),
   source_timestamp_(source_timestamp),
-  source_gid_(source_gid)
+  source_gid_(source_gid),
+  zid_(std::move(zid))
 {
+  // Do nothing.
 }
 
+///=============================================================================
 AttachmentData::AttachmentData(AttachmentData && data)
 {
   sequence_number_ = std::move(data.sequence_number_);
   source_timestamp_ = std::move(data.source_timestamp_);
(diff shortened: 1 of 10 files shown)


Yadunund:
@JEnoch I'm not sure if there is a way in Zenoh to automatically filter out publications from entities in the same zenoh session so I implemented this in the sample cb side.

YuanYuYuan:
Hi @Yadunund, in Zenoh, we can set "allowed_origin" to only receive messages from remote.

https://docs.rs/zenoh/latest/zenoh/pubsub/struct.SubscriberBuilder.html#method.allowed_origin

JEnoch:
@Yadunund Yes there is: in [`z_subscriber_options_t`](https://zenoh-c.readthedocs.io/en/latest/api.html#_CPPv422z_subscriber_options_t) you can set the `allowed_origin` field to [`ZC_LOCALITY_REMOTE `](https://zenoh-c.readthedocs.io/en/latest/api.html#_CPPv413zc_locality_t). This makes the Subscriber to receive only remote publications and to ignore publication from the same Session.

For `zenoh-cpp`: see [`SubscriberOptions.allowed_origin`](https://zenoh-cpp.readthedocs.io/en/latest/session.html#_CPPv4N5zenoh7Session17SubscriberOptionsE).

ahcorde:
I implemented here another option using zenoh-cpp API https://github.com/ros2/rmw_zenoh/pull/508

Yadunund:
Closing in favor of #508

PR #509, opened 2025-03-06
Config: tune some values for ROS 2 use cases

Mallets:
ROS 2 systems can experience a large number of nodes (e.g. >200) starting all together.
This may lead to having a large number of connections be initialised at the same time during the startup phase.

This PR updates some Zenoh default limits that are deemed too small for the ROS 2 cases for both the Router and the Nodes. The reason for each update is documented directly in the config files.

No behaviour change is expected to be introduced by this PR, just a smoother out-of-the-box experience for large use cases.


Yadunund (approved):
LGTM with usual CI passing.

Yadunund:

Issue #527, opened 2025-03-11
[Flaky] Exception in GraphListener thread: rcl_wait unexpectedly timed out


JEnoch:
The same error log has been reported by several users in different contexts, using latest release, or commit 

1. in [#455](https://github.com/ros2/rmw_zenoh/issues/455#issuecomment-2706129395) - using v0.2.2 (release for jazzy)
```log
2025-03-07T08:07:28.822902000Z [ERROR] [diagnostics_summarizer-29]: process has died [pid 662, exit code -6, cmd '/install/harvey/lib/harvey/diagnostics_summarizer --ros-args -r __node:=diagnostics_summarizer --params-file /tmp/launch_params_i1lgx4ix'].
2025-03-07T08:07:33.852035000Z [diagnostics_summarizer-29] [ERROR] [1741334848.118082488] [rclcpp]: caught std::exception exception in GraphListener thread: rcl_wait unexpectedly timed out
2025-03-07T08:07:33.882358000Z [diagnostics_summarizer-29] terminate called after throwing an instance of 'std::runtime_error'
2025-03-07T08:07:33.882532000Z [diagnostics_summarizer-29]   what():  rcl_wait unexpectedly timed out
2025-03-07T08:07:33.884842000Z [diagnostics_summarizer-29] Stack trace (most recent call last) in thread 1024:
2025-03-07T08:07:33.891791000Z [diagnostics_summarizer-29] #11   Object "", at 0xffffffffffffffff, in 
2025-03-07T08:07:33.891994000Z [diagnostics_summarizer-29] #10   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fc25b62bc3b, in 
2025-03-07T08:07:33.894063000Z [diagnostics_summarizer-29] #9    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7fc25b59eaa3, in 
2025-03-07T08:07:33.895319000Z [diagnostics_summarizer-29] #8    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7fc25b830db3, in 
2025-03-07T08:07:33.922034000Z [diagnostics_summarizer-29] #7    Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7fc25bb96ab7, in 
[...]

PR #528, opened 2025-03-11
Take wait_set_lock before condition_variable notification for subscriptions

Mallets:
In [rmw_wait](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2194) condition variables are used to wait for events on the provided `wait_set`. Unlike in [events](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/event.cpp#L253), [guard conditions](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/guard_condition.cpp#L40), [client data](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp#L235), and [service data](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/rmw_service_data.cpp#L264), no lock is taken before notifying for [subscription data](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp#L539).

This PR adds the lock for the *subscription data* case, which should fix the problem described in https://github.com/ros2/rmw_zenoh/issues/527.

@evshary could you check if `rcl_wait` errors disappear in `nav2_system_test` with this PR?

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
@@ -537,6 +537,7 @@ void SubscriptionData::add_new_message(
   // Since we added new data, trigger user callback and guard condition if they are available
   data_callback_mgr_.trigger_callback();
   if (wait_set_data_ != nullptr) {
+    std::lock_guard<std::mutex> wait_set_lock(wait_set_data_->condition_mutex);
     wait_set_data_->triggered = true;
     wait_set_data_->condition_variable.notify_one();
   }


evshary (commented):
The PR is verified with `nav2_system_test`
https://github.com/ros2/rmw_zenoh/issues/527#issuecomment-2714583097

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -537,6 +537,7 @@ void SubscriptionData::add_new_message(
   // Since we added new data, trigger user callback and guard condition if they are available
   data_callback_mgr_.trigger_callback();
   if (wait_set_data_ != nullptr) {
+    std::lock_guard<std::mutex> wait_set_lock(wait_set_data_->condition_mutex);
```
Yadunund, line 540: Should we lock above line 539 so that the check for `nullptr` is also protected by the lock?

Mallets: The lock resides inside the `wait_set_data` structure, so trying to take the lock before checking the validity of the pointer will very likely result in a `Segfault`.

Additionally the lock is taken in the same position as in [guard conditions](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/guard_condition.cpp#L40), [client data](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp#L235), and [service data](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/detail/rmw_service_data.cpp#L264).

ahcorde (approved):

ahcorde:

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
Yadunund: Oops I misread the code. Sorry for the noise!

Issue #523, opened 2025-03-10
[Test Failure] ros2 cli


ahcorde:
I noticed some failures in:

 - ros2node
```bash
======================================================================
FAIL: test_info_node[rmw_zenoh_cpp] (test_cli.TestROS2NodeCLI.test_info_node[rmw_zenoh_cpp])
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ahcorde/ros2_rolling/install/lib/python3.12/site-packages/launch_testing/markers.py", line 61, in _wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ahcorde/ros2_rolling/src/ros2/ros2cli/ros2node/test/test_cli.py", line 169, in test_info_node
    assert launch_testing.tools.expect_output(
AssertionError: Output does not match:
  /complex_node
    Subscribers:
      /strings: test_msgs/msg/Strings
    Publishers:
      /rosout: rcl_interfaces/msg/Log
      /parameter_events: rcl_interfaces/msg/ParameterEvent
      /arrays: test_msgs/msg/Arrays
    Service Servers:
      /complex_node/describe_parameters: rcl_interfaces/srv/DescribeParameters
      /complex_node/get_parameters: rcl_interfaces/srv/GetParameters
      /complex_node/get_parameter_types: rcl_interfaces/srv/GetParameterTypes
      /complex_node/list_parameters: rcl_interfaces/srv/ListParameters
      /complex_node/set_parameters: rcl_interfaces/srv/SetParameters
      /complex_node/set_parameters_atomically: rcl_interfaces/srv/SetParametersAtomically
      /complex_node/get_type_description: type_description_interfaces/srv/GetTypeDescription
      /basic: test_msgs/srv/BasicTypes
    Service Clients:
  
    Action Servers:
      /fibonacci: test_msgs/action/Fibonacci
    Action Clients:
  
  
assert False
 +  where False = <function expect_output at 0x7bed35d7ed40>(expected_lines=<itertools.chain object at 0x7bed30719600>, text='/complex_node\n  Subscribers:\n    /strings: test_msgs/msg/Strings\n  Publishers:\n    /rosout: rcl_interfaces/msg/Lo...BasicTypes\n  Service Clients:\n\n  Action Servers:\n    /fibonacci: test_msgs/action/Fibonacci\n  Action Clients:\n\n', strict=False)
[...]

PR #536, opened 2025-03-12
Sort alphabetically fill_names_and_types

ahcorde:
ros2node test are failing because the names are not sorted alphabetically, this should fix the issue

Related with this issue https://github.com/ros2/rmw_zenoh/issues/523

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -768,9 +768,17 @@ rmw_ret_t fill_names_and_types(
         RMW_SAFE_FWRITE_TO_STDERR("failed to cleanup names and types during error handling");
       }
     });
+
+  // sort entity_map alphabetically, entity_map is sorted by insertion
+  std::map<std::string, GraphNode::TopicTypeMap> ordered_entity_map;
+  for (const std::pair<std::string, GraphNode::TopicTypeMap> & item : entity_map) {
+    ordered_entity_map[item.first] = item.second;
+  }
+
   // Fill topic names and types.
   std::size_t index = 0;
-  for (const std::pair<std::string, GraphNode::TopicTypeMap> & item : entity_map) {
+  for (const std::pair<std::string, GraphNode::TopicTypeMap> item : ordered_entity_map) {
+
     names_and_types->names.data[index] = rcutils_strdup(item.first.c_str(), *allocator);
     if (!names_and_types->names.data[index]) {
       return RMW_RET_BAD_ALLOC;
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -778,7 +778,6 @@ rmw_ret_t fill_names_and_types(
   // Fill topic names and types.
   std::size_t index = 0;
   for (const std::pair<std::string, GraphNode::TopicTypeMap> item : ordered_entity_map) {
-
     names_and_types->names.data[index] = rcutils_strdup(item.first.c_str(), *allocator);
     if (!names_and_types->names.data[index]) {
       return RMW_RET_BAD_ALLOC;


Yadunund:
hmm we actually introduced `ordered_map` in https://github.com/ros2/rmw_zenoh/pull/250 to preserve the order in which entities were introduced into the system. That was required to fix some other tests. 
Could you double check that sorting alphabetically does not regress those other tests? Also it might be worthwhile to look into how the topics are filled in DDS-based middlewares which utilize [rmw_dds_common::GraphCache](https://github.com/ros2/rmw_dds_common/blob/abbcbb996532c4aa31048fc27122815d5d47547c/rmw_dds_common/src/graph_cache.cpp#L783).

ahcorde:
> hmm we actually introduced `ordered_map` in #250 to preserve the order in which entities were introduced into the system. That was required to fix some other tests. Could you double check that sorting alphabetically does not regress those other tests? Also it might be worthwhile to look into how the topics are filled in DDS-based middlewares which utilize [rmw_dds_common::GraphCache](https://github.com/ros2/rmw_dds_common/blob/abbcbb996532c4aa31048fc27122815d5d47547c/rmw_dds_common/src/graph_cache.cpp#L783).

@Yadunund According with the documentation in the code, we use a ordered map but remember by insertion https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L85-L88

I reruned all the stack I can't see any new error

Yadunund:
> > hmm we actually introduced `ordered_map` in #250 to preserve the order in which entities were introduced into the system. That was required to fix some other tests. Could you double check that sorting alphabetically does not regress those other tests? Also it might be worthwhile to look into how the topics are filled in DDS-based middlewares which utilize [rmw_dds_common::GraphCache](https://github.com/ros2/rmw_dds_common/blob/abbcbb996532c4aa31048fc27122815d5d47547c/rmw_dds_common/src/graph_cache.cpp#L783).
> 
> @Yadunund According with the documentation in the code, we use a ordered map but remember by insertion https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L85-L88
> 
> I reruned all the stack I can't see any new error

My point is that we explicitly switched to the custom ordered_map which preserves insertion order as some parts of the ROS 2 stack required this when calling get_names_and _types. 
So I'd like to be careful about this change and first ascertain what the original need was to switch to ordered_map.
If not required (unlikely) we might as well declare TopicMap as a map that stores keys alphabetically. Creating a temp map and inserting topics into that as you have it now adds unnecessary overhead at runtime. Hence, it would also be good to look at how the DDS graph cache handles both insertion ordering and alphabetical ordering requirements.

YuanYuYuan:
I browsed the other DDS-based implementations. They use the [`GraphCache::get_names_and_types`](https://github.com/ros2/rmw_dds_common/blob/abbcbb996532c4aa31048fc27122815d5d47547c/rmw_dds_common/src/graph_cache.cpp?plain=1#L800) to return the topics, which create a `NamesAndTypes topics` (basically a `std::map`). 

To align with them, we can preserve the `std::ordered_map` structure for some requirement by client libraries and create an additional sorted map for `rmw_get_topic_names_and_types` call. This is exactly what @ahcorde does in this PR.

YuanYuYuan:
A small suggestion: For better efficiency, we can create and sort only the keys. I also tried wrapping the `ordered_map` key with `std::reference_wrapper`, but it doesn't seem to be supported. So, we need to clone the `std::string` with minimal overhead.

```diff
diff --git a/rmw_zenoh_cpp/src/detail/graph_cache.cpp b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
index e24a53e..e8c1d8d 100644
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -770,15 +770,24 @@ rmw_ret_t fill_names_and_types(
     });
   // Fill topic names and types.
   std::size_t index = 0;
-  for (const std::pair<std::string, GraphNode::TopicTypeMap> & item : entity_map) {
-    names_and_types->names.data[index] = rcutils_strdup(item.first.c_str(), *allocator);
+
+  // Sort the topics of entity_map alphabetically
+  std::vector<std::string> sorted_topics;
+  for (const auto & item : entity_map) {
+    sorted_topics.push_back(item.first);
+  }
+  std::sort(sorted_topics.begin(), sorted_topics.end());
+
+  for (const auto & topic : sorted_topics) {
+    names_and_types->names.data[index] = rcutils_strdup(topic.c_str(), *allocator);
     if (!names_and_types->names.data[index]) {
       return RMW_RET_BAD_ALLOC;
     }
 
+    auto topic_map = entity_map.at(topic);
     rcutils_ret_t rcutils_ret = rcutils_string_array_init(
       &names_and_types->types[index],
-      item.second.size(),
+      topic_map.size(),
       allocator);
     if (RCUTILS_RET_OK != rcutils_ret) {
       RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
@@ -786,7 +795,7 @@ rmw_ret_t fill_names_and_types(
     }
 
     size_t type_index = 0;
-    for (const std::pair<const std::string, GraphNode::TopicQoSMap> & type : item.second) {
+    for (const std::pair<const std::string, GraphNode::TopicQoSMap> & type : topic_map) {
       char * type_name = rcutils_strdup(_demangle_if_ros_type(type.first).c_str(), *allocator);
       if (!type_name) {
         RMW_SET_ERROR_MSG("failed to allocate memory for type name");
```

ahcorde:
If I use a `std::map` o `std::unordered_map` instead of the `tsl::ordered_map`. This was the PR that this was introduced https://github.com/ros2/rmw_zenoh/commit/c12ff3ed2b73169ff0bd7a2a341f8911082fc1e9

I can see some error in rcl:

```bash
	 97 - test_get_node_names__rmw_zenoh_cpp (Failed)
```

or rclcpp

```bash
	 44 - test_node_interfaces__node_graph (Failed)
```

YuanYuYuan:
> If I use a `std::map` o `std::unordered_map` instead of the `tsl::ordered_map`. This was the PR that this was introduced [c12ff3e](https://github.com/ros2/rmw_zenoh/commit/c12ff3ed2b73169ff0bd7a2a341f8911082fc1e9)
> 
> I can see some error in rcl:
> 
> ```shell
> 	 97 - test_get_node_names__rmw_zenoh_cpp (Failed)
> ```
> 
> or rclcpp
> 
> ```shell
> 	 44 - test_node_interfaces__node_graph (Failed)
> ```

Hi @ahcorde, I tested this [branch](https://github.com/ZettaScaleLabs/rmw_zenoh/commits/test/ordered_map/) against rcl, rclcpp, and system_tests. I didn't see any failures.
[refactor-ordered-map-results.log](https://github.com/user-attachments/files/19249138/refactor-ordered-map-results.log)

ahcorde:
@YuanYuYuan That's true, I had a hanging process which made these tests to fail

Do you min to try this one https://github.com/ros2/rmw_zenoh/pull/546?

YuanYuYuan:
> @YuanYuYuan That's true, I had a hanging process which made these tests to fail
> 
> Do you min to try this one #546?

Sure. Will do.

YuanYuYuan:
1. #546 works fine. I believe we can use a `std::map` now.
2. I can reproduce the failure on rcl/test_get_node_names__rmw_zenoh_cpp and rclcpp/test_node_interfaces__node_graph if I have some hang processes caused by the failing ros2topic.

Yadunund:
> If I use a `std::map` o `std::unordered_map` instead of the `tsl::ordered_map`. This was the PR that this was introduced [c12ff3e](https://github.com/ros2/rmw_zenoh/commit/c12ff3ed2b73169ff0bd7a2a341f8911082fc1e9)
> 
> I can see some error in rcl:
> 
> ```shell
> 	 97 - test_get_node_names__rmw_zenoh_cpp (Failed)
> ```
> 
> or rclcpp
> 
> ```shell
> 	 44 - test_node_interfaces__node_graph (Failed)
> ```

So here's where i'm confused.

`rcl/test_get_node_names` calls `rcl_get_node_names_with_enclaves` which in turn calls `rmw_get_node_names_with_enclaves`. The implementation of `rmw_get_node_names_with_enclaves` in rmw_zennoh calls [GraphCache::get_node_names](https://github.com/ros2/rmw_zenoh/blob/63f0623d1e538b0b5a0b5cbb510ade31cdb45190/rmw_zenoh_cpp/src/detail/graph_cache.cpp#L618) which simply iterates over an [std::unordered_map](https://github.com/ros2/rmw_zenoh/blob/63f0623d1e538b0b5a0b5cbb510ade31cdb45190/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L261) to get the list of node names. So it should not require any ordering. 

My original question still stands, which is the test that actually requires topic names in a node to be ordered? Because the ordered_map is only used [here](https://github.com/ros2/rmw_zenoh/blob/63f0623d1e538b0b5a0b5cbb510ade31cdb45190/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L88). I'm not comfortable with switching to std::map without ascertaining this.

ahcorde:
> > If I use a `std::map` o `std::unordered_map` instead of the `tsl::ordered_map`. This was the PR that this was introduced [c12ff3e](https://github.com/ros2/rmw_zenoh/commit/c12ff3ed2b73169ff0bd7a2a341f8911082fc1e9)
> > I can see some error in rcl:
> > ```shell
> > 	 97 - test_get_node_names__rmw_zenoh_cpp (Failed)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or rclcpp
> > ```shell
> > 	 44 - test_node_interfaces__node_graph (Failed)
> > ```
> 
> So here's where i'm confused.
> 
> `rcl/test_get_node_names` calls `rcl_get_node_names_with_enclaves` which in turn calls `rmw_get_node_names_with_enclaves`. The implementation of `rmw_get_node_names_with_enclaves` in rmw_zennoh calls [GraphCache::get_node_names](https://github.com/ros2/rmw_zenoh/blob/63f0623d1e538b0b5a0b5cbb510ade31cdb45190/rmw_zenoh_cpp/src/detail/graph_cache.cpp#L618) which simply iterates over an [std::unordered_map](https://github.com/ros2/rmw_zenoh/blob/63f0623d1e538b0b5a0b5cbb510ade31cdb45190/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L261) to get the list of node names. So it should not require any ordering.
> 
> My original question still stands, which is the test that actually requires topic names in a node to be ordered? Because the ordered_map is only used [here](https://github.com/ros2/rmw_zenoh/blob/63f0623d1e538b0b5a0b5cbb510ade31cdb45190/rmw_zenoh_cpp/src/detail/graph_cache.hpp#L88). I'm not comfortable with switching to std::map without ascertaining this.

@Yadunund forget about this failures as I mentioned I had a hanging process that was making my local tests to fail.

I made this other PR https://github.com/ros2/rmw_zenoh/pull/546 and CI is green @YuanYuYuan confirmed that it's working

ahcorde:
closing this one in favour of this other one https://github.com/ros2/rmw_zenoh/pull/546

PR #546, opened 2025-03-14
Switch to std::map for TopicTypeMap

ahcorde:
I open this PR just for visibility, it's related with this PR https://github.com/ros2/rmw_zenoh/pull/536 and in particular with this comment https://github.com/ros2/rmw_zenoh/pull/536#issuecomment-2719897617


Yadunund (commented):
@clalancette I'm sure had a good reason to introduce `tsl::ordered_map` in https://github.com/ros2/rmw_zenoh/pull/536. Do you recall which tests were failing without it? Could it have been that the requirements were to return topic names in alphabetical order and not insertion order and that the failing test actually added topics in alphabetical order?

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -85,7 +85,7 @@ struct GraphNode
   // Map topic name to TopicTypeMap
   // This uses a map that remembers insertion order because some parts of the client libraries
   // expect that these are returned in the order that they were created.
-  using TopicMap = tsl::ordered_map<std::string, TopicTypeMap>;
+  using TopicMap = std::map<std::string, TopicTypeMap>;
```
Yadunund, line 87: Update documentation above.

Yadunund (changes requested):
We should also delete

https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/ordered_map.hpp and https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/ordered_hash.hpp

And update CI to not reference those file https://github.com/ros2/rmw_zenoh/blob/caaeddef64707682b205d20d75c8454d0ccd232c/.github/workflows/style.yaml#L22-L24

clalancette:
> @clalancette I'm sure had a good reason to introduce `tsl::ordered_map` in #536. Do you recall which tests were failing without it? Could it have been that the requirements were to return topic names in alphabetical order and not insertion order and that the failing test actually added topics in alphabetical order?

Oof, this is a tough one.  I made a big mistake in PR #250 in that I did not specify which tests were failing.  And there definitely were failing tests at that point that adding the ordered map fixed.

So I went ahead and rebuilt a workspace with this PR in it, and ran all tests.  With this in place, every test passed.  So whatever it was that was failing before, was fixed in a different way.  Thus, I think we can go forward with this change (with one change, which I'll put in a review).

clalancette (changes requested):
We should also remove the MIT license from the package.xml, as it is no longer needed.

rmw_zenoh_cpp/CMakeLists.txt
```diff
@@ -94,8 +94,6 @@ if(BUILD_TESTING)
   find_package(ament_cmake_uncrustify REQUIRED)
   find_package(ament_cmake_xmllint REQUIRED)
 
-  set(_linter_excludes src/detail/ordered_hash.hpp src/detail/ordered_map.hpp)
-
   ament_copyright(EXCLUDE src/detail/simplified_xxhash3.cpp src/detail/simplified_xxhash3.hpp)
   ament_cpplint(EXCLUDE ${_linter_excludes})
```
clalancette: Sorry, I just noticed this one.  Since we are no longer setting `_linter_excludes`, this should be:

```suggestion
  ament_cpplint()
```

(same for `ament_uncrustify` down below, which I can't put a comment in due to GitHub limitations)

clalancette (changes requested):

clalancette (approved):
This looks good to me, thanks @ahcorde

rmw_zenoh_cpp/src/detail/graph_cache.hpp
```diff
@@ -83,9 +82,9 @@ struct GraphNode
   // Map topic type to QoSMap
   using TopicTypeMap = std::unordered_map<std::string, TopicQoSMap>;
   // Map topic name to TopicTypeMap
-  // This uses a map that remembers insertion order because some parts of the client libraries
-  // expect that these are returned in the order that they were created.
-  using TopicMap = tsl::ordered_map<std::string, TopicTypeMap>;
+  // This uses a map that sort element by name because some parts of the client libraries
```
Yadunund, line 85: Nit
```suggestion
  // This uses a map that sorts keys alphabetically because some parts of the client libraries
```

ahcorde:
Thank you for the review @clalancette :+1:

ahcorde:

PR #547, opened 2025-03-14
Make reliable topics use blocking congestion control

### LINKED ISSUE #543: Unable to send 10 MB messages through the network
While running ping-pong test across a network, a large-size message can't be received.
If sending with a 1MB message, data might be missed. With a 10 MB message, we can never receive that.

We test with the package: https://github.com/ZettaScaleLabs/ros2-simple-performance in the topology

```
     host 1                   host 2
[ping ----- zenohd] ----- [zenohd ----- pong]
```

* Host 1: `ros2 run simple_performance ping --ros-args -p warmup:=1.0 -p size:=10000000 -p samples:=10 -p rate:=1`
* Host 2: `ros2 run simple_performance pong`

We can't receive data on pong size, but it works when size is smaller.

Here are some investigations on the pong side:

* Analyzing with gdb, the callback inside `SubscriptionData` isn't triggered.
* Enabling with `RUST_LOG=z=trace`, Zenoh didn't show the message payload.
* We indeed see the packets in the Wireshark.
* Using the same configuration as rmw_zenoh, Zenoh examples `z_ping` and `z_pong` can work.

The issue is originally reported [here](https://github.com/ZettaScaleLabs/rmw_zenoh/issues/38#issuecomment-2698119036)

Hugal31:
Make all reliable topics use the BLOCK congestion control. Rational: drop would lose messages, which is not what reliable mean.

Closes #543

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
@@ -120,14 +120,12 @@ std::shared_ptr<PublisherData> PublisherData::make(
   zenoh::KeyExpr pub_ke(entity->topic_info()->topic_keyexpr_);
   // Set congestion_control to BLOCK if appropriate.
   auto pub_opts = zenoh::Session::PublisherOptions::create_default();
-  pub_opts.congestion_control = Z_CONGESTION_CONTROL_DROP;
   if (adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
     pub_opts.reliability = Z_RELIABILITY_RELIABLE;
-    if (adapted_qos_profile.history == RMW_QOS_POLICY_HISTORY_KEEP_ALL) {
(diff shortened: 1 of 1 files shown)


Yadunund:
I don't think we want to do this for reasons described by @JEnoch https://github.com/ros2/rmw_zenoh/issues/543#issuecomment-2724842457

fujitatomoya:
@Yadunund @Hugal31 are you guys going to close this?

Hugal31:
Well I still think that's it is an imperfect solution, but it is less surprising than the current behavior. I'd like to continue the conversation in #543 and find a proper solution though.

Yadunund:
@Hugal31 with https://github.com/ros2/rmw_zenoh/pull/591, Zettascale improved e2e reliability with HeartbeatSporadic while avoiding issues with congestion. Please try out the latest rolling and report your findings. Happy to continue discussion in a new ticket if needed. Closing out this PR for now.

PR #551, opened 2025-03-17
feat: support zenoh config override

YuanYuYuan:
This PR allows users to override the zenoh config via an environmental variable. 

For instance, we can easily change the port of ros2 nodes and the zenohd as below.

```bash
export ZENOH_CONFIG_OVERRIDE='connect/endpoints=["tcp/127.0.0.1:7448"]'
ros2 run demo_nodes_cpp listener

export ZENOH_CONFIG_OVERRIDE='listen/endpoints=["tcp/127.0.0.1:7448"]'
ros2 run rmw_zenoh_cpp rmw_zenohd
``` 

Enable multicast scouting without a zenohd router temporarily.
```bash
export ZENOH_CONFIG_OVERRIDE='scouting/multicast/enabled=true'
ros2 run demo_nodes_cpp listener

export ZENOH_CONFIG_OVERRIDE='scouting/multicast/enabled=true'
ros2 run demo_nodes_cpp talker
```

Multiple key-value pairs are also supported through the syntax `ZENOH_CONFIG_OVERRIDE='k1=v1;k2=v2;...'`.
The config generation order is `ZENOH_CONFIG_URI` and then apply `ZENOH_CONFIG_OVERRIDE`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,28 @@ For example, to set the path to a custom `Zenoh router` configuration file,
 export ZENOH_ROUTER_CONFIG_URI=$HOME/MY_ZENOH_ROUTER_CONFIG.json5
 ```
 
+`rmw_zenoh` allows you to override configuration fields using the `ZENOH_CONFIG_OVERRIDE` environment variable .
+These overrides apply to ROS2 nodes and the Zenoh router **after** the `ZENOH_SESSION_CONFIG_URI` or `ZENOH_ROUTER_CONFIG_URI` (if specified)
+has been processed.
+
+You can specify multiple key-value pairs using the following syntax:
+```bash
+export ZENOH_CONFIG_OVERRIDE="key/path/to/field1=value1;key/path/to/field2=value2"
+```
+
+#### Examples
+
+- Specify custom endpoints for listening and connecting:
+```bash
+export ZENOH_CONFIG_OVERRIDE='listen/endpoints=["tcp/127.0.0.1:7448"];connect/endpoints=["tcp/192.168.0.1:7449", "tcp/192.168.0.2:7449"]'
+```
+This configuration sets the node to listen on
+
+- Enable multicast scouting (disabled by default) that allows ROS2 nodes to discover each other without requiring a Zenoh router:
+```bash
+export ZENOH_CONFIG_OVERRIDE='scouting/multicast/enabled=true'
+```
+
 ### Connecting multiple hosts
 By default, all discovery & communication is restricted within a host, where a host is a machine running a `Zenoh router` along with various ROS 2 nodes with their default [configurations](rmw_zenoh_cpp/config/).
 To bridge communications across two or more hosts, the `Zenoh router` configuration for one of the hosts must be updated to connect to the other host's `Zenoh router` at startup.
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ export ZENOH_ROUTER_CONFIG_URI=$HOME/MY_ZENOH_ROUTER_CONFIG.json5
 ```
 
 `rmw_zenoh` allows you to override configuration fields using the `ZENOH_CONFIG_OVERRIDE` environment variable .
-These overrides apply to ROS2 nodes and the Zenoh router **after** the `ZENOH_SESSION_CONFIG_URI` or `ZENOH_ROUTER_CONFIG_URI` (if specified)
+These overrides apply to ROS 2 nodes and the Zenoh router **after** the `ZENOH_SESSION_CONFIG_URI` or `ZENOH_ROUTER_CONFIG_URI` (if specified)
 has been processed.
 
 You can specify multiple key-value pairs using the following syntax:
--- a/rmw_zenoh_cpp/src/detail/zenoh_config.cpp
+++ b/rmw_zenoh_cpp/src/detail/zenoh_config.cpp
@@ -17,6 +17,7 @@
 #include <rcutils/env.h>
 
 #include <limits>
+#include <sstream>
 #include <string>
 
 #include <zenoh.hxx>
--- a/rmw_zenoh_cpp/src/detail/zenoh_config.cpp
+++ b/rmw_zenoh_cpp/src/detail/zenoh_config.cpp
@@ -44,8 +44,8 @@ static const std::unordered_map<ConfigurableEntity,
 
 static const char * router_check_attempts_envar = "ZENOH_ROUTER_CHECK_ATTEMPTS";
 /// Allow users to override the configuration using key-value pairs.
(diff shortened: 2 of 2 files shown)


YuanYuYuan:
Open question: I chose the word "override" rather than "overwrite" since it's not a simple json5 field replacement. The configuration data has been constructed (by default or by a given file path) and then modified by the environment variable. But I'm open to any suggestion :smiley: .

rmw_zenoh_cpp/src/detail/zenoh_config.cpp
```diff
@@ -69,6 +73,33 @@ std::optional<zenoh::Config> _get_z_config(
...
     return std::nullopt;
   }
+
+  // Override the zenoh config by key/value pairs.
+  const char * key_val_pairs_str;
+  if (NULL != rcutils_get_env(zenoh_config_override, &key_val_pairs_str)) {
+    // NULL is returned if everything is ok.
+    RMW_ZENOH_LOG_ERROR_NAMED(
+      "rmw_zenoh_cpp", "Envar %s cannot be read.", zenoh_config_override);
+    return std::nullopt;
```
Yadunund:
```suggestion
      "rmw_zenoh_cpp", "Envar %s cannot be read. Ignoring override...", zenoh_config_override);
    return config;
```

rmw_zenoh_cpp/src/detail/zenoh_config.cpp
```diff
@@ -69,6 +73,33 @@ std::optional<zenoh::Config> _get_z_config(
...
+    while (std::getline(ss, pair, ';')) {
+      std::stringstream pair_stream(pair);
+      std::string key, val;
+      if (std::getline(pair_stream, key, '=') && std::getline(pair_stream, val)) {
+        config.insert_json5(key, val, &result);
+        if (result != Z_OK) {
+          RMW_ZENOH_LOG_ERROR_NAMED(
+            "rmw_zenoh_cpp",
+            "Invalid configuration key/value pair: (%s, %s)", key.c_str(), val.c_str());
+          return std::nullopt;
```
Yadunund:
```suggestion
            "Invalid configuration key/value pair: (%s, %s). Ignoring override...", key.c_str(), val.c_str());
          continue;
```

Yadunund (changes requested):
This is a nice addition! Left some feedback to better handle bad envars. Specifically, proceeding with using the config file as is if the `ZENOH_CONFIG_OVERRIDE` envar cannot be read and also skipping (key, value) pairs that are not valid.

Else running 
```
ZENOH_CONFIG_OVERRIDE='foo=bar' ros2 run demo_nodes_cpp talker
```

will throw a runtime error 

```bash
2025-03-18T00:34:47.977107Z ERROR ThreadId(01) zenohc::config: Failed to insert value 'bar' for key 'foo' into config: Json5Err(Message { msg: " --> 1:1\n  |\n1 | bar\n  | ^---\n  |\n  = expected array, boolean, null, number, object, or string", location: Some(Location { line: 1, column: 1 }) }) at /home/yadunund/.cargo/git/checkouts/zenoh-cc237f2570fab813/b661454/zenoh/src/api/config.rs:67.
[ERROR] [1742258087.977152367] [rmw_zenoh_cpp]: Invalid configuration key/value pair: (foo, bar)
terminate called after throwing an instance of 'rclcpp::exceptions::RCLBadAlloc'
  what():  std::bad_alloc
[ros2run]: Aborted
```

but it will be better if we proceed with initialization while skipping these overrides instead.

rmw_zenoh_cpp/src/detail/zenoh_config.cpp
YuanYuYuan: Addressed in [26e0c8b](https://github.com/ros2/rmw_zenoh/pull/551/commits/26e0c8b5e84cd482951b4d33dc3190d211182925).

rmw_zenoh_cpp/src/detail/zenoh_config.cpp
YuanYuYuan: Addressed in [26e0c8b](https://github.com/ros2/rmw_zenoh/pull/551/commits/26e0c8b5e84cd482951b4d33dc3190d211182925).

YuanYuYuan:
Thanks @Yadunund. I've implemented your suggestion with a slight modification to display a warning if an invalid key-value pair is detected. This is the same policy we used in [the upstream zenohd](https://github.com/eclipse-zenoh/zenoh/blob/27bb8df3c21ae359cb286dee2caacc14ac0f2c14/zenohd/src/main.rs?plain=1#L249).

README.md
```diff
@@ -121,6 +121,28 @@ For example, to set the path to a custom `Zenoh router` configuration file,
 export ZENOH_ROUTER_CONFIG_URI=$HOME/MY_ZENOH_ROUTER_CONFIG.json5
 ```
 
+`rmw_zenoh` allows you to override configuration fields using the `ZENOH_CONFIG_OVERRIDE` environment variable .
+These overrides apply to ROS2 nodes and the Zenoh router **after** the `ZENOH_SESSION_CONFIG_URI` or `ZENOH_ROUTER_CONFIG_URI` (if specified)
```
ahcorde:
```suggestion
These overrides apply to ROS 2 nodes and the Zenoh router **after** the `ZENOH_SESSION_CONFIG_URI` or `ZENOH_ROUTER_CONFIG_URI` (if specified)
```

rmw_zenoh_cpp/src/detail/zenoh_config.cpp
```diff
@@ -69,6 +73,32 @@ std::optional<zenoh::Config> _get_z_config(
...
+    // NULL is returned if everything is ok.
+    RMW_ZENOH_LOG_ERROR_NAMED(
+    "rmw_zenoh_cpp", "Envar %s cannot be read. Ignoring override...", zenoh_config_override);
+    return config;
+  }
+  if (key_val_pairs_str[0] != '\0') {
+    std::stringstream ss(key_val_pairs_str);
+    std::string pair;
+    while (std::getline(ss, pair, ';')) {
+      std::stringstream pair_stream(pair);
```
ahcorde, line 90: include `<sstream>`

ahcorde (changes requested):

rmw_zenoh_cpp/src/detail/zenoh_config.cpp
Yadunund: https://github.com/ros2/rmw_zenoh/pull/551/commits/435dac41d88534ebf29aee181e27eed5f4e2e509

Yadunund (approved):

Yadunund:

PR #556, opened 2025-03-18
Align the config with the latest Zenoh

evshary:
Align with the latest Zenoh configuration to make future comparisons easier.
We don't change any value of the config in the PR, so the behavior shouldn't be changed.


ahcorde (approved):

ahcorde:

PR #575, opened 2025-03-26
Set wait_set->triggered flag to false

Mallets:
This PR addresses https://github.com/ros2/rmw_zenoh/issues/527#issuecomment-2744152823.

By taking the lock before the `check_and_attach_condition` and resetting the `triggered` flag, a race condition is solved when the set was updated (or cleared) but the triggered status was not reset.

The error
```
[rclcpp]: caught std::exception exception in GraphListener thread: rcl_wait unexpectedly timed out
```
seems now to have disappeared from the [flaky test](https://github.com/ros2/rmw_zenoh/issues/527).

However, a more realistic test on a real robot would be highly appreciated.

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -2122,6 +2122,8 @@ check_and_attach_condition(
       }
     }
   }
+  // No guard conditions are available. Set the triggered flag of the wait_set to false.
+  wait_set_data->triggered = false;
 
   if (events) {
     for (size_t i = 0; i < events->event_count; ++i) {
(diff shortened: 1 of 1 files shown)


rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2122,6 +2122,8 @@ check_and_attach_condition(
       }
     }
   }
+  // No guard conditions are available. Set the triggered flag of the wait_set to false.
+  wait_set_data->triggered = false;
```
JEnoch: The `wait_set_data->triggered` flag is not just about guard conditions.
I think it shall be set to `false` only when all the conditions are unset. i.e. at the very end of this function when returning `false`.

Mallets: Fixed

YuanYuYuan:
@evshary can you test if the deadlock in Autoware could be fixed by this?

evshary:
> @evshary can you test if the deadlock in Autoware could be fixed by this?

No, it doesn't fix the issue. ~~However, I believe the issue in Autoware is also related to rmw_wait.~~
The [issue](https://github.com/evshary/autoware_rmw_zenoh/issues/6) in Autoware needs more analysis. I don't think it should block the PR.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2228,33 +2231,35 @@ rmw_wait(
   // signals to the upper layers that it isn't ready.  If something is ready, then we leave it as
   // a valid pointer.
 
-  bool skip_wait = check_and_attach_condition(
-    subscriptions, guard_conditions, services, clients, events, wait_set_data);
-  if (!skip_wait) {
+  {
```
Yadunund, line 2236: Could we add some documentation here on why we now take the lock before calling `check_and_attach_condition`?

rmw_zenoh_cpp/src/rmw_zenoh.cpp
```diff
@@ -2184,6 +2184,9 @@ check_and_attach_condition(
     }
   }
 
+  // No conditions are available. Set the triggered flag of the wait_set to false.
```
Yadunund, line 2187: Should we append a note here that we've already locked `wait_set_data->condition_mutex` prior to calling `check_and_attach_condition`?

Yadunund (commented):
The fix makes sense to me. Left some minor comments to consider. Will run full CI with the changes here before merging.

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Mallets: Done

rmw_zenoh_cpp/src/rmw_zenoh.cpp
Mallets: Done

Yadunund (approved):
Thanks for addressing the feedback!

Yadunund:
CI:

Gist: https://gist.githubusercontent.com/Yadunund/c4074b12ea5bdedcddae7b60f60c34bc/raw/19a1f54f01b6e523ea1f3070cd516a93c314293d/ros2_ci.repos

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22945/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17166/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2431/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23820/)

Yadunund:

PR #592, opened 2025-04-04
Update policy_parser to use nlohmann_json.hpp

mjcarroll:
Since we have the dependency now via #583, this is a potential improvement to the current string concatenation.

Not blocking, just a suggestion.


zenoh_security_configuration_tools/package.xml
```diff
@@ -7,6 +7,9 @@
   <maintainer email="alejandro@openrobotics.org">Alejandro Hernanadez</maintainer>
   <license>Apache License 2.0</license>
 
+
```
Yadunund, line 10:
```suggestion
```

Yadunund (approved):

Yadunund (approved):

Issue #623, opened 2025-04-27
Setting DiffServe value of select topics


gstangel:
### Description

For use cases involving command and control ROS2 streams, setting the Differentiated Services field of a IP frame is a useful feature.  For example, in drone swarms that may be connected via a smart meshed WiFi Radio, the differentiated service field can be used in times of contention to ensure that the radio selectively prioritizes command and control streams.

### Motivation

ROS2 has no inherent ability to set a 'hard' network level priority on data.  The different QoS Settings are applied at the software layer.  This doesn't necessarily make sense with a TCP Stream, but for UDP data it can be helpful

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-news-for-the-week-of-may-12th-2025/43803/1

JEnoch:
Zenoh now supports DiffServ via a new `dscp` (differentiated services code point) configuration per-endpoint.
See https://github.com/eclipse-zenoh/zenoh/pull/1937 which resolves https://github.com/eclipse-zenoh/zenoh/issues/1924.

It will be available in `rmw_zenoh` with the next bump of `zenoh-c` commit id.

JEnoch:
@gstangel `zenoh-c` and `zenoh-cpp` are now synced with `zenoh` wrt. `dscp` feature.
Waiting for the bump in `rmw_zenoh`, could you test this feature in advance ?

You need to:
1. build from sources, changing the commit ids in zenoh_cpp_vendor/CMakeLists.txt with:
   - `zenoh-c`: d00d5da2226b084d83cedf0545d3976d5a630c7c
   - `zenoh-cpp`: c96600c9a91ea5a63046c3830e5e442fd255f3a1
  2. configure 2 routers across you meshed WiFi to establish 2 TCP connections with different `priority` and `dscp`, and to configure priority overwrite when routing over the WiFi interface

Here is an example of the changes to apply in the routers config (assuming one has an interface "en0" over WiFi with IP 192.168.0.2).
Router 1 (listening on 192.168.0.2 over WiFi):
```json5
{
    listen: {
        endpoints: [
            "tcp/127.0.0.1:7447",  // For the Nodes over loopback
            "tcp/192.168.0.2:8888?prio=4-7",   // Low priority over WiFi
            "tcp/192.168.0.2:9999?prio=0-3#dscp=0x48"]  // High priority over WiFi
    },
[...]

Issue #447, opened 2025-02-04
Excessive disk usage in build directory - 4.4 GB


PaulBouchier:
When building using the command line specified in the README.md, 
`colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release`
the build directory grows to 4.4 GB. This seems to be due to stuff in a couple of directories called debug way down in the build tree. 

This problem causes issues on small platforms like Raspberry Pi (I ran out of space), and the requirement to give --cmake-args makes it not fit well into a standard workspace build, where you can just type colcon build.

Is there some short-term workaround to avoid the bloat? And can the build be made so that the default colcon build just does what it should?

Yadunund:
@PaulBouchier thanks for the ticket. I can confirm the build dir for `zenoh_cpp_vendor` is ~4.3GB.

It's unclear whether this is "bloat" introduced by our build system or whether building Zenoh-c/cpp takes up this much space. We'd appreciate if you could investigate by building zenoh-cpp following the instructions upstream https://github.com/eclipse-zenoh/zenoh-cpp?tab=readme-ov-file#building-and-running-tests while passing along the flags as we do in rmw_zenoh, https://github.com/ros2/rmw_zenoh/blob/2e43a482ce82a475f4b561f0722bdbe7becd18c3/zenoh_cpp_vendor/CMakeLists.txt#L18-L40

Anyways, building code from source on RPi will likely lead to resource constraints. Why not rely on `rmw_zenoh` binaries instead?

PaulBouchier:
Thanks for the suggestion to use binaries. I did that and they work fine. I added README.md documentation about using binary packages to my PR. I'll respond here with findings from the investigation you requested.

Yadunund:
This is just a byproduct of rust builds which end up building all dependencies from source (no dynamic linking) so there's not much we can do about it. I'll close this out as there are workarounds like installing binaries, upgrading storage, etc.

PaulBouchier:
Sounds good. Thanks for the explanation. I'm sorry I didn't get to my
assignment.

Paul

Yadunund:
No worries! Thanks for your help with updating the README! Let us know if you run into any other issues with rmw_zenoh!

PR #660, opened 2025-06-05
Enable zenoh/multilink feature

Hugal31:
## Description

Enable the `zenoh/transport_multilink` feature.

Users will now be able to connect two nodes with more than one link. User not using this feature will not see any change.

No.

### Additional Information

I have seen multiple examples hinting the use of multilink (e.g. [the doc mentioning transport being chosen depending on the reliability](https://github.com/ros2/rmw_zenoh/blob/rolling/docs/design.md#quality-of-service) and [this example for an issue](https://github.com/ros2/rmw_zenoh/issues/623#issuecomment-2897449936)), yet, it was not enable.

### DIFF
--- a/zenoh_cpp_vendor/CMakeLists.txt
+++ b/zenoh_cpp_vendor/CMakeLists.txt
@@ -15,7 +15,7 @@ find_package(ament_cmake_vendor_package REQUIRED)
 # Note: We separate the two args needed for cargo with "$<SEMICOLON>" and not ";" as the
 # latter is a list separater in cmake and hence the string will be split into two
 # when expanded.
-set(ZENOHC_CARGO_FLAGS "--no-default-features$<SEMICOLON>--features=shared-memory zenoh/transport_compression zenoh/transport_tcp zenoh/transport_udp zenoh/transport_tls")
+set(ZENOHC_CARGO_FLAGS "--no-default-features$<SEMICOLON>--features=shared-memory zenoh/transport_compression zenoh/transport_multilink zenoh/transport_tcp zenoh/transport_udp zenoh/transport_tls")
 
 # Set VCS_VERSION to 1.4.0 commits of zenoh/zenoh-c/zenoh-cpp to benefit from:
 # - Add a "bind" config option for endpoints:


Yadunund:
Thanks for the pull request. 

We're thinking about how to best support more users who want to exploit capabilities of Zenoh in `rmw_zenoh`. Having PRs that add features one by one may not make the most sense.

We could have enabled all features by default but chose to selectively enable those that we strictly need to minimize the impact on the ROS Buildfarm as building `zenoh_cpp_vendor` is a time and memory consuming step. We're already seeing build and testing jobs fail due to timeouts from the increased buildtime. We've also seen complaints on excessive disk usage during builds https://github.com/ros2/rmw_zenoh/issues/447. 

One long-term strategy in mind is to not distribute `zenoh-cpp` via the Buildfarm and instead install `zenoh-cpp` debs distributed directly from the Official Ubuntu Repository or the Eclipse one. These debs will have all features enabled.

To address the concerns with increased memory and build time on the Buildfarm, it would help if you could benchmark the increase with all features enabled (instead of just the one in this PR). We can make a decision based on that.

In the meanwhile, our recommendation is to build `rmw_zenoh` from source by overwriting the `ZENOHC_CARGO_FLAGS` as needed.

JEnoch:
Here is a benchmark made with `rolling` branch in container on a laptop with this spec:
- Host: AMD Ryzen 9 7940HS, 16 cores, 32 GB RAM, SSD
- Docker image: ghcr.io/sloretz/ros:rolling-desktop
- Container limits: 4 CPUs, 16 GB RAM

| features                        | build time | build dir | libzenohc.so |
|---------------------------------|------------|-----------|--------------|
| current                         | 6min 15s   | 4.8 GB    | 21.0 MB      |
| current + multilink             | 6min 5s    | 4.9 GB    | 21.4 MB      |
| current + serial                | 5min 57s   | 4.9 GB    | 21.2 B       |
| current + quic                  | 6min 12s   | 4.9 GB    | 22.1 MB      |
| current + multilink+serial+quic | 6min 24s   | 5.1GB     | 22.7 MB      |
| ALL features                    | 6min 30s   | 5.2 GB    | 23.8 MB      |

Note that on several runs with the same features set, the time to build can vary by a few seconds, even if the host was dedicated to the build.

Yadunund:
@JEnoch thanks a lot for running these benchmarks. I believe the Buildfarm runners are also configured with 4 cores and 16GB of RAM. 

In light of this, I'm in favor of enabling all features by default. What do you think?

JEnoch:
I would no enable all features by default, since in the future `zenoh-c` might add some features that would impact performances (e.g. `stats` which makes zenoh to compute statistics).

I would go with [its default features](https://github.com/eclipse-zenoh/zenoh-c/blob/68936a0a6e1266c59ec79f34e5cb8036a2262fb0/Cargo.toml#L49C1-L60C2) which already include the features that are currently used in rmw_zenoh, and also `multilink` and `transport_quic`.
And I would add `shared-memory` and `transport_serial`.

I'm creating a new PR with this proposal.

PR #685, opened 2025-06-25
Only configure the timeout of the action-related service `get_result` to maximum value.

evshary:
## Description

Now in rmw_zenoh_cpp, the timeout for queries is configured to the maximum value.
This is because the service is also used in action, and the timeout of one of the services (get_result) can't be estimated.
https://design.ros2.org/articles/actions.html
However, this might introduce a memory-growing issue if users keep sending many queries, and some of them are missed.
Here is the result with the irobot benchmark tool:

image

I think we can check if the keyexpr matches the `**/_action/get_result/**` to avoid configuring all other services to the maximum timeout, which is not reasonable.

No

No

### Additional Information

The default timeout is now inherited from the configuration `queries_default_timeout`, which is 60000 ms by default.

### DIFF
--- a/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
+++ b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
@@ -215,8 +215,7 @@
 
   /// The default timeout to apply to queries in milliseconds.
   /// ROS setting: increase the value to avoid timeout at launch time with a large number of Nodes starting all together.
-  ///              Note: the requests to services and actions are hard-coded with an infinite timeout. Hence, this setting
-  ///              only applies to the queries made by the Advanced Subscriber for TRANSIENT_LOCAL implementation.
+  ///              Note that only action-related service get_result is hard-coded with an infinite timeout.
   queries_default_timeout: 60000,
 
   /// The routing strategy to use and it's configuration.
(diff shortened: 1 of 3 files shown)


JEnoch:
This PR looks good to me.
However I suggest to wait for #694 to be merged as it will conflict with this one:
the check on keyexpr will have to be made at Querier creation.

evshary:
@JEnoch Now the #694 is merged and the conflict is resolved. The advantage of using Querier is that we don't need to check the key_expr every time we send a query.

ahcorde (approved):

ahcorde:
Pulls: ros2/rmw_zenoh#685
Gist: https://gist.githubusercontent.com/ahcorde/9107645ef71f13ddcc9f1049f35796da/raw/2f7bed01f12821af163388655c9382fa83ebc030/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16406
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24198/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18281/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3633/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24805/)

JEnoch (approved):

ahcorde:

PR #363, opened 2024-12-27
Shared Memory on C++ API

yellowhatter:
SHM functionality implemented on C++ API

Transparently places ROS messages that are above configurable size threshold into shared-memory.

PR #724, opened 2025-07-28
Dev/1.0.0 cpp shm humble

yellowhatter:
SHM ported to Humble


JEnoch:
Closing since #363 has been merged into `rolling` and backported to humble via #743

PR #759, opened 2025-08-27
README: clarify shared memory config

JEnoch:
In `README.md`, clarify that the shared memory config shall be enabled for all the sessions and the router.
Also add a note on lowering the `ZENOH_SHM_MESSAGE_SIZE_THRESHOLD` value.


ahcorde (approved):

JEnoch:

PR #784, opened 2025-09-10
refactor: change the logging level of SHM allocation failure from DEBUG to WARN

YuanYuYuan:
This PR aims to explicitly inform users that messages are failing to be allocated in the SHM buffer. 

The performance overhead is acceptable since it only happens when SHM is enabled and the message is large enough. This is exactly the case to send with SHM.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
@@ -260,7 +260,7 @@ rmw_ret_t PublisherData::publish(
       shm_buf = std::make_optional(std::move(buf));
     } else {
       // Print a warning and revert to regular allocation
-      RMW_ZENOH_LOG_DEBUG_NAMED(
+      RMW_ZENOH_LOG_WARN_NAMED(
         "rmw_zenoh_cpp", "Failed to allocate a SHM buffer, fallback to non-SHM");
     }
   }


ahcorde (approved):

ahcorde:
Pulls: ros2/rmw_zenoh#784
Gist: https://gist.githubusercontent.com/ahcorde/b49da2f82e64653ff8d40a619b7d0d9c/raw/9c1ac1278b224f9e86423bdf28abd1e3655d1281/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16924
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24841/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18870/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4290/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25354/)

JEnoch:
I have a concern about this: 
in case of SHM congestion, the Node will log 1 warning per-message. For Lidar+Camera nodes, this could mean hundreds of logs per-seconds, flooding the logfile. 
While at DEBUG level that's kind of expected, at WARN level, which is likely enabled in production, I guess is could be problematic.

@Yadunund , @ahcorde what do you think ?

YuanYuYuan:
After discussion, we decided to postpone this until the Zenoh SHM API can return distinct failure results to clarify the cause:

- Out of pool size: If allocation fails due to insufficient pool size, warn users to increase the SHM arena.
- Other failures: If it fails under congestion or other conditions, report it in the debug logs.

Yadunund:
I agree with @JEnoch that this would lead to a lot of printouts for certain topics. Users can always set the desired log level at runtime.
Even with the distinction of error classes proposed, it would lead to a flood of printouts for the out of pool size situation. Hence, I'm inclined to leaving this as a debug log.

YuanYuYuan:
> I agree with @JEnoch that this would lead to a lot of printouts for certain topics. Users can always set the desired log level at runtime. Even with the distinction of error classes proposed, it would lead to a flood of printouts for the out of pool size situation. Hence, I'm inclined to leaving this as a debug log.

I just checked what FastDDS responds to this, and it seems to report nothing. So let's go ahead and close this.

PR #857, opened 2025-11-12
Use shared transport SHM provider instead of own instance of SHM provider

yellowhatter:
## Description

`rmw_zenoh` is allocating it's own shared memory segment, while Zenoh is also allocating one for implicit transport optimization. This means 2 segments are allocated while only 1 can be allocated.

This PR relies on Zenoh 1.7.0 that now exposes the shared memory provider used internally for transport optimization.
It makes `rmw_zenoh` to use this provider instead of creating its own.

As a consequence, the size of the allocated segment can be set via the `transport/shared_memory/transport_optimization/pool_size` configuration value in the Zenoh config file.

For backward compatibility, this PR keeps the support of both `ZENOH_SHM_ALLOC_SIZE` and `ZENOH_SHM_MESSAGE_SIZE_THRESHOLD` environment variables. When set, `rmw_zenoh` override the values set in the config file.

Only less shared memory segments used on the host.

### Additional Information

This PR can be backported to humble, jazzy and kilted.

In the futur L release, the `ZENOH_SHM_ALLOC_SIZE` and `ZENOH_SHM_MESSAGE_SIZE_THRESHOLD` environment variables can be deprecated or removed.


rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
```diff
@@ -763,9 +763,9 @@
           /// Implicitly puts large messages into shared memory for transports with SHM-compatible connection.
           enabled: true,
           /// SHM memory size in bytes used for transport optimization (default `16 * 1024 * 1024`).
-          pool_size: 16777216,
+          pool_size: 50331648,
```
evshary, line 776: It would be better if we could add `ROS setting:` in the comment like this:
https://github.com/ros2/rmw_zenoh/blob/684644f5020402928361055bb429d74ff2d70bff/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5#L137

zenoh_cpp_vendor/CMakeLists.txt
```diff
@@ -43,8 +43,8 @@ set(ZENOHC_CARGO_FLAGS "--features=shared-memory zenoh/transport_serial")
 # * Precommit SHM pages, avoiding overcommit and possible crash if SHM exhausts
 #   - https://github.com/eclipse-zenoh/zenoh/pull/2175
 ament_vendor(zenoh_c_vendor
-  VCS_URL https://github.com/eclipse-zenoh/zenoh-c.git
-  VCS_VERSION f376456ccf75ed837a21a186bdf5191cba50eb3b
+  VCS_URL https://github.com/ZettaScaleLabs/zenoh-c.git
```
evshary: I think you will switch back to `eclipse-zenoh` after the corresponding PR is merged, right?

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -51,66 +51,66 @@ namespace rmw_zenoh_cpp
 // by a RELIABLE + TRANSIENT_LOCAL Publisher
 #define SAMPLE_MISS_DETECTION_HEARTBEAT_PERIOD 500
 
-///=============================================================================
+  ///=============================================================================
```
evshary: Just out of curiosity, do you leave the indent here intentionally?

evshary (commented):
Some coding styles is changed, which is not related to the PR. I think it would be better to keep the same style.

zenoh_cpp_vendor/CMakeLists.txt
yellowhatter: Sure!

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
yellowhatter: I think this is auto-formatter artifact.

rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
yellowhatter: fixed

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
evshary: I think it would be better to keep the same coding style. I see many style changes in `rmw_publisher_data.cpp` are not related to the PR.

yellowhatter: Fixed!

zenoh_cpp_vendor/CMakeLists.txt
yellowhatter: switched

JEnoch (changes requested):
I think that for backward compatibility we need to keep support of both `ZENOH_SHM_ALLOC_SIZE` and `ZENOH_SHM_MESSAGE_SIZE_THRESHOLD` environment variable. They should override the config values `transport_optimization/pool_size` and `transport_optimization/message_size_threshold` respectively.

We could add a deprecation warning log when set, explaining those variables will no longer be supported in Lyrical Luth distro and that the `transport_optimization` config shall be used instead.
Then before Lyrical code branch (03/2026) we remove those environment variables from the rolling branch only.

The README.md has to be updated also with regard to this change.

yellowhatter:
> I think that for backward compatibility we need to keep support of both `ZENOH_SHM_ALLOC_SIZE` and `ZENOH_SHM_MESSAGE_SIZE_THRESHOLD` environment variable. They should override the config values `transport_optimization/pool_size` and `transport_optimization/message_size_threshold` respectively.
> 
> We could add a deprecation warning log when set, explaining those variables will no longer be supported in Lyrical Luth distro and that the `transport_optimization` config shall be used instead. Then before Lyrical code branch (03/2026) we remove those environment variables from the rolling branch only.
> 
> The README.md has to be updated also with regard to this change.

Done!

JEnoch:
As agreed during the dev meeting on Dec. 4th the `ZENOH_SHM_ALLOC_SIZE` and `ZENOH_SHM_MESSAGE_SIZE_THRESHOLD` environment variables must not be deprecated for Humble, Jazzy and Kilted distros. I removed the related warning logs.

In a separate PR, we will deprecate those with warning logs only in the Rolling branch.
Before the code freeze for Lyrical Luth we will decide if we remove or keep supporting those env vars.

JEnoch:
Pulls: ros2/rmw_zenoh#857
Gist: https://gist.githubusercontent.com/JEnoch/5eef46ff1b9843b247b848c0fb73a9cc/raw/6f7573209861da4266f852226b6127f1c03e84ee/ros2.repos
BUILD args: "--continue-on-error" --packages-above-and-dependencies zenoh_cpp_vendor zenoh_security_tools rmw_zenoh_cpp
TEST args:  --packages-above zenoh_cpp_vendor zenoh_security_tools rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17806
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26663/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20202/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6459/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26253/)

JEnoch (approved):

JEnoch:

zygfrydw:
@yellowhatter @JEnoch this pull request prohibits the [workaround](https://github.com/ros2/rmw_zenoh/issues/978) for SHM race condition [#2628](https://github.com/eclipse-zenoh/zenoh/issues/2628). I have reported issue   [#1030](https://github.com/ros2/rmw_zenoh/issues/1030)

PR #889, opened 2026-01-08
Update documentation for Zenoh timestamping configuration

jordanburklund:
## Description

Update documentation for Zenoh timstamping configurations for server and router configs. When testing locally with a simple "transient local" publisher and subscriber, the comment in the default configurations seems to be outdated. "transient local" topics were still working even when setting `timestamping: { enabled: false }` for both sending topics locally on the same machine and sending topics between two different machines over the network.

No

No

### Additional Information

### DIFF
--- a/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
+++ b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
@@ -205,6 +205,9 @@
   timestamping: {
     /// Whether data messages should be timestamped if not already.
     /// Accepts a single boolean value or different values for router, peer and client.
+    ///
+    /// ROS setting: timestamping is required for TRANSIENT_LOCAL + BEST_EFFORT publishers.
+    ///              See rmw_zenoh_cpp/docs/design.md - "Quality of Service" chapter - for details.
     enabled: { router: true, peer: true, client: true },
     /// Whether data messages with timestamps in the future should be dropped or not.
     /// If set to false (default), messages with timestamps in the future are retimestamped.
--- a/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
+++ b/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
@@ -205,9 +205,6 @@
   timestamping: {
     /// Whether data messages should be timestamped if not already.
     /// Accepts a single boolean value or different values for router, peer and client.
-    ///
-    /// ROS setting: PublicationCache which is required for transient_local durability
-    ///              only works when time-stamping is enabled.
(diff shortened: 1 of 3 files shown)


rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
```diff
@@ -205,9 +205,6 @@
   timestamping: {
     /// Whether data messages should be timestamped if not already.
     /// Accepts a single boolean value or different values for router, peer and client.
-    ///
-    /// ROS setting: PublicationCache which is required for transient_local durability
```
jordanburklund, line 209: I wasn't sure what to replace this with, so I left it as a WIP and will update it based on feedback from the maintainers.

JEnoch:
I agree that the original comment is outdated. The `PublicationCache` has been replaced by the `AdvancedPublisher` for `TRANSIENT_LOCAL` support.

However timestamping is still required, but only for `TRANSIENT_LOCAL`+`BEST_EFFORT` publishers. Let me explain:
At startup a Zenoh `AdvancedSubscriber` receives both historical data and live publications at the same time. It needs to reorder those data before to provide them to the callback API. He can do it, but only if one of the two conditions applies:
- Either a timestamp is attached with each data - that requires `timestamping` config to be enabled
- Either a sequence number and a source id are attached with each data - that's the case with `RELIABLE` Publishers that set a `miss_sample_detection` option for the `AdvancedPublisher`. See [here](https://github.com/ros2/rmw_zenoh/blob/94403d41c6015494c2c66645fe33098d8ef339d2/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp#L123-L131)

If `timestamping` is disabled and a `TRANSIENT_LOCAL`+`BEST_EFFORT` Publisher is created, Zenoh will exit with this error log:
```log
2026-06-01T15:58:54.671250Z ERROR ThreadId(10) zenohc::advanced_publisher: Cannot create AdvancedPublisher 0/chatter/std_msgs::msg::dds_::String_/RIHS01_df668c740482bbd48fb39d76a70dfd4bd59db1288021743503259e948f6b1a18 with Sequencing::Timestamp: the 'timestamping' setting must be enabled in the Zenoh configuration. at /root/.cargo/git/checkouts/zenoh-cc237f2570fab813/2687c51/zenoh-ext/src/advanced_publisher.rs:368.
[ERROR 1780329534.671329133] [rmw_zenoh_cpp]: Unable to make PublisherData. (Data() at /home/ubuntu/rmw_zenoh/src/rmw_zenoh/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp:265)
```

You can see this running `ros2 topic pub /chatter std_msgs/String "data: Hello World" --qos-durability transient_local  --qos-reliability best_effort`

JEnoch:
@jordanburklund I amended your PR to add an explanation wrt. timestamping requirement in `docs/design.md`, and to clarify the comment in config files.
Does it make sense for you ?

PR #908, opened 2026-02-15
Allow use of non-vendored Zenoh if present

oysstu:
## Description

Uses zenoh-c and zenoh-cpp if found by cmake. It's easier to handle distributing and patching of Zenoh on non-standard platforms if the rust build can be handled externally.

Should not affect any users, unless they are building from source and have installed the libraries to the cmake search path.

No

### Additional Information
There's also https://github.com/ros2/rmw_zenoh/pull/695 but it seems to have stalled. Would appreciate if this could be considered in the meantime.

### DIFF
--- a/zenoh_cpp_vendor/CMakeLists.txt
+++ b/zenoh_cpp_vendor/CMakeLists.txt
@@ -10,6 +10,9 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_vendor_package REQUIRED)
 
+find_package(zenohc QUIET)
+find_package(zenohcxx QUIET)
+
 # Add a set of usefull features that are not enabled by default in zenoh-c.
 # For a complete list of features see: https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/Cargo.toml
 # Note: the list of features are separated with whitespaces.
@@ -29,6 +32,7 @@ set(ZENOHC_CARGO_FLAGS "--features=shared-memory zenoh/transport_serial")
 # * Metrics per key (requires to add "zenoh/stats" feature to ZENOHC_CARGO_FLAGS for build)
 #   - https://github.com/eclipse-zenoh/zenoh/pull/2284
 ament_vendor(zenoh_c_vendor
+  SATISFIED ${zenohc_FOUND}
   VCS_URL https://github.com/eclipse-zenoh/zenoh-c.git
   VCS_VERSION 73e7a0bdcf78b8263a02b7aee453ee6de8187589
   CMAKE_ARGS
@@ -41,13 +45,16 @@ ament_vendor(zenoh_c_vendor
 ament_export_dependencies(zenohc)
 
 ament_vendor(zenoh_cpp_vendor
+  SATISFIED ${zenohcxx_FOUND}
   VCS_URL https://github.com/eclipse-zenoh/zenoh-cpp
   VCS_VERSION 47a80d345d74ebef16f3783423bf3bbd89cd6c30
   CMAKE_ARGS
     -DZENOHCXX_ZENOHC=OFF
 )
 
-externalproject_add_stepdependencies(zenoh_cpp_vendor configure zenoh_c_vendor)
+if(NOT zenohcxx_FOUND)
+  externalproject_add_stepdependencies(zenoh_cpp_vendor configure zenoh_c_vendor)
+endif()
 
 ament_export_dependencies(zenohcxx)
 
--- a/zenoh_cpp_vendor/CMakeLists.txt
+++ b/zenoh_cpp_vendor/CMakeLists.txt
@@ -10,52 +10,53 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_vendor_package REQUIRED)
 
-find_package(zenohc QUIET)
-find_package(zenohcxx QUIET)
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

ahcorde:
Pulls: ros2/rmw_zenoh#908
Gist: https://gist.githubusercontent.com/ahcorde/4a89ff63bbd4d0605513a0531b729473/raw/c804702c650470234b0499c77d5d9dd09464593f/ros2.repos
BUILD args:  --packages-above-and-dependencies zenoh_cpp_vendor
TEST args:  --packages-above zenoh_cpp_vendor
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18209
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27670/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20726/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7547/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26689/)

JEnoch (commented):
I believe using non-vendored Zenoh is useful but should be an option, and possibly not the default behaviour. Imagine a host where a version of `zenoh-c` is installed for non-ROS applications, but `rmw_zenoh` requires another version. How to deal with that ?

The commit ids which are set in `zenoh_cpp_vendor/CMakeLists.txt` are the ones that maintainers ensure `rmw_zenoh` will work as expected. I think using another version should be made on purpose with an explicit flag.

And it might be useful to log (info or warning) whenever another zenoh-c or zenoh-cpp version than the vendor one is used.

oysstu:
> I believe using non-vendored Zenoh is useful but should be an option, and possibly not the default behaviour. Imagine a host where a version of `zenoh-c` is installed for non-ROS applications, but `rmw_zenoh` requires another version. How to deal with that ?
> 
> The commit ids which are set in `zenoh_cpp_vendor/CMakeLists.txt` are the ones that maintainers ensure `rmw_zenoh` will work as expected. I think using another version should be made on purpose with an explicit flag.
> 
> And it might be useful to log (info or warning) whenever another zenoh-c or zenoh-cpp version than the vendor one is used.

That seems reasonable, I've updated the PR. 

I decided to make it fail if USE_SYSTEM_ZENOH is set and it does not find the system installation instead of falling back on the vendored libraries.

zenoh_cpp_vendor/CMakeLists.txt
```diff
@@ -10,45 +10,53 @@ endif()
...
+  find_package(zenohcxx REQUIRED)
+  message(STATUS "Using system zenohc version: ${zenohc_VERSION}")
+  message(STATUS "Using system zenohcxx version: ${zenohcxx_VERSION}")
+else()
+  # Add a set of usefull features that are not enabled by default in zenoh-c.
+  # For a complete list of features see: https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/Cargo.toml
+  # Note: the list of features are separated with whitespaces.
+  # However, if you want to add other cargo build flags, use "$<SEMICOLON>" as separator and not ";" as the
+  # latter is a list separater in cmake and hence the string will be split into two when expanded.
+  set(ZENOHC_CARGO_FLAGS "--features=shared-memory zenoh/transport_serial")
```
JEnoch, line 26: This `ZENOHC_CARGO_FLAGS` should be set for both non-vendored and vendored cases.

JEnoch (changes requested):

JEnoch (commented):
A last comment: could you please also add a note in the README (in "Source Installation") wrt. this new `USE_SYSTEM_ZENOH` flag.

zenoh_cpp_vendor/CMakeLists.txt
oysstu: Would this have an effect when the system libraries are already built? It would be up to the user to ensure that the provided zenoh library is built with the unstable api and shared memory support (otherwise rmw_zenoh_cpp will fail to compile).

JEnoch: Oh! Indeed you're right, my bad.

JEnoch:
Pulls: ros2/rmw_zenoh#908
Gist: https://gist.githubusercontent.com/JEnoch/7bba8c32677d4dce5a5f3c6a8b1ca1a1/raw/c804702c650470234b0499c77d5d9dd09464593f/ros2.repos
BUILD args: "--continue-on-error" --packages-above-and-dependencies zenoh_cpp_vendor zenoh_security_tools rmw_zenoh_cpp
TEST args:  --packages-above zenoh_cpp_vendor zenoh_security_tools rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18214
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27675/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20731/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7552/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26694/)

JEnoch (approved):

JEnoch:

PR #930, opened 2026-03-17
Add support for rosidl::Buffer-aware per-endpoint pub/sub

nvcyc:
## Description

This pull request adds full `rosidl::Buffer` support to `rmw_zenoh_cpp`, enabling per-endpoint Zenoh publishers and subscribers for zero-copy buffer transport between compatible backends. When a publisher and subscriber share compatible non-CPU buffer backends, data can be transferred via a lightweight descriptor; when backends are incompatible, the system falls back to standard CPU-based buffer serialization.

This pull request consists of the following key changes:

- **Backend lifecycle**: Calls `rosidl_buffer_backend_registry::initialize_buffer_backends()` / `shutdown_buffer_backends()` during RMW init/shutdown to load and tear down buffer backend plugins.
- **Liveliness key-expression extension**: Extended key-expressions to advertise each endpoint's supported backends, enabling dynamic discovery.
- **Graph cache discovery callbacks**: Buffer-aware publishers and subscribers register discovery callbacks to detect each other and dynamically create per-endpoint Zenoh publishers/subscribers.
- **Buffer-aware publishers**: Create per-subscriber Zenoh endpoints; endpoint info is passed to the typesupport serialization layer, which delegates compatibility to each backend's `create_descriptor_with_endpoint()` (nullptr: CPU fallback). Publisher creation explicitly adds `"cpu"` to `backend_aux_info`.
- **Fallback publish**: `publish()` first sends endpoint-aware messages via `publish_buffer_aware()`, then conditionally falls through to the standard base-key publish path only when the total matched subscription count exceeds discovered buffer-aware subscribers -- avoiding unnecessary CPU conversion when all subscribers are buffer-aware.
- **Buffer-aware subscribers**: Create per-publisher Zenoh subscriptions; the `Message` struct owns endpoint info via `std::optional<EndpointInfoStorage>` Endpoint info is passed into deserialization for correct backend reconstruction.
- **`acceptable_buffer_backends`**: Parses the subscription option -- `NULL`/empty/`"cpu"`: CPU-only (advertises `"cpu"` in liveliness token); `"any"`: all installed; specific names: filtered. In `on_publisher_discovered()`, CPU is always added to the publisher's backend list.

This pull request does not change existing `rmw_zenoh_cpp` behavior for standard (non-Buffer) messages. For messages with `uint8[]` fields, the per-endpoint transport is transparent -- publishers and subscribers share backend info automatically, and CPU fallback ensures correctness when backends are incompatible.

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with creating an initial prototype version of the changes contained in this PR.
### Additional Information

This PR is part of the broader ROS 2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).


rmw_zenoh_cpp/src/detail/graph_cache.cpp
```diff
@@ -185,14 +185,76 @@ void GraphCache::update_topic_map_for_put(
...
     if (is_pub) {
       existing_graph_topic->pubs_.insert(entity);
+
+      // Trigger publisher discovery callbacks (notify subscribers about this publisher)
+      std::lock_guard<std::mutex> disc_lock(discovery_mutex_);
+      if (publisher_discovery_callbacks_.count(graph_topic_data->info_.name_)) {
+        for (const auto & [gid_hash,
+          callback] : publisher_discovery_callbacks_[graph_topic_data->info_.name_])
+        {
+          callback(*entity);
```
mjcarroll: I think there is a risk of this callback trying to do something with the graph that is also under the `discovery_mutex_` (or under a lock further down in DDS) and we could end up with a deadlock.

The better approach is probably to follow the `BufferEndpointRegistry::register_subscriber_discovery_callback` pattern where you push all of the callbacks into a temporary vector, release the lock, and then iterate them.

hidmic: +1 to that. Same elsewhere.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -466,6 +526,29 @@ Entity::Entity(
     keyexpr_parts[KeyexprIndex::TopicType] = mangle_name(topic_info.type_);
     keyexpr_parts[KeyexprIndex::TopicTypeHash] = mangle_name(topic_info.type_hash_);
     keyexpr_parts[KeyexprIndex::TopicQoS] = qos_to_keyexpr(topic_info.qos_);
+
+    // Add backend metadata if present (only for Buffer message types)
+    if (topic_info.backend_metadata_.has_value() && !topic_info.backend_metadata_.value().empty()) {
+      std::vector<std::string> backend_names;
+      backend_names.reserve(topic_info.backend_metadata_->size());
+      for (const auto & pair : topic_info.backend_metadata_.value()) {
```
hidmic: @nvcyc nit: would it make sense to get a reference to the `backend_metadata` value and reuse?

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
```diff
@@ -586,7 +669,8 @@ std::shared_ptr<Entity> Entity::make(const std::string & keyexpr)
 
   // Populate topic_info if we have a token for an entity other than a node.
   if (entity_type != EntityType::Node) {
-    if (parts.size() < KEYEXPR_INDEX_MAX + 1) {
+    // Minimum required: up to TopicQoS (backends is optional)
+    if (parts.size() < KeyexprIndex::TopicQoS + 1) {
```
hidmic: @nvcyc meta: perhaps there should be a `MANDATORY_KEYEXPR_INDEX_MAX` macro to future proof this clause.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -202,9 +308,152 @@ PublisherData::PublisherData(
...
   events_mgr_ = std::make_shared<EventsManager>();
+
+  // For simple publishers, create a single base endpoint
+  if (!is_buffer_aware_) {
+    auto base_endpoint = std::make_shared<PublisherEndpoint>();
+    base_endpoint->key = entity_->topic_info()->topic_keyexpr_;
+    base_endpoint->pub = std::optional<zenoh::ext::AdvancedPublisher>(std::move(pub_));
+    endpoints_[base_endpoint->key] = base_endpoint;
+  }
+  // For buffer-aware publishers, endpoints are created dynamically on subscriber discovery
```
hidmic, line 313: @nvcyc meta: won't this interfere with system introspection? IIRC `ros2 topic list` need not subscribe to publishers to find them.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -202,9 +308,152 @@ PublisherData::PublisherData(
...
+      continue;
+    }
+    auto endpoint = endpoint_it->second;
+
+    size_t max_data_length = type_support_->get_estimated_serialized_size(
+      ros_message, type_support_impl_);
+
+    // Quadruple the buffer size for safety (endpoint-aware serialization needs more space)
+    // TODO(native-buffer): Fix the size estimation in buffer_serialization.hpp to be more accurate
+    max_data_length = max_data_length * 4 + 16384;
```
hidmic: @nvcyc maybe something along the lines of having a maximum size for descriptors would work here?

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -218,6 +467,31 @@ rmw_ret_t PublisherData::publish(
...
+  // Buffer-aware publishers: send endpoint-aware messages to per-subscriber
+  // endpoints, then conditionally fall through to the standard path to also
+  // publish on the base key.  Buffer-aware subscribers only listen on
+  // per-publisher key expressions, so the base publication only reaches
+  // non-buffer-aware subscribers (no duplicates).
+  //
+  // We only fall through when the total matched subscription count exceeds
+  // the number of discovered buffer-aware subscribers, meaning at least one
+  // non-buffer-aware subscriber exists.  This avoids an unnecessary CPU
+  // conversion (to_vector()) of vendor-backed buffer data on every publish.
```
hidmic: @nvcyc meta: isn't this a different approach compared to the one chosen for `rmw_fastrtps_cpp`? IIUC the latter skips buffer aware publishing entirely if it discovers non-buffer aware subscribers.

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
```diff
@@ -18,35 +18,87 @@
...
 #include "rmw/error_handling.h"
 #include "rmw/get_topic_endpoint_info.h"
 #include "rmw/impl/cpp/macros.hpp"
+#include "rosidl_runtime_c/type_hash.h"
+
+#include "rosidl_typesupport_fastrtps_cpp/message_type_support.h"
+
+namespace
+{
+std::string gid_to_hex(const rmw_gid_t & gid)
```
hidmic: @nvcyc hmm, aren't these functions duplicated in `rmw_publisher_data.cpp`?

hidmic (commented):
First pass. Partial pass. I need more brain to parse all of this.

ahcorde:
@YuanYuYuan or @JEnoch do you mind to take a look ? mw freeze it's next monday (6th April)

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
nvcyc: Great question. You caught this diff when we updated `rmw_fastrtps_cpp` and then were in the progress of updating `rmw_zenoh_cpp`. This part of the logics is now the same with new commits pushed.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
nvcyc: In rmw_zenoh, publisher discovery (used by `ros2 topic list`, `ros2 topic info`, etc.) works entirely through liveliness tokens and the graph cache, therefore `ros2 topic list` will not be affected by these changes in the Zenoh data-plane publisher/subscriber creation.

rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
```diff
@@ -0,0 +1,167 @@
...
+    std::string backend_type = backend->get_backend_type();
+    RCUTILS_LOG_DEBUG_NAMED(
+      kLoggerName, "Processing backend '%s' (type: %s)",
+      backend_name.c_str(), backend_type.c_str());
+
+    context.backend_instances[backend_type] = backend;
+
+    rosidl::BufferDescriptorOps ops;
+
+    auto backend_ptr = backend;
```
wjwwood: I feel like this is violating the "_almost_ always auto" ideal. I love `auto` as much as the next person, but here, this is hiding important information, which isn't impossible to deduce from the context, but I'm left frustrated wondering, is this a shared ptr, a unique ptr, something else? It's definitely a nitpick, but I think it would be better to avoid using `auto` unless it's very obvious from the right-hand side of the `=`.

In this particular case the type is important because we're capturing it in the next lambda.

rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
```diff
@@ -0,0 +1,167 @@
...
+    ops.create_descriptor_with_endpoint = [backend_ptr](
```
wjwwood: Is it safe to capture this (what I assume is a shared_ptr) in this lambda? I get nervous any time I see a shared pointer being captured by value in a lambda, because it extends the life of the ptr to be the same as the lambda. So feel free to resolve this if it's not the case, but maybe this should be captured as a weak_ptr and locked in the callback each time.

wjwwood (commented):
Partial review

JEnoch:
> @YuanYuYuan or @JEnoch do you mind to take a look ? mw freeze it's next monday (6th April)

Apologies for not having been able to attend the working group sessions or catch up on the discussions over the past few weeks. Unfortunately I still don't have bandwidth to do a thorough review of this PR at this point. I suspect the same is true for my colleague @YuanYuYuan.

I just have one suggestion:  it would be great if `docs/design.md` could be updated as part of this PR to document the new buffer-aware pub/sub architecture and the changes to all key expression formats (liveliness tokens and topics).

If the other reviewers are satisfied, I'm happy to defer to their judgment and approve the merge.

YuanYuYuan:
> @YuanYuYuan or @JEnoch do you mind to take a look ? mw freeze it's next monday (6th April)

Just returned from a long holiday :smile: I went through the deadlock issue and verified it last week. A patch has been proposed at https://github.com/ros2/rmw_zenoh/pull/955

asymingt:
Just a quick note that we are in a RMW freeze right now for the ROS 2 Lyrical release. Please reach out to @sloretz before you merge this one.

nvcyc:
Thanks for the reminder.
Yes, I'm aware of the current RMW freeze state for the ROS 2 Lyrical release.

Based on the discussion in the ROS 2 Lyrical release working group meeting on 2026/4/6, we'll target this PR for Lyrical patch 1 release to match the same level of `rosidl::Buffer` support `rmw_fastrtps_cpp` currently has, so I'm continuing the development here mainly for that goal.

I'll confirm in the working group with @sloretz before merging this PR when it's ready.

nvcyc:
Pulls: ros2/rmw_zenoh#930
Gist: https://gist.githubusercontent.com/nvcyc/4facd927bf9cb229ad82de132920f12e/raw/360295ca2557d4167aea0499f2665a75e408a3c8/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19244
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29048/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21845/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8918/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27950/)

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
nvcyc: Good suggestion. Updated to get and reuse the reference.

rmw_zenoh_cpp/src/detail/liveliness_utils.cpp
nvcyc: Added `MANDATORY_KEYEXPR_INDEX_MAX` as suggested.

rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp
nvcyc: Good catch! Common functions have been moved to a helper header file: [rmw_zenoh_cpp/src/detail/buffer_endpoint_helpers.hpp](https://github.com/ros2/rmw_zenoh/blob/native_buffer/rmw_zenoh_cpp/src/detail/buffer_endpoint_helpers.hpp)

rmw_zenoh_cpp/src/detail/graph_cache.cpp
nvcyc: Agreed. This has been taken care of in @YuanYuYuan's PR https://github.com/ros2/rmw_zenoh/pull/955 merged into this PR.

nvcyc:
Pulls: ros2/rmw_zenoh#930
Gist: https://gist.githubusercontent.com/nvcyc/58f6bd3266a2b466d2e3c9a025572c89/raw/360295ca2557d4167aea0499f2665a75e408a3c8/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19254
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29059/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21855/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8928/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27968/)

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
nvcyc: Agreed. I've updated this part to align with `rmw_fastrtps_cpp` to get estimated size from calling `message_type_support_callbacks_t`'s `get_serialized_size()` (which is invoked inside `type_support_->get_estimated_serialized_size()`) that handles `rosidle::Buffer` size in a standard way.

rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
nvcyc: Good point. Replaced the relevant auto declarations (backend, backend_names, desc, the lambda captures) with explicit types so it's obvious that we're carrying around a `std::shared_ptr<rosidl::BufferBackend>` and a `std::vector<std::string>`.

rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
nvcyc: That's a valid point.
It is not an issue here because the backend is owned by `BufferBackendContext::backend_instances`, and `shutdown_buffer_backends()` clears the descriptor ops/serializers before the instances map, so today the lifetime is bounded.

Yet, I agree with your suggestion and capturing by weak_ptr is much more robust.
I switched all three lambdas (`create_descriptor_with_endpoint`, from_descriptor_with_endpoint, descriptor deserialize) to capture a `std::weak_ptr<rosidl::BufferBackend>` and `lock()` it inside the callback, throwing `std::runtime_error` if it's invoked after backend tear-down.

nvcyc:
Pulls: ros2/rmw_zenoh#930
Gist: https://gist.githubusercontent.com/nvcyc/779e2927eb344675e062b7588954a13f/raw/360295ca2557d4167aea0499f2665a75e408a3c8/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19258
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29064/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21859/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8932/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27978/)

nvcyc:
Pulls: ros2/rmw_zenoh#930
Gist: https://gist.githubusercontent.com/nvcyc/03c2ecfd49160673f25e2f95486d7e80/raw/360295ca2557d4167aea0499f2665a75e408a3c8/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19262
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21863/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8936/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27983/)

nvcyc:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21873/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27986/)

nvcyc:
Pulls: ros2/rmw_zenoh#930
Gist: https://gist.githubusercontent.com/nvcyc/1af6f816c8361a5c556fd017f0038f0b/raw/360295ca2557d4167aea0499f2665a75e408a3c8/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19387
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29203/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21994/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9086/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28155/)

nvcyc:
I rebased the PR and triggered CI tests again (as shown above).
Two of the platoforms passed and it seems the remaining test failures are unrelated to the changes here in rmw_zenoh.
I think this PR should be in a good shape now. @YuanYuYuan

YuanYuYuan (approved):
All the review comments have been resolved and rebased onto the latest rolling.

YuanYuYuan:
Thanks for this @nvcyc! The `Check logging macros` CI job is failing on two distinct things — one is a real fix needed here, the other is a false positive being handled separately.

### 1. `buffer_backend_loader.cpp` — needs a change

This file includes `rcutils/logging_macros.h` and calls `RCUTILS_LOG_DEBUG_NAMED` / `RCUTILS_LOG_ERROR_NAMED` directly. Those route over `/rosout` and can deadlock under concurrent graph callbacks, which is exactly what the check guards against (#182).

The rest of the rosidl_buffer work in this PR already uses the dedicated `RMW_ZENOH_ROSIDL_BUFFER_LOG_*` wrappers from `detail/logging_macros.hpp` (the ones tagged "to be removed when rosidl_buffer is finalized") — see `rmw_publisher_data.cpp`, `rmw_subscription_data.cpp`, and `graph_cache.cpp`. Since `buffer_backend_loader.cpp` is squarely rosidl_buffer code, it should use the same buffer-specific wrappers for consistency. The call signatures are identical, so it's a drop-in swap plus an include change:

```diff
 #include "buffer_backend_loader.hpp"

+#include "logging_macros.hpp"
+
 #include <memory>
 ...
 #include "rcutils/error_handling.h"
-#include "rcutils/logging_macros.h"
 #include "rosidl_buffer_backend_registry/buffer_backend_registry.hpp"
```

Then at the 8 call sites:
- `RCUTILS_LOG_DEBUG_NAMED(` → `RMW_ZENOH_ROSIDL_BUFFER_LOG_DEBUG_NAMED(`
- `RCUTILS_LOG_ERROR_NAMED(` → `RMW_ZENOH_ROSIDL_BUFFER_LOG_ERROR_NAMED(`

To apply the whole change in one step, from the repo root on this PR's branch:

```bash
curl -sL https://gist.githubusercontent.com/YuanYuYuan/e3266d5723a5673980b121a983226c3c/raw/pr930-buffer-backend-loader-logging.patch | git apply
```

<details>
<summary>Full patch (applies cleanly on the current PR head)</summary>

```diff
diff --git a/rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp b/rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
--- a/rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
+++ b/rmw_zenoh_cpp/src/detail/buffer_backend_loader.cpp
@@ -14,6 +14,8 @@

 #include "buffer_backend_loader.hpp"

+#include "logging_macros.hpp"
+
 #include <memory>
 #include <cstring>
 #include <string>
@@ -22,7 +24,6 @@
 #include <vector>

 #include "rcutils/error_handling.h"
-#include "rcutils/logging_macros.h"
 #include "rosidl_buffer_backend_registry/buffer_backend_registry.hpp"
 #include "rosidl_typesupport_fastrtps_cpp/identifier.hpp"
 #include "rosidl_typesupport_fastrtps_cpp/message_type_support_decl.hpp"
@@ -42,20 +43,20 @@ void initialize_buffer_backends(BufferBackendContext & context)
   auto & backend_ops = context.serialization_context.descriptor_ops;
   auto & serializers = context.serialization_context.descriptor_serializers;
   std::vector<std::string> backend_names = registry.get_backend_names();
-  RCUTILS_LOG_DEBUG_NAMED(
+  RMW_ZENOH_ROSIDL_BUFFER_LOG_DEBUG_NAMED(
     kLoggerName, "Buffer backends: found %zu backend(s)", backend_names.size());

   for (const std::string & backend_name : backend_names) {
     std::shared_ptr<rosidl::BufferBackend> backend =
       registry.create_backend_instance(backend_name);
     if (!backend) {
-      RCUTILS_LOG_ERROR_NAMED(
+      RMW_ZENOH_ROSIDL_BUFFER_LOG_ERROR_NAMED(
         kLoggerName, "Backend '%s' pointer is null", backend_name.c_str());
       continue;
     }

     std::string backend_type = backend->get_backend_type();
-    RCUTILS_LOG_DEBUG_NAMED(
+    RMW_ZENOH_ROSIDL_BUFFER_LOG_DEBUG_NAMED(
       kLoggerName, "Processing backend '%s' (type: %s)",
       backend_name.c_str(), backend_type.c_str());

@@ -98,7 +99,7 @@ void initialize_buffer_backends(BufferBackendContext & context)
     const rosidl_message_type_support_t * descriptor_ts =
       backend->get_descriptor_type_support();
     if (!descriptor_ts) {
-      RCUTILS_LOG_ERROR_NAMED(
+      RMW_ZENOH_ROSIDL_BUFFER_LOG_ERROR_NAMED(
         kLoggerName,
         "  Backend '%s' returned null descriptor type support",
         backend_type.c_str());
@@ -107,7 +108,7 @@ void initialize_buffer_backends(BufferBackendContext & context)
     const rosidl_message_type_support_t * fastrtps_descriptor_ts = get_message_typesupport_handle(
       descriptor_ts, rosidl_typesupport_fastrtps_cpp::typesupport_identifier);
     if (!fastrtps_descriptor_ts) {
-      RCUTILS_LOG_ERROR_NAMED(
+      RMW_ZENOH_ROSIDL_BUFFER_LOG_ERROR_NAMED(
         kLoggerName,
         "  Backend '%s' descriptor type support could not be resolved to "
         "rosidl_typesupport_fastrtps_cpp",
@@ -119,7 +120,7 @@ void initialize_buffer_backends(BufferBackendContext & context)
     const auto * callbacks = static_cast<const message_type_support_callbacks_t *>(
       fastrtps_descriptor_ts->data);
     if (!callbacks) {
-      RCUTILS_LOG_ERROR_NAMED(
+      RMW_ZENOH_ROSIDL_BUFFER_LOG_ERROR_NAMED(
         kLoggerName,
         "  Backend '%s' descriptor callbacks are null",
         backend_type.c_str());
@@ -169,7 +170,7 @@ void initialize_buffer_backends(BufferBackendContext & context)
       };
     serializers[backend_type] = std::move(desc_ser);

-    RCUTILS_LOG_DEBUG_NAMED(
+    RMW_ZENOH_ROSIDL_BUFFER_LOG_DEBUG_NAMED(
       kLoggerName, "  FastCDR descriptor serializers registered for '%s'",
       backend_type.c_str());
   }
@@ -183,7 +184,7 @@ void shutdown_buffer_backends(BufferBackendContext & context)
     context.backend_instances.clear();
     context.registry.reset();
   } catch (const std::exception & e) {
-    RCUTILS_LOG_ERROR_NAMED(
+    RMW_ZENOH_ROSIDL_BUFFER_LOG_ERROR_NAMED(
       kLoggerName, "Error during buffer backend shutdown: %s", e.what());
   }
 }
```
</details>

### 2. `logging_macros.hpp` — false positive, no change needed

The check also flags the two new `RMW_ZENOH_ROSIDL_BUFFER_LOG_*` macro definitions, because they expand to `RCUTILS_LOG_SEVERITY_DEBUG`/`_ERROR` and the guard matches the bare substring `RCUTILS_LOG_`. That's the severity *enum*, not the logging call — it's exactly how the existing `RMW_ZENOH_LOG_*` wrappers are defined, so it's a false positive in the check itself, not a problem in your code.

I've opened ros2/rmw_zenoh#979 to make the guard precise (match the call macros and the include, but allow the `RCUTILS_LOG_SEVERITY_*` enum). Once that merges, please rebase on `rolling` — your `logging_macros.hpp` additions will pass as-is, and only the `buffer_backend_loader.cpp` change above will be needed.

nvcyc:
Pulls: ros2/rmw_zenoh#930
Gist: https://gist.githubusercontent.com/nvcyc/86785788d2e8fe2c8bd4e9dd63cfbc99/raw/360295ca2557d4167aea0499f2665a75e408a3c8/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rmw_zenoh_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19451
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29282/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22073/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9168/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28228/)

nvcyc:
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9179/)

Triggered `Linux-rhel` again to pick up a fix for the failure (unrelated to changes in this PR) observed in the CI failure in previous run.

As a result, all Linux-based platforms have CI passed.

The CI test failures on Windows are unrelated to changes in this PR as those failures are also observed in other unrelated CI runs (like this [one](https://ci.ros2.org/job/ci_windows/28235/#showFailuresLink)).

@YuanYuYuan I fixed the logging issue as you suggested and rebased. Then, based on the CI test results shown above, I think this PR is ready.

YuanYuYuan:
We finally made it. Many thanks to all the reviewers, and special kudos to the author, @nvcyc!

YuanYuYuan:
https://github.com/Mergifyio backport lyrical

mjcarroll:
Thanks for the iteration here folks! People will be excited to get using this.

Yadunund:
Great work @nvcyc 🙌🏼

PR #946, opened 2026-04-03
Fix issue following bump to Zenoh 1.8.0

JEnoch:
## Description

The bump to Zenoh 1.8.0 made in https://github.com/ros2/rmw_zenoh/pull/935 brings an issue related to a change in zenoh-c. When building with zenoh-c with `ZENOHC_MSRV_1_75` flag, its `Cargo.lock` is ignored, leading to several build issues:
- the latest Rust dependencies are used and some are not supporting Rust 1.75
- the dependency to zenoh is using it's `main` branch instead of a commit for 1.8.0
 
We discussed within Zenoh team today, and our conclusion is that until ROS for Ubuntu can get rid of Rust 1.75, we need to maintain distinct `Cargo.lock` in zenoh-c, for Rust 1.75 and Rust 1.88+ which is our current MSRV without pinning crates.
Rather than tinkering the build to swap the `Cargo.lock` depending on Rust version, we prefer to maintain 2 branches for the short term - i.e. until a solution is found to allow the bump of Rust version in Ubuntu:
- [main](https://github.com/eclipse-zenoh/zenoh-c/tree/main) for projects using recent Rust
- [ROS/rust-1.75](https://github.com/eclipse-zenoh/zenoh-c/tree/ROS/rust-1.75) for ROS humble, jazzy and kilted

This PR make `rmw_zenoh` to use commit https://github.com/eclipse-zenoh/zenoh-c/commit/0e3ec084142382096d09f7fc2ad59385998b92e6 from branch [ROS/rust-1.75](https://github.com/eclipse-zenoh/zenoh-c/tree/ROS/rust-1.75).
It depends on https://github.com/eclipse-zenoh/zenoh/commit/f27282e78c6e42e04249349049fa68cfbc941369

This PR must be merged to humble, jazzy and kilted branches.

When the rolling branch will be switched to use Ubuntu resolute and for the incoming lyrical branch, `zenoh_cpp_vendor/CMakeLists.txt` shall be changed to use https://github.com/eclipse-zenoh/zenoh-c/commit/c1a477d3f5efc3abc0f0cf95e0773225d00dea3b (still without the `ZENOHC_MSRV_1_75`)

PR #945, opened 2026-04-02
Build against rust >= 1.75 for ROS Lyrical

sloretz:
## Description

Currently zenoh_cpp_vendor sets `ZENOHC_MSRV_1_75` for compatibility with rust 1.75. This triggers logic that ignores Cargo.lock added in https://github.com/eclipse-zenoh/zenoh-c/pull/1237 , which results in static_init 1.0.3 being used instead of 1.0.4

Starting in ROS Lyrical the target rust version is 1.85, so it seems like this flag is no longer necessary.

Fixes #936

### Did you use Generative AI?

I asked Gemini to investigate why static_init 1.0.3 was being used, and I had it write the `get_cargo_version` function.

PR #947, opened 2026-04-05
fix: update wrong hash for zenoh-c 1.8.0

wep21:
## Description

zenoh-c/cpp hash is not correct with release/1.8.0.
https://github.com/eclipse-zenoh/zenoh-c/tree/release/1.8.0

Fixes # (issue)


zenoh_cpp_vendor/CMakeLists.txt
wep21:
```suggestion
    VCS_VERSION c3f98b7fad41bd91b7eb845d7853faadaf2249a5
```

JEnoch:
This was on purpose to include https://github.com/eclipse-zenoh/zenoh/pull/2493 which landed just after 1.8.0 release.

There was another issue with those commit ids explained in https://github.com/ros2/rmw_zenoh/pull/946.
But it's now fixed in #945, along another concern wrt. resolute.

PR #955, opened 2026-04-06
fix(graph_cache): release graph_mutex_ before invoking discovery callbacks

YuanYuYuan:
## Summary

`update_topic_map_for_put()` collected discovery callbacks under `discovery_mutex_` (correct)
but then invoked them while `graph_mutex_` was still held via the `lock_guard` in `parse_put()`.
Any callback that re-enters `GraphCache` — for example, creating a per-endpoint subscription
which calls `register_publisher_discovery_callback()` — would attempt to re-acquire
`graph_mutex_` on the same thread and deadlock immediately.

A ROS 2 stress test (6 × pub + 6 × sub, rapid create/destroy) with buffer-aware subscriptions
froze completely within ~40 s: the heartbeat thread stopped, and every thread blocked on
`futex_do_wait` in `/proc/PID/task/*/wchan`. The ABBA lock-order inversion between
`graph_mutex_` and `discovery_mutex_` (originally introduced before `e91c15a`) caused the
initial hang; the remaining hazard addressed here is the callback invocation under
`graph_mutex_`.

## Key Changes

- `update_topic_map_for_put()` returns `std::vector<EntityDiscoveryCallback>` instead of
  invoking callbacks internally; the same change propagates to `update_topic_maps_for_put()`
- `parse_put()` switches from `std::lock_guard` to `std::unique_lock` on `graph_mutex_`,
  calls `lock.unlock()` after all graph mutations are done, then iterates over the returned
  callbacks without holding any lock
- Header declarations updated to match new return types, with doc comments explaining the
  required call sequence

## Breaking Changes

None — the changed functions are `private` members of `GraphCache`.

## Relation to e91c15a

`e91c15a` fixed the ABBA by standardising lock order in `register_*_discovery_callback()`
to `graph_mutex_` → `discovery_mutex_`.  This PR is a complementary defensive measure:
even with consistent lock ordering, invoking external callbacks under an internal mutex
is an API contract violation that creates fragility when callbacks evolve.

---

## Did you use Generative AI?

Yes. Claude (claude-sonnet-4-6) via Claude Code was used to assist with creating an initial prototype version of the changes contained in this PR.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/graph_cache.cpp
+++ b/rmw_zenoh_cpp/src/detail/graph_cache.cpp
@@ -91,16 +91,18 @@ std::shared_ptr<GraphNode> GraphCache::make_graph_node(const Entity & entity) co
 }
 
 ///=============================================================================
-void GraphCache::update_topic_maps_for_put(
+std::vector<GraphCache::EntityDiscoveryCallback>
+GraphCache::update_topic_maps_for_put(
   GraphNodePtr graph_node,
   liveliness::ConstEntityPtr entity)
 {
   if (entity->type() == EntityType::Node) {
     // Nothing to update for a node entity.
-    return;
+    return {};
   }
 
   // First update the topic map within the node.
+  // These calls never fire discovery callbacks (report_events = false).
   if (entity->type() == EntityType::Publisher) {
     update_topic_map_for_put(graph_node->pubs_, entity);
   } else if (entity->type() == EntityType::Subscription) {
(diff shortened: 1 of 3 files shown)


YuanYuYuan:
Reproduction gist: https://gist.github.com/YuanYuYuan/5af4f1ca4183bf4ffa5c631f53b3f4b3

Contains `deadlock_repro_node.py` — a ROS 2 stress test (6 pub + 6 sub churn pairs at 2 ms
create/destroy rate) that reproduces the freeze.  Against a build from `de7cfab` with
`is_buffer_aware=true` patched, all 12 threads block on `futex_do_wait` within ~40 s and
the heartbeat file (written via `os.write()`, no GIL) also stops.

The `gist_repro.md` file documents the lock-order analysis, observed output, and relation
to `e91c15a`.

PR #937, opened 2026-03-31
Don't hold both locks when processing event data

haudren-woven:
## Description

Fixes #921. A deadlock can happen due to two code paths acquiring both locks in opposite order. The simplest fix is to take the locks in sequence, but not at the same time.

### Did you use Generative AI?

Github Copilot GPT-4.1 was used for some completions.

### Additional Information

I am not sure that this the best way to address the issue, but I'm hoping to get the conversation started.

PR #973, opened 2026-05-12
Avoid deadlocks when dispatching event callbacks

zhufangda-vita:
## Description

Avoid deadlocks caused by invoking event callbacks while holding internal mutexes.

This change moves callback execution out of `event_mutex_`, `events_mutex_`, and
`graph_mutex_` critical sections. Graph event callbacks are collected while the
graph is updated, enqueued in graph mutation order, and dispatched after internal
locks are released.

Key changes:
- Copy event callbacks and user data under `event_mutex_`, then invoke them after
  releasing the lock.
- Collect graph event callbacks under the existing lock order, enqueue them in a
  FIFO dispatch queue while `graph_mutex_` is still held, and drain the queue
  after releasing internal locks.
- Prevent recursive dispatch with a `dispatching_` flag.
- Catch and log exceptions from queued graph callbacks so one failing callback
  does not leave dispatch stuck or block later events.
- Use weak subscription data captures for graph event callbacks.

This follows the same general direction as #937 and #955: avoid holding internal
locks while invoking callbacks that may re-enter `rmw_zenoh`.

Fixes # (issue)

This should not change normal user-facing behavior. It is intended to prevent
deadlocks/hangs when graph or event callbacks are triggered concurrently.

Yes. ChatGPT/Codex/Claude/Copilot was used to help analyze the lock-order issue and draft parts
of the implementation.

### Additional Information

Validation performed:
- `git diff --check`
- `ament_uncrustify rmw_zenoh_cpp/src/detail/event.cpp rmw_zenoh_cpp/src/detail/graph_cache.cpp rmw_zenoh_cpp/src/detail/graph_cache.hpp rmw_zenoh_cpp/src/rmw_event.cpp`

A full `colcon build` could not be completed locally because the environment is
missing `cargo`; the build stops in `zenoh_cpp_vendor` before compiling
`rmw_zenoh_cpp`.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/event.cpp
+++ b/rmw_zenoh_cpp/src/detail/event.cpp
@@ -68,19 +68,31 @@ rmw_zenoh_event_status_t::rmw_zenoh_event_status_t()
 void DataCallbackManager::set_callback(
   const void * user_data, rmw_event_callback_t callback)
 {
-  std::lock_guard<std::mutex> lock_mutex(event_mutex_);
+  rmw_event_callback_t pending_callback = nullptr;
+  const void * pending_user_data = nullptr;
+  size_t pending_unread_count = 0;
 
-  if (callback) {
-    // Push events arrived before setting the the executor callback.
-    if (unread_count_) {
-      callback(user_data, unread_count_);
-      unread_count_ = 0;
+  {
+    std::lock_guard<std::mutex> lock_mutex(event_mutex_);
+
+    if (callback) {
+      // Push events arrived before setting the the executor callback.
+      if (unread_count_) {
+        pending_callback = callback;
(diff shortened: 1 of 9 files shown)


JEnoch:
Your PR is based on forked and modified versions of zenoh-c and zenoh-cpp:
- https://github.com/BOURBONCASK/zenoh-c.git
- https://github.com/BOURBONCASK/zenoh-cpp

If you think changes are required in Zenoh, please also submit a PR there and link it to this PR.
We'll need to review and merge the changes in Zenoh at first.

Yadunund:
Closing as the issue should be resolved by #992 and will be backported to `humble`. Please open a ticket if the issue persists.

PR #990, opened 2026-06-09
fix: keep pub_ valid for simple publishers (debug nightly crash)

### LINKED ISSUE #989: Consistent failures in rmw_zenoh with debug in nightlies
Looks to be a recent regression in rmw_zenoh across all nightly debugs.

Ref: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/3865/

Looks to be related to https://github.com/ros2/rmw_zenoh/pull/930

YuanYuYuan:
Closes #989.

## Summary

Fix a null advanced-publisher dereference that makes every `rmw_zenoh_cpp` pub/sub test fail in the `nightly_linux_debug` job (e.g. [nightly_linux_debug #3865](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/3865/) — 265 failures, all the same panic).

In debug builds every publish aborts with:

```
thread '<unnamed>' panicked at zenoh_c_vendor/src/advanced_publisher.rs:297:10:
unsafe precondition(s) violated: hint::unreachable_unchecked must never be reached
thread caused non-unwinding panic. aborting.
```

In release builds the same path is undefined behaviour (the check is compiled out), which is why only the debug nightly surfaces it.

## Root cause

The per-endpoint refactor (#930) added a base endpoint for simple (non-buffer-aware) publishers. The `PublisherData` constructor initialises `pub_` from its argument and then, for simple publishers, **moves `pub_` into `base_endpoint->pub`**:

```cpp
pub_(std::move(pub)),
...
base_endpoint->pub = std::optional<zenoh::ext::AdvancedPublisher>(std::move(pub_));
```

After the move, `pub_` is in a moved-from (null) state. But `publish()`, `publish_serialized_message()` and `shutdown()` all still operate on `pub_`, so the first `pub_.put()` calls `ze_advanced_publisher_loan_mut()` on a null owned publisher, hitting `unwrap_unchecked()` on `None`.

## Approach

Rather than just remove the stray move, this unifies how a publisher is represented. The deeper problem is that ownership lived in two divergent places — `PublisherData::pub_` and `endpoints_->pub` — with parallel creation and publish/shutdown paths that can drift out of sync (which is exactly how the bug arose). Patching the move fixes this instance but leaves the same trap for the next change. Collapsing to a single representation makes the inconsistent state unrepresentable, which is the better long-term-maintenance outcome (per review discussion).

## Key Changes

- Remove the `PublisherData::pub_` member. The standard/base publisher now lives in `base_endpoint_` (a `PublisherEndpoint`), like every other endpoint.
- `create_publisher_endpoint(full_key, bool buffer_aware)` is now the single place that declares an `AdvancedPublisher`: `buffer_aware=false` builds the base publisher (keeping the sample-miss-detection heartbeat), `buffer_aware=true` builds per-subscriber buffer-aware endpoints.
- `make()` creates the base publisher through `create_publisher_endpoint()` after construction instead of declaring it inline and passing it to the constructor.
- `publish()` / `publish_serialized_message()` publish through `base_endpoint_`; `shutdown()` undeclares `base_endpoint_` and every buffer-aware endpoint uniformly.

## Verification

Built the rolling stack in Debug (so zenoh-c carries `debug_assertions`) and ran the previously-failing tests against the patched RMW:

- Minimal repro (create one publisher and publish a message): no panic, clean exit.
- `test_communication`: 0 failures (17/17 message-type pub/sub, plus the launch-based requester/replier/action tests).
- `demo_nodes_cpp` / `logging_demo` / `image_tools` / `pendulum_control`: 16/16 pass, covering all the demo/tutorial/pendulum failures from #3865.

## Breaking Changes

None.

## Did you use Generative AI?

Yes. Claude (claude-opus-4-8) via Claude Code was used to assist with creating an initial prototype version of the changes contained in this PR.

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
@@ -153,53 +153,7 @@ std::shared_ptr<PublisherData> PublisherData::make(
     return nullptr;
   }
 
-  using AdvancedPublisherOptions = zenoh::ext::SessionExt::AdvancedPublisherOptions;
-  using SampleMissDetectionOptions = AdvancedPublisherOptions::SampleMissDetectionOptions;
-  auto adv_pub_opts = AdvancedPublisherOptions::create_default();
-
-  if (adapted_qos_profile.durability == RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL) {
-    // Allow this publisher to be detected through liveliness.
-    adv_pub_opts.publisher_detection = true;
-    adv_pub_opts.cache = AdvancedPublisherOptions::CacheOptions::create_default();
-    adv_pub_opts.cache->max_samples = adapted_qos_profile.depth;
-    if (adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
-      // If RELIABLE + TRANSIENT_LOCAL activate sample miss detection for subscriber
-      // to detect missed samples and retrieve those from the Publisher cache.
-      // HeartbeatSporadic is used to prevent excessive background traffic
-      adv_pub_opts.sample_miss_detection = SampleMissDetectionOptions{};
-      adv_pub_opts.sample_miss_detection->heartbeat =
-        SampleMissDetectionOptions::HeartbeatSporadic{
-        SAMPLE_MISS_DETECTION_HEARTBEAT_PERIOD};
-    }
-  }
-
-  zenoh::KeyExpr pub_ke(entity->topic_info()->topic_keyexpr_);
-  // Set congestion_control to BLOCK if appropriate.
-  auto pub_opts = zenoh::Session::PublisherOptions::create_default();
-  pub_opts.congestion_control = Z_CONGESTION_CONTROL_DROP;
-  if (adapted_qos_profile.reliability == RMW_QOS_POLICY_RELIABILITY_RELIABLE) {
-    pub_opts.reliability = Z_RELIABILITY_RELIABLE;
-    if (adapted_qos_profile.history == RMW_QOS_POLICY_HISTORY_KEEP_ALL) {
-      pub_opts.congestion_control = Z_CONGESTION_CONTROL_BLOCK;
-    }
-  } else {
-    pub_opts.reliability = Z_RELIABILITY_BEST_EFFORT;
-  }
-  adv_pub_opts.publisher_options = pub_opts;
-
   zenoh::ZResult result;
-  auto adv_pub = session->ext().declare_advanced_publisher(
-    pub_ke, std::move(adv_pub_opts), &result);
-  if (result != Z_OK) {
-    RMW_SET_ERROR_MSG("unable to create zenoh publisher cache");
-    return nullptr;
-  }
-
-  if (result != Z_OK) {
-    RMW_SET_ERROR_MSG("Unable to create Zenoh publisher.");
-    return nullptr;
-  }
-
   std::string liveliness_keyexpr = entity->liveliness_keyexpr();
   auto token = session->liveliness_declare_token(
     zenoh::KeyExpr(liveliness_keyexpr),
@@ -218,13 +172,23 @@ std::shared_ptr<PublisherData> PublisherData::make(
       node,
       std::move(entity),
       std::move(session),
-      std::move(adv_pub),
       std::move(token),
       type_support->data,
       std::move(message_type_support),
       has_buffer_fields,
       backend_metadata
     });
+
+  // Create the base publisher through the same path as every endpoint,
+  // stored in endpoints_ under the topic key expression.
+  const std::string base_key = pub_data->entity_->topic_info()->topic_keyexpr_;
+  auto base_endpoint = pub_data->create_publisher_endpoint(base_key, /*buffer_aware=*/false);
+  if (!base_endpoint) {
+    RMW_SET_ERROR_MSG("Unable to create Zenoh publisher.");
+    return nullptr;
+  }
+  pub_data->base_endpoint_ = base_endpoint.get();
+  pub_data->endpoints_[base_key] = std::move(base_endpoint);
 
   if (has_buffer_fields) {
     pub_data->local_endpoint_info_ =
@@ -282,7 +246,6 @@ PublisherData::PublisherData(
   const rmw_node_t * rmw_node,
   std::shared_ptr<liveliness::Entity> entity,
   std::shared_ptr<zenoh::Session> sess,
-  zenoh::ext::AdvancedPublisher pub,
   zenoh::LivelinessToken token,
   const void * type_support_impl,
   std::unique_ptr<MessageTypeSupport> type_support,
@@ -292,7 +255,6 @@ PublisherData::PublisherData(
   rmw_node_(rmw_node),
   entity_(std::move(entity)),
   sess_(std::move(sess)),
-  pub_(std::move(pub)),
   token_(std::move(token)),
   type_support_impl_(type_support_impl),
   type_support_(std::move(type_support)),
@@ -302,15 +264,6 @@ PublisherData::PublisherData(
   backend_metadata_(std::move(backend_metadata))
 {
   events_mgr_ = std::make_shared<EventsManager>();
-
-  // For simple publishers, create a single base endpoint
-  if (!is_buffer_aware_) {
-    auto base_endpoint = std::make_shared<PublisherEndpoint>();
-    base_endpoint->key = entity_->topic_info()->topic_keyexpr_;
-    base_endpoint->pub = std::optional<zenoh::ext::AdvancedPublisher>(std::move(pub_));
-    endpoints_[base_endpoint->key] = base_endpoint;
-  }
-  // For buffer-aware publishers, endpoints are created dynamically on subscriber discovery
 }
 
 ///=============================================================================
@@ -666,7 +619,7 @@ rmw_ret_t PublisherData::publish(
 
   TRACETOOLS_TRACEPOINT(
     rmw_publish, static_cast<const void *>(rmw_publisher_), ros_message, source_timestamp);
-  pub_.put(std::move(payload), std::move(opts), &result);
+  base_endpoint_->pub.value().put(std::move(payload), std::move(opts), &result);
   if (result != Z_OK) {
     if (result == Z_ESESSION_CLOSED) {
       RMW_ZENOH_LOG_WARN_NAMED(
@@ -740,7 +693,7 @@ rmw_ret_t PublisherData::publish_serialized_message(
       rmw_publish, static_cast<const void *>(rmw_publisher_), serialized_message,
       source_timestamp);
 
-    pub_.put(std::move(payload), std::move(opts), &result);
+    base_endpoint_->pub.value().put(std::move(payload), std::move(opts), &result);
   } else {
     std::vector<uint8_t> raw_image(
       serialized_message->buffer,
@@ -751,7 +704,7 @@ rmw_ret_t PublisherData::publish_serialized_message(
       rmw_publish, static_cast<const void *>(rmw_publisher_), serialized_message,
         source_timestamp);
 
-    pub_.put(std::move(payload), std::move(opts), &result);
+    base_endpoint_->pub.value().put(std::move(payload), std::move(opts), &result);
   }
 
   if (result != Z_OK) {
@@ -955,7 +908,7 @@ std::shared_ptr<PublisherData::PublisherEndpoint> PublisherData::get_or_create_e
 
 ///=============================================================================
 std::shared_ptr<PublisherData::PublisherEndpoint> PublisherData::create_publisher_endpoint(
-  const std::string & full_key)
+  const std::string & full_key, bool buffer_aware)
 {
   auto endpoint = std::make_shared<PublisherEndpoint>();
   endpoint->key = full_key;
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

ahcorde:
Pulls: ros2/rmw_zenoh#990
Gist: https://gist.githubusercontent.com/ahcorde/c0e8b53809a6362e1295c3dfe955b638/raw/dc3af366bd2984f1fbe71795cce23560164475a6/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19507
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29343/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22136/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9240/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28296/)

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -303,11 +303,16 @@ PublisherData::PublisherData(
...
+  // bookkeeping. The actual publisher stays owned by pub_, which is what
+  // publish()/publish_serialized_message()/shutdown() operate on. The base
+  // endpoint must NOT take ownership of pub_: moving it here would leave pub_ in
+  // a moved-from (null) state, and the next pub_.put() would dereference a null
+  // advanced publisher (debug builds abort in ze_advanced_publisher_loan_mut,
+  // release builds hit undefined behaviour).
   if (!is_buffer_aware_) {
     auto base_endpoint = std::make_shared<PublisherEndpoint>();
     base_endpoint->key = entity_->topic_info()->topic_keyexpr_;
-    base_endpoint->pub = std::optional<zenoh::ext::AdvancedPublisher>(std::move(pub_));
```
Yadunund, line 310: The fix is fine but leaving a suggestion to improve readability / internal APIs.

We have a [PublisherEndpoint::pub](https://github.com/ros2/rmw_zenoh/blob/e95c62df287143f78bdb41c452b6bf1e257b0c0d/rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp#L113) but it looks like we only want to assign it a value if the publisher is buffer-aware. 

We also have [zenoh::ext::AdvancedPublisher pub_](https://github.com/ros2/rmw_zenoh/blob/e95c62df287143f78bdb41c452b6bf1e257b0c0d/rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp#L158) in `PublisherData` and we have diverging codepaths to reference `pub_` or `endpoints_->pub` depending on whether we are buffer-aware. 

Here we directly instantiate `std::make_shared<PublisherEndpoint>()` and in other places we call [PublisherData::create_publisher_endpoint](https://github.com/ros2/rmw_zenoh/blob/e95c62df287143f78bdb41c452b6bf1e257b0c0d/rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp#L957).

To make things more consistent, how about
1. Moving `PublisherData::pub_` into `PublisherEndpoint` as well. We can either have an `std::variant` to get a an `AdvancedPublisher` that is simple or buffer-aware or just have two `std::optonal<AdvancedPublisher` objects for each type resp. 
2. Update `create_publisher_endpoint()` to create both types of endpoints by accepting a new argument, eg., `bool buffer_aware=false`.
3. Always rely on `create_publisher_endpoint()` throughout the codebase.

Yadunund (changes requested):
Lmk if the feedback makes sense! I'm still catching up on the buffer aware changes

nvcyc:
The fix direction is right. The base publisher should not be moved in the first place.
I'm suggesting a more complete removal here: https://github.com/ros2/rmw_zenoh/commit/0d653ae87e5e78124d9cfa31ed38abbdefef4edc

YuanYuYuan:
Thanks @Yadunund @nvcyc. I went with the full unification rather than just dropping the move.

The reason is long-term maintainability: the root issue isn't only the stray `std::move(pub_)` — it's that publisher ownership had two divergent representations (`PublisherData::pub_` vs `endpoints_->pub`) with parallel creation and publish/shutdown paths that could (and did) drift out of sync. Patching the move fixes this instance but leaves the same trap for the next change. Collapsing to a single representation makes the inconsistent state unrepresentable.

What changed:
- Removed `PublisherData::pub_`. The standard/base publisher now lives in `base_endpoint_` (a `PublisherEndpoint`), like every other endpoint.
- `create_publisher_endpoint(full_key, bool buffer_aware)` is now the single place that declares an `AdvancedPublisher` (`buffer_aware=false` builds the base publisher and keeps the sample-miss-detection heartbeat; `buffer_aware=true` builds per-subscriber buffer-aware endpoints).
- `make()` creates the base publisher through that path after construction instead of declaring it inline and passing it to the constructor.
- `publish()`/`publish_serialized_message()` publish through `base_endpoint_`; `shutdown()` undeclares `base_endpoint_` and every buffer-aware endpoint uniformly.

Verified on a Debug rolling build (zenoh-c with `debug_assertions`): the minimal publisher, `test_communication` (17/17, including the launch-based requester/replier/action tests), and `demo_nodes_cpp`/`logging_demo`/`image_tools`/`pendulum_control` (16/16) all pass — the `advanced_publisher.rs` panic is gone.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp
```diff
@@ -154,8 +155,10 @@ class PublisherData final
   std::shared_ptr<liveliness::Entity> entity_;
   // A shared session.
   std::shared_ptr<zenoh::Session> sess_;
-  // An owned AdvancedPublisher (for simple publishers, this is the only one).
-  zenoh::ext::AdvancedPublisher pub_;
+  // The standard/base publisher endpoint. Used by simple publishers and as the
+  // base/fallback channel for buffer-aware publishers. Created (like every
+  // endpoint) via create_publisher_endpoint().
```
Yadunund: The extra line in the comment seems very review specific. 

```suggestion
  // base/fallback channel for buffer-aware publishers.
```

rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp
```diff
@@ -169,8 +172,8 @@ class PublisherData final
   // Buffer-aware publisher fields
   bool is_buffer_aware_;
   std::unordered_map<std::string, std::string> backend_metadata_;
-  // For simple publishers: endpoints_ contains only base endpoint
-  // For buffer-aware: multiple endpoints based on discovered subscribers
+  // Additional buffer-aware publisher endpoints, keyed by full key and created
+  // on subscriber discovery. The standard publisher lives in base_endpoint_.
```
Yadunund: I'm still not a fan of having separate `base_endpoint_` and a map for `endpoints_`. Is there a way to combine them into a single map where the key for a simple publisher is static?

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -302,15 +264,9 @@ PublisherData::PublisherData(
...
-  if (!is_buffer_aware_) {
-    auto base_endpoint = std::make_shared<PublisherEndpoint>();
-    base_endpoint->key = entity_->topic_info()->topic_keyexpr_;
-    base_endpoint->pub = std::optional<zenoh::ext::AdvancedPublisher>(std::move(pub_));
-    endpoints_[base_endpoint->key] = base_endpoint;
-  }
-  // For buffer-aware publishers, endpoints are created dynamically on subscriber discovery
+  // base_endpoint_ (the standard publisher) is created by make() via
+  // create_publisher_endpoint(); buffer-aware endpoints_ are created on demand
+  // when subscribers are discovered.
```
Yadunund: Comments seem very review specific...Remove?

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
```diff
@@ -218,14 +172,24 @@ std::shared_ptr<PublisherData> PublisherData::make(
...
       std::move(token),
       type_support->data,
       std::move(message_type_support),
       has_buffer_fields,
       backend_metadata
     });
 
+  // Create the standard/base publisher through the same unified path used for
+  // every endpoint. The publisher is owned by base_endpoint_ (not a separate
+  // member), so there is a single representation of "this publisher's channels".
```
Yadunund: Extra lines seem more catered to addressing review feedback
```suggestion
  // every endpoint. 
```

Yadunund (changes requested):
Left more suggestions below. 

Wondering if we should merge @nvcyc suggestion above first to immediately unblock the nightly builds while we take our time with the refactor here.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp
YuanYuYuan: Done in c9efd9b. There is now a single `endpoints_` map: the base publisher lives in it under the static topic key expression, and buffer-aware endpoints under their full keys. `base_endpoint_` is no longer a separate owning member — it is a non-owning `PublisherEndpoint*` cached into the map so `publish()` reaches the base channel without a per-message map lookup (and is cleared in `shutdown()`). I also disambiguated the comments so "base publisher" always refers to the object and "standard" is used only for the CDR path.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.hpp
YuanYuYuan: Trimmed in c9efd9b — the comment now just describes `base_endpoint_` as a non-owning cache of the base publisher stored in `endpoints_`, without the review-specific line.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
YuanYuYuan: Removed in c9efd9b — the constructor no longer carries that comment.

rmw_zenoh_cpp/src/detail/rmw_publisher_data.cpp
YuanYuYuan: Trimmed in c9efd9b — reduced to a one-line description of what the statement does.

YuanYuYuan:
@Yadunund thanks for the review — all of it is addressed in c9efd9b (force-pushed; the per-comment threads have details):

- **Single representation**: removed the separate `pub_`/`base_endpoint_` ownership. There is now one `endpoints_` map — the base publisher lives under the static topic key expression, buffer-aware endpoints under their full keys. `base_endpoint_` is just a non-owning `PublisherEndpoint*` cached into that map so `publish()` avoids a per-message lookup (cleared in `shutdown()`).
- **Comment cleanup**: trimmed/removed the review-specific comments (ctor, make(), the `base_endpoint_`/`endpoints_` headers), and disambiguated terminology — "base publisher" now always means the object, "standard" only the CDR path.

Verified on a Debug rolling build (zenoh-c with debug_assertions, the original repro condition): the minimal publisher, `test_communication` (17/17, incl. the launch-based requester/replier/action tests), and `demo_nodes_cpp`/`logging_demo`/`image_tools`/`pendulum_control` (16/16) all pass — the `advanced_publisher.rs` panic is gone. I also exercised the buffer-aware path end-to-end (rosidl_buffer_backends_tutorials/robot_arm_demo, CPU + GPU subscribers) and confirmed the per-backend endpoints (base + cpu-group + accelerated) are created as expected.

Ready for another look when you have a moment.

Yadunund (approved):
I think there is room to make the API clearner without relying on a raw ptr but let's not hold this PR back any longer since nightly is broken. We can follow up in a subsequent PR.

Yadunund:
CI started with: https://gist.githubusercontent.com/Yadunund/c4074b12ea5bdedcddae7b60f60c34bc/raw/2ece5a6b002ddaee9f9a1f3000e72192b3f5fb20/ros2_ci.repos
```
  ros2/rmw_zenoh:
    type: git
    url: https://github.com/ZettaScaleLabs/rmw_zenoh.git
    version: fix/advanced-pub-debug-crash
  ```
  

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29463/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22250/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9351/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28412/)

ahcorde:
Pulls: ros2/rmw_zenoh#990
Gist: https://gist.githubusercontent.com/ahcorde/3a6f714e70cbc4b3298b2da76a7fe1f4/raw/dc3af366bd2984f1fbe71795cce23560164475a6/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19687
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29544/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22331/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9434/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28521/)

miguelgonrod:
Hi everyone, thanks for the work!! Is this ready to merge?

ahcorde:
@YuanYuYuan should we backport this to older distros?

YuanYuYuan:
https://github.com/Mergifyio backport lyrical

ahcorde:
https://github.com/Mergifyio backport lyrical

PR #994, opened 2026-06-12
Shut down entities on the application thread when deleting them

### LINKED ISSUE #993: Destroying a client while its reply callback is in flight deadlocks the callback thread (~ClientData runs the blocking z_undeclare_querier on its own callback thread)
## Summary

If `rmw_destroy_client` races with an in-flight service reply, the last `shared_ptr<ClientData>` reference can be dropped **inside the client's own reply closure**. `~ClientData()` then calls `ClientData::shutdown()`, which runs the blocking `z_undeclare_querier()` — and the undeclare waits for in-flight callbacks of that querier to finish, i.e. **for the very callback the thread is currently executing**. The thread deadlocks permanently.

The deadlock is not contained to the destroyed client: the stuck thread keeps whatever locks it was holding further down the stack. In our reproduction it holds `ServiceData::mutex_` (it was inside `ServiceData::send_response` → inline reply delivery), so even an unrelated thread calling `send_request` against that service blocks forever. On a real robot, the victim was the zenoh transport **rx task**, which killed all inbound traffic for the whole session — including zenoh's link-lease handling, so the session never noticed the link was dead and never reconnected. Only a process restart recovers.

## Affected versions

* `rolling` @ ec43c4c — reproduced deterministically with vanilla `ros-rolling-zenoh-cpp-vendor` (gtest + gdb stacks below).
* `humble` — same code in `NodeData::delete_client_data` / `~ClientData`; this is where we first hit it in production.

## Mechanism

1. The reply closure registered in `ClientData::send_request` captures a `weak_ptr<ClientData>` and locks it for the duration of the callback (`rmw_client_data.cpp`).
2. `rmw_destroy_client` → `NodeData::delete_client_data` only does `clients_.erase(client)` — it drops the owning reference without shutting the entity down first.
3. If a reply callback is executing at that moment, its transient `shared_ptr` (from `weak_ptr::lock()`) is now the **last** reference. When the closure finishes (or is dropped), `~ClientData()` runs **on the zenoh callback thread**.
4. `~ClientData()` → `shutdown()` → `std::move(querier_).undeclare()` blocks until all in-flight callbacks of the querier return — including the one this thread is executing. Self-deadlock.

The same erase-without-shutdown pattern exists in `delete_pub_data`, `delete_sub_data` and `delete_service_data`, which have the equivalent hazard with their own callbacks.

## Deterministic reproduction

`ClientData::add_new_reply` invokes the user's new-response callback (set via `rmw_client_set_on_new_response_callback`) synchronously while the reply closure still holds its strong reference. A gtest can therefore:

1. block inside the new-response callback,
2. destroy the client from the main thread while the callback is blocked,
3. release the callback, and
4. verify with a probe request that the callback thread survived.

On unpatched `rolling` the test deadlocks every time; gdb shows:

```
Thread 2 (executor thread, executing the service callback):
#0  syscall ()                                  ← parked forever
#4  z_undeclare_querier ()
#5  zenoh::Querier::undeclare(...) &&            querier.hxx:151
#6  rmw_zenoh_cpp::ClientData::shutdown          rmw_client_data.cpp:531
#7  rmw_zenoh_cpp::ClientData::~ClientData       rmw_client_data.cpp:468
#8  std::_Sp_counted_ptr<ClientData*>::_M_dispose      ← last reference dropped in the closure
#11-14 (libzenohc: inline reply delivery / closure drop)
#15 z_query_reply ()
#17 rmw_zenoh_cpp::ServiceData::send_response    rmw_service_data.cpp:463   ← still holds ServiceData state
#18 rcl_send_response ()

Thread 1 (main, collateral damage — blocked by the dead thread's ServiceData::mutex_):
#7  rmw_zenoh_cpp::ServiceData::add_new_query    rmw_service_data.cpp:240
#12 z_querier_get_with_parameters_substr ()
#14 rmw_zenoh_cpp::ClientData::send_request      rmw_client_data.cpp:408
```

## Field incident (humble)

A long-running process created a short-lived client every 16 minutes (`create_client` → call → scope exit). After ~7 h one destroy raced the in-flight reply: the zenoh transport rx thread deadlocked in `z_undeclare_querier` inside its own reply callback. From then on the session delivered nothing, the only WARNs were

```
zenoh::net::routing::dispatcher::queries: ... Didn't receive final reply for query ...: Timeout(600s)!
zenoh::api::session: Received ResponseFinal for unknown Request: N
```

every cycle, the kernel eventually dropped the TCP link to the router (lease handling was dead too, since it runs on the stalled rx path), and the session stayed a zombie — routing table still holding the dead transport face, no reconnect ever attempted.

We will open a PR with a fix and the deterministic regression test shortly.

BOURBONCASK:
## Description

Fixes #993.

`NodeData::delete_{pub,sub,service,client}_data` only erased the owning `shared_ptr` from the map, leaving `shutdown()` to run from whichever thread drops the last reference. If an entity is destroyed while one of its zenoh callbacks is in flight, the callback's transient `shared_ptr` (locked from the closure's `weak_ptr`) is the last reference, so the destructor runs the blocking undeclare on the zenoh callback thread itself — which waits for that same callback to finish. The thread deadlocks permanently and takes the entity mutex (and, on a deployed system, the whole session inbound path) with it. Full stacks and a field incident are in #993.

This PR:

* takes the entity out of the map under the lock, then shuts it down on the calling (application) thread outside the lock; the destructor's own `shutdown()` becomes a no-op via the existing `is_shutdown_` CAS regardless of which thread runs it (`rmw_node_data.cpp`);
* makes `ClientData::shutdown()` and `ServiceData::shutdown()` public, matching `PublisherData` / `SubscriptionData`;
* sets `initialized_ = true` at the end of `ServiceData::make()`. It was only ever initialized to `false`, so both undeclares in `ServiceData::shutdown()` were unreachable and the queryable was still undeclared by the member destructor — i.e. on the callback thread — even after the change above (found by @otamachan, see below);
* adds three deterministic regression tests in `test_rmw_zenoh_cpp`, one per entity kind that has a zenoh callback:
  * `test_client_destroy_during_reply`: blocks inside the new-response callback (invoked synchronously from `ClientData::add_new_reply` while the reply closure still holds its strong reference), destroys the client from the main thread, releases the callback, and verifies the callback thread can still serve a probe request;
  * `test_service_destroy_during_request`: same shape for a service, blocking inside the new-request callback invoked from `ServiceData::add_new_query` while the query closure holds its strong reference;
  * `test_subscription_destroy_during_sample`: parks one publishing thread inside the new-message callback (invoked from `add_new_message()` with `mutex_` held), starts the destroy so `shutdown()` queues on `mutex_`, then sends a second sample from another thread so a second callback is in flight when the undeclare runs. This guards the lock-release-before-undeclare ordering in `SubscriptionData::shutdown()` that #930 introduced on `rolling`; it is the AB-BA @otamachan hit on `jazzy`, where that ordering does not exist yet.

Rebased onto current `rolling`.

Verification (Docker `ros:rolling-ros-base`, Ubuntu 26.04 / x86_64, rolling binaries from `ros2-testing`, rebased onto `rolling` at 7fe3e65):

* `colcon test` for `rmw_zenoh_cpp` + `test_rmw_zenoh_cpp`: 236 tests, 0 failures, linters included. The three regression tests take ~1.1 s each and passed 3/3 reruns.
* Each test catches its bug when the corresponding fix is reverted:
  * `test_service_destroy_during_request` fails after its 10 s watchdog ("sender thread never returned … deadlocked") without the `initialized_ = true` line;
  * `test_subscription_destroy_during_sample` fails with "destroy returned: false, publisher B returned: false" when `SubscriptionData::shutdown()` is changed back to hold `mutex_` across the undeclare (the pre-#930 shape @otamachan hit on `jazzy`);
  * `test_client_destroy_during_reply` hangs with the original `delete_client_data` (caught by the test timeout), as in the previous revision of this PR.

No public API changes. It fixes a permanent deadlock/hang when a client, service or subscription is destroyed concurrently with an in-flight callback.

Yes. Claude (claude-fable-5 / claude-fable-5-1 and claude-opus-4-8) via Claude Code was used to assist with root cause analysis, reproducing the deadlock, and creating an initial prototype of the changes in this PR.

https://claude.ai/code/session_01317fxjjrM6WrQipxhN8qtm

### DIFF
--- a/rmw_zenoh_cpp/src/detail/rmw_service_data.cpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_service_data.cpp
@@ -189,6 +189,12 @@ std::shared_ptr<ServiceData> ServiceData::make(
     return nullptr;
   }
 
+  // Mark the entity as fully initialized so that shutdown() undeclares the
+  // queryable and the liveliness token itself, on the calling thread, instead
+  // of leaving that to the member destructors (which may run on a zenoh
+  // callback thread).
+  service_data->initialized_ = true;
+
   return service_data;
 }
 
--- a/test_rmw_zenoh_cpp/test/test_client_destroy_during_reply.cpp
+++ b/test_rmw_zenoh_cpp/test/test_client_destroy_during_reply.cpp
@@ -1,0 +1,157 @@
+// Copyright 2026 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <atomic>
+#include <chrono>
+#include <future>
+#include <memory>
+#include <thread>
+
+#include <rclcpp/rclcpp.hpp>
+#include <test_msgs/srv/basic_types.hpp>
+
+using namespace std::chrono_literals;
+
+namespace
+{
+template<typename Pred>
+bool poll_until(Pred && pred, std::chrono::seconds timeout)
+{
+  const auto deadline = std::chrono::steady_clock::now() + timeout;
+  while (std::chrono::steady_clock::now() < deadline) {
+    if (pred()) {
+      return true;
+    }
+    std::this_thread::sleep_for(10ms);
+  }
+  return pred();
+}
+}  // namespace
+
+class TestClientDestroyDuringReply : public ::testing::Test
+{
+public:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+  static void TearDownTestCase()
+  {
+    rclcpp::shutdown();
+  }
+};
+
+// Regression test: destroying a client while its zenoh reply callback is
+// still executing must not deadlock.
+//
+// The reply callback captures a weak_ptr to ClientData and locks it for the
+// duration of the callback. If rmw_destroy_client races with an in-flight
+// reply, the callback's transient shared_ptr can become the last strong
+// reference, so ~ClientData() — and with it the blocking querier undeclare,
+// which drains in-flight callbacks — runs on the very thread executing that
+// callback, waiting for itself forever.
+//
+// The test makes the race deterministic: ClientData::add_new_reply()
+// synchronously invokes the new-response callback while the reply closure
+// still holds its strong reference, so blocking inside that callback pins
+// the "reply in flight" state while the main thread destroys the client.
+TEST_F(TestClientDestroyDuringReply, DestroyWhileReplyCallbackInFlight)
+{
+  // The service lives on its own node, spun by a dedicated executor thread.
+  // The client node is intentionally never added to an executor so that
+  // client destruction order is fully controlled by this test.
+  auto service_node = std::make_shared<rclcpp::Node>("destroy_during_reply_service_node");
+  auto client_node = std::make_shared<rclcpp::Node>("destroy_during_reply_client_node");
+
+  auto service = service_node->create_service<test_msgs::srv::BasicTypes>(
+    "client_destroy_during_reply",
+    [](const std::shared_ptr<test_msgs::srv::BasicTypes::Request>,
+    std::shared_ptr<test_msgs::srv::BasicTypes::Response> response) {
+      response->bool_value = true;
+    });
+
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(service_node);
+  std::thread spinner([&executor]() {executor.spin();});
+
+  auto client = client_node->create_client<test_msgs::srv::BasicTypes>(
+    "client_destroy_during_reply");
+  ASSERT_TRUE(poll_until([&]() {return client->service_is_ready();}, 10s));
+
+  std::promise<void> reply_in_flight;
+  std::promise<void> release_reply;
+  std::shared_future<void> release_future = release_reply.get_future().share();
+  std::atomic<bool> first_reply{true};
+
+  // Fires inside ClientData::add_new_reply(), on the thread executing the
+  // zenoh reply closure, while that closure still holds a strong reference
+  // to ClientData.
+  client->set_on_new_response_callback(
+    [&reply_in_flight, release_future, &first_reply](size_t) {
+      if (first_reply.exchange(false)) {
+        reply_in_flight.set_value();
+        release_future.wait_for(10s);
+      }
+    });
+
+  client->async_send_request(std::make_shared<test_msgs::srv::BasicTypes::Request>());
+
+  ASSERT_EQ(reply_in_flight.get_future().wait_for(10s), std::future_status::ready)
+    << "service reply never reached the client";
+
+  // Keep the reply callback blocked while the client is destroyed, then let
+  // it finish: pre-fix, the callback thread then drops the last ClientData
+  // reference and self-deadlocks in the querier undeclare.
+  std::thread releaser([&release_reply]() {
+      std::this_thread::sleep_for(500ms);
+      release_reply.set_value();
+    });
+
+  client.reset();
+  releaser.join();
+
+  // Probe: the thread that executed the reply callback (the service node's
+  // spinner, since replies are delivered inline from rmw_send_response in a
+  // same-process setup) must still be alive and able to serve a new call.
+  auto probe = client_node->create_client<test_msgs::srv::BasicTypes>(
+    "client_destroy_during_reply");
+  ASSERT_TRUE(poll_until([&]() {return probe->service_is_ready();}, 10s));
+
+  std::promise<void> probe_replied;
+  std::atomic<bool> probe_first{true};
+  probe->set_on_new_response_callback(
+    [&probe_replied, &probe_first](size_t) {
+      if (probe_first.exchange(false)) {
+        probe_replied.set_value();
+      }
+    });
+  probe->async_send_request(std::make_shared<test_msgs::srv::BasicTypes::Request>());
+
+  const bool alive =
+    probe_replied.get_future().wait_for(10s) == std::future_status::ready;
+  if (!alive) {
+    // The executor thread is deadlocked; it can never be joined. Detach so
+    // the failure is reported instead of hanging in the thread destructor.
+    spinner.detach();
+    FAIL() << "no reply after destroying a client during its reply callback: "
+      "the zenoh callback thread is deadlocked (~ClientData ran a "
+      "blocking undeclare on its own callback thread)";
+  }
+
+  executor.cancel();
+  spinner.join();
+}
--- a/test_rmw_zenoh_cpp/test/test_service_destroy_during_request.cpp
+++ b/test_rmw_zenoh_cpp/test/test_service_destroy_during_request.cpp
@@ -1,0 +1,147 @@
+// Copyright 2026 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <atomic>
+#include <chrono>
+#include <future>
+#include <memory>
+#include <thread>
+
+#include <rclcpp/rclcpp.hpp>
+#include <test_msgs/srv/basic_types.hpp>
+
+using namespace std::chrono_literals;
+
+namespace
+{
+template<typename Pred>
+bool poll_until(Pred && pred, std::chrono::seconds timeout)
+{
+  const auto deadline = std::chrono::steady_clock::now() + timeout;
+  while (std::chrono::steady_clock::now() < deadline) {
+    if (pred()) {
+      return true;
+    }
+    std::this_thread::sleep_for(10ms);
+  }
+  return pred();
+}
+}  // namespace
+
+class TestServiceDestroyDuringRequest : public ::testing::Test
+{
+public:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+  static void TearDownTestCase()
+  {
+    rclcpp::shutdown();
+  }
+};
+
+// Regression test: destroying a service while its zenoh query callback is
+// still executing must not deadlock.
+//
+// The queryable callback captures a weak_ptr to ServiceData and locks it for
+// the duration of the callback. If rmw_destroy_service races with an in-flight
+// request, the callback's transient shared_ptr can become the last strong
+// reference, so ~ServiceData() — and with it the blocking queryable undeclare,
+// which drains in-flight callbacks — runs on the very thread executing that
+// callback, waiting for itself forever.
+//
+// rmw_destroy_service is expected to shut the entity down on the calling
+// thread instead. That only works if ServiceData::shutdown() actually
+// undeclares the queryable, which it skips unless the entity was marked
+// initialized.
+//
+// The test makes the race deterministic: ServiceData::add_new_query()
+// synchronously invokes the new-request callback while the query closure still
+// holds its strong reference, so blocking inside that callback pins the
+// "request in flight" state while the main thread destroys the service.
+TEST_F(TestServiceDestroyDuringRequest, DestroyWhileRequestCallbackInFlight)
+{
+  // Neither node is added to an executor: the request only has to reach the
+  // rmw layer, and service destruction order is fully controlled by this test.
+  auto service_node = std::make_shared<rclcpp::Node>("destroy_during_request_service_node");
+  auto client_node = std::make_shared<rclcpp::Node>("destroy_during_request_client_node");
+
+  auto service = service_node->create_service<test_msgs::srv::BasicTypes>(
+    "service_destroy_during_request",
+    [](const std::shared_ptr<test_msgs::srv::BasicTypes::Request>,
+    std::shared_ptr<test_msgs::srv::BasicTypes::Response> response) {
+      response->bool_value = true;
+    });
+
+  auto client = client_node->create_client<test_msgs::srv::BasicTypes>(
+    "service_destroy_during_request");
+  ASSERT_TRUE(poll_until([&]() {return client->service_is_ready();}, 10s));
+
+  std::promise<void> request_in_flight;
+  std::promise<void> release_request;
+  std::shared_future<void> release_future = release_request.get_future().share();
+  std::atomic<bool> first_request{true};
+
+  // Fires inside ServiceData::add_new_query(), on the thread executing the
+  // zenoh query closure, while that closure still holds a strong reference
+  // to ServiceData.
+  service->set_on_new_request_callback(
+    [&request_in_flight, release_future, &first_request](size_t) {
+      if (first_request.exchange(false)) {
+        request_in_flight.set_value();
+        release_future.wait_for(10s);
+      }
+    });
+
+  // In a same-process setup the query is delivered to the local queryable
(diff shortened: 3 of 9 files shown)


Yadunund:
@BOURBONCASK thanks for reporting the issue and opening a PR! Do you mind updating your PR description to follow the default template?

BOURBONCASK:
> @BOURBONCASK thanks for reporting the issue and opening a PR! Do you mind updating your PR description to follow the default template?

done — updated the description to follow the template. Thanks for taking a look!

otamachan:
We hit the same deadlock and tested this PR against a reproducer that destroys subscriptions and
services, not only clients. The client path is fixed; the subscription and service paths are not.
Applied the `rmw_zenoh_cpp` changes of this PR to the `jazzy` branch at 0.2.10.

Runs that deadlocked / runs:

| | unpatched | this PR | this PR + the two changes below |
|---|---|---|---|
| destroy subscriptions | 5/5 | 3/3 | 0/5 |
| destroy clients | 2/3 | 0/3 | 0/3 |
| destroy services | 5/5 | 3/3 | 0/5 |

Reproducer: https://gist.github.com/otamachan/515e3e699dbfe592207894789782eeb1

It runs in one process with `mode="peer"` and no router. A same-session publication is delivered
synchronously on the publishing thread (`Session::resolve_put::call_local`), so that thread takes
the role of the callback thread. Exit code 3 means the deadlock was hit.

### 1. Subscriptions: the self-deadlock becomes an AB-BA

`SubscriptionData::shutdown()` holds `mutex_` across the `undeclare`, and the sample callback takes
the same `mutex_` in `add_new_message()`. With this PR the application thread blocks in
`SyncGroup::wait` while holding `mutex_`, and the callback it waits for blocks on that `mutex_`.

```
Thread 13 (application thread):
#3  zenoh::api::cancellation::SyncGroup::wait
#4  ze_undeclare_advanced_subscriber
#6  rmw_zenoh_cpp::SubscriptionData::shutdown (this=0x5555556a73b0)   <- holds mutex_
#7  rmw_zenoh_cpp::NodeData::delete_sub_data                          rmw_node_data.cpp:284
#8  rmw_destroy_subscription

Thread 12 (in-flight sample callback):
#4  std::lock_guard<std::mutex>::lock_guard (__mutex=0x5555556a73c0)  <- SubscriptionData::mutex_
#5  rmw_zenoh_cpp::SubscriptionData::add_new_message (this=0x5555556a73b0)
#7  Closure<SubscriptionData::init()::<lambda(const zenoh::Sample&)>>::call
#9  zenoh_ext::advanced_subscriber::handle_sample
```

`shutdown()` needs to take `mutex_`, mark the shutdown, move `token_` / `sub_` / `sess_` into
locals, release the lock, and undeclare outside it.

### 2. Services: `ServiceData::shutdown()` never undeclares anything

`initialized_` is assigned only in the constructor's init list (`false`). `ServiceData::make()` does
not set it to `true`, and both undeclares in `ServiceData::shutdown()` are behind
`if (initialized_)`. The queryable and the liveliness token are released by the member destructors
instead, so calling `shutdown()` from `delete_service_data()` does not move the undeclare off the
callback thread.

```
#3  zenoh::api::cancellation::SyncGroup::wait
#4  z_undeclare_queryable
#6  zenoh::Owned<z_owned_queryable_t>::~Owned
#13 std::optional<zenoh::Queryable<void> >::~optional
#14 rmw_zenoh_cpp::ServiceData::~ServiceData                 rmw_service_data.cpp:474
#17 std::_Sp_counted_base<...>::_M_release_last_use_cold
#18 <F as zenoh::api::handlers::callback::CallbackImpl<T>>::call
#19 zenoh::api::session::Session::handle_query
#23 rmw_zenoh_cpp::ClientData::send_request                  rmw_client_data.cpp:400
```

`ServiceData::make()` should set `initialized_ = true` before returning, as
`SubscriptionData::init()` and `ClientData::make()` do.

Both changes are small:
https://github.com/otamachan/rmw_zenoh/commit/733f0e08d639e4fbf8c230026177394fe847443d

That branch is `jazzy` with this PR cherry-picked on top plus that commit, which is what the numbers
above were measured on. We can send it as a follow-up PR against `rolling`, or you can fold the
changes in here.

BOURBONCASK:
@otamachan thanks for the reproducer and the measurements. I was able to confirm both on my side:

* **Service**: folded into this PR. `ServiceData::make()` now sets `initialized_ = true`, so `shutdown()` really undeclares the queryable on the application thread.
* **Subscription**: on `rolling` this appears to be covered already by #930, which moved the undeclare in `SubscriptionData::shutdown()` outside `mutex_`. `jazzy` and `kilted` don't have that, and #930 itself doesn't look backportable (it depends on `rosidl_buffer_backend_registry`, which only exists on `rolling`). So for `jazzy` / `kilted` the `SubscriptionData::shutdown()` part of your commit would be needed as a separate PR — happy to defer to the maintainers on how to sequence that with the backport of this one.

I also added regression tests for the service and subscription paths next to the client one. The subscription test deadlocks against the pre-#930 `shutdown()`, so a backport of this PR without that change would fail loudly rather than regress silently.

Rebased onto current `rolling`; all three tests pass here and each fails when its fix is reverted.

otamachan:
Thanks for the regression tests and for tracking down #930.

I will open the `SubscriptionData::shutdown()` change as a separate PR against `jazzy` / `kilted`
after this one is backported, if it is still needed then.

We hit this deadlock in the field, so I hope this PR can be merged soon.

test_rmw_zenoh_cpp/test/test_client_destroy_during_reply.cpp
```diff
@@ -0,0 +1,157 @@
...
+  auto service = service_node->create_service<test_msgs::srv::BasicTypes>(
+    "client_destroy_during_reply",
+    [](const std::shared_ptr<test_msgs::srv::BasicTypes::Request>,
+    std::shared_ptr<test_msgs::srv::BasicTypes::Response> response) {
+      response->bool_value = true;
+    });
+
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(service_node);
+  std::thread spinner([&executor]() {executor.spin();});
```
YuanYuYuan, line 89: The failure path below detaches `spinner` when the probe never gets a reply. A detached thread that outlives this function must not hold a reference to a stack local -- `executor` here does, via `[&executor]`.

```suggestion
  // Heap-allocated and captured by value: the failure path below can
  // detach this thread, and a deadlocked spinner would then outlive
  // the test function while still using this executor.
  auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
  executor->add_node(service_node);
  std::thread spinner([executor]() {executor->spin();});
```

test_rmw_zenoh_cpp/test/test_client_destroy_during_reply.cpp
```diff
@@ -0,0 +1,157 @@
...
+  if (!alive) {
+    // The executor thread is deadlocked; it can never be joined. Detach so
+    // the failure is reported instead of hanging in the thread destructor.
+    spinner.detach();
+    FAIL() << "no reply after destroying a client during its reply callback: "
+      "the zenoh callback thread is deadlocked (~ClientData ran a "
+      "blocking undeclare on its own callback thread)";
+  }
+
+  executor.cancel();
```
YuanYuYuan, line 155: Follows from the `executor` change above.

```suggestion
  executor->cancel();
```

YuanYuYuan (commented):
Reviewed the fix and the three new regression tests. The core change (moving shutdown onto the calling thread, the is_shutdown_ CAS making the destructor's second call a no-op) is sound and the lock is released before the blocking undeclare.

Worth noting: this mirrors a defect one layer down, in zenoh core itself. [eclipse-zenoh/zenoh#2801](https://github.com/eclipse-zenoh/zenoh/pull/2801) fixes `SyncGroup::wait()` (called by `wait_callbacks()` at every zenoh-c teardown site). That function waits for a permit each running callback holds. It self-deadlocks when a callback drops the last reference to its own entity and that drop calls back into `wait_callbacks()` on the same stack. This PR fixes the same shape of hazard one layer up: a blocking undeclare that could run on its own callback thread. Both PRs are open. If zenoh core ships that fix, it closes part of this risk at the transport layer, but the rmw-layer fix here still matters for how this PR moves shutdown onto the calling thread and orders `NodeData`'s lock around it.

A few smaller findings, filed as inline suggestions where the diff allows it. Two findings (a leftover `initialized_` flag and an early-return in `shutdown()` that skips the second undeclare on error) touch code this PR doesn't otherwise change, so GitHub won't let me anchor a suggestion there -- left as one comment with a patch attached instead.

test_rmw_zenoh_cpp/test/test_service_destroy_during_request.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  // In a same-process setup the query is delivered to the local queryable
+  // synchronously from rmw_send_request, so the sender thread is the zenoh
+  // callback thread for this test.
+  std::promise<void> sender_done;
+  std::thread sender(
+    [&client, &sender_done]() {
+      client->async_send_request(std::make_shared<test_msgs::srv::BasicTypes::Request>());
+      sender_done.set_value();
+    });
```
YuanYuYuan, line 119: Same issue as `test_client_destroy_during_reply.cpp`: `sender` can be detached below, and `client`/`sender_done` are captured by reference to stack locals.

```suggestion
  // Heap-allocated and captured by value: the failure path below can
  // detach this thread, and a deadlocked sender would then outlive
  // the test function while still using them.
  auto sender_done = std::make_shared<std::promise<void>>();
  std::thread sender(
    [client, sender_done]() {
      client->async_send_request(std::make_shared<test_msgs::srv::BasicTypes::Request>());
      sender_done->set_value();
    });
```

test_rmw_zenoh_cpp/test/test_service_destroy_during_request.cpp
```diff
@@ -0,0 +1,147 @@
...
+      std::this_thread::sleep_for(500ms);
+      release_request.set_value();
+    });
+
+  service.reset();
+  releaser.join();
+
+  // The thread that executed the request callback must be able to return.
+  const bool alive =
+    sender_done.get_future().wait_for(10s) == std::future_status::ready;
```
YuanYuYuan, line 137:
```suggestion
    sender_done->get_future().wait_for(10s) == std::future_status::ready;
```

test_rmw_zenoh_cpp/test/test_subscription_destroy_during_sample.cpp
```diff
@@ -0,0 +1,166 @@
...
+      }
+    });
+
+  // Step 1: park thread A inside the callback.
+  std::promise<void> publisher_a_done;
+  std::thread publisher_a(
+    [&pub_a, &publisher_a_done]() {
+      pub_a->publish(test_msgs::msg::Empty());
+      publisher_a_done.set_value();
+    });
```
YuanYuYuan, line 121: `publisher_a`, `destroyer` and `publisher_b` below can all be detached on the failure path, and all three capture by reference to stack locals (`pub_a`/`pub_b`/`sub` and their `_done` promises). Splitting this across a few comments since the three threads are set up at different points in the function; this one covers `publisher_a`.

```suggestion
  // Heap-allocated and captured by value, same as their step-2/step-3
  // counterparts below: the failure path can detach these threads, and
  // a deadlocked thread would then outlive the test function while
  // still using them.
  auto publisher_a_done = std::make_shared<std::promise<void>>();
  std::thread publisher_a(
    [pub_a, publisher_a_done]() {
      pub_a->publish(test_msgs::msg::Empty());
      publisher_a_done->set_value();
    });
```

test_rmw_zenoh_cpp/test/test_subscription_destroy_during_sample.cpp
```diff
@@ -0,0 +1,166 @@
...
+    << "sample never reached the subscription";
+
+  // Step 2: destroy the subscription. shutdown() blocks on mutex_ behind A.
+  std::promise<void> destroy_done;
+  std::thread destroyer(
+    [&sub, &destroy_done]() {
+      sub.reset();
+      destroy_done.set_value();
+    });
+  std::this_thread::sleep_for(200ms);
```
YuanYuYuan, line 132: For `destroyer`: `sub` needs to stay the *sole* owner, so it has to be moved into the thread rather than copied -- a captured copy would leave the outer `sub` holding a reference, and `.reset()` inside the thread would then drop the refcount to one instead of zero, so the subscription would never actually be destroyed.

```suggestion
  // Moved into the thread, not copied: only the sole remaining owner
  // resetting to zero actually destroys the subscription.
  auto destroy_done = std::make_shared<std::promise<void>>();
  std::thread destroyer(
    [s = std::move(sub), destroy_done]() mutable {
      s.reset();
      destroy_done->set_value();
    });
  std::this_thread::sleep_for(200ms);
```

test_rmw_zenoh_cpp/test/test_subscription_destroy_during_sample.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  // Step 3: a second in-flight sample, blocked on mutex_ in add_new_message().
+  std::promise<void> publisher_b_done;
+  std::thread publisher_b(
+    [&pub_b, &publisher_b_done]() {
+      pub_b->publish(test_msgs::msg::Empty());
+      publisher_b_done.set_value();
+    });
+  std::this_thread::sleep_for(200ms);
```
YuanYuYuan, line 141: Same as `publisher_a` above, for `publisher_b`.

```suggestion
  auto publisher_b_done = std::make_shared<std::promise<void>>();
  std::thread publisher_b(
    [pub_b, publisher_b_done]() {
      pub_b->publish(test_msgs::msg::Empty());
      publisher_b_done->set_value();
    });
  std::this_thread::sleep_for(200ms);
```

test_rmw_zenoh_cpp/test/test_subscription_destroy_during_sample.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  // Step 4: let A go. The destroyer now undeclares while B is in flight.
+  release_sample.set_value();
+
+  const bool destroyed =
+    destroy_done.get_future().wait_for(10s) == std::future_status::ready;
+  const bool b_returned =
+    publisher_b_done.get_future().wait_for(10s) == std::future_status::ready;
+  const bool a_returned =
+    publisher_a_done.get_future().wait_for(10s) == std::future_status::ready;
```
YuanYuYuan, line 151:
```suggestion
  const bool destroyed =
    destroy_done->get_future().wait_for(10s) == std::future_status::ready;
  const bool b_returned =
    publisher_b_done->get_future().wait_for(10s) == std::future_status::ready;
  const bool a_returned =
    publisher_a_done->get_future().wait_for(10s) == std::future_status::ready;
```

rmw_zenoh_cpp/src/detail/rmw_service_data.cpp
```diff
@@ -189,6 +189,12 @@ std::shared_ptr<ServiceData> ServiceData::make(
     return nullptr;
   }
 
+  // Mark the entity as fully initialized so that shutdown() undeclares the
+  // queryable and the liveliness token itself, on the calling thread, instead
+  // of leaving that to the member destructors (which may run on a zenoh
+  // callback thread).
+  service_data->initialized_ = true;
```
YuanYuYuan, line 196: Three related findings that don't touch a line this PR changes, so I can't anchor them as suggestions here:

1. `initialized_` duplicates state `token_`/`qable_` (both `std::optional`) already carry, and is the flag whose stale default (`false`) is the bug this line fixes. Gating `shutdown()`'s undeclares on `token_.has_value()`/`qable_.has_value()` instead removes the flag entirely, and also closes a latent leak: if `make()` returns `nullptr` after `qable_` is set but before `token_` is, the local `shared_ptr`'s destructor runs `shutdown()` immediately, and with `initialized_` still `false` there the already-created queryable is never undeclared.
2. [`ServiceData::shutdown()`](https://github.com/ros2/rmw_zenoh/blob/536c1dd7/rmw_zenoh_cpp/src/detail/rmw_service_data.cpp#L523-L554) and [`ClientData::shutdown()`](https://github.com/ros2/rmw_zenoh/blob/536c1dd7/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp#L510-L540) both return as soon as the first undeclare fails, leaving the second resource (queryable/querier) to the destructor's own undeclare -- the same callback-thread hazard this PR removes for the first resource, left standing for the second whenever the first happens to fail.
3. **Correction to an earlier version of this comment**: I originally wrote that `SubscriptionData::shutdown()` "already attempts every undeclare and reports the first error." That's only true of its `endpoints_to_destroy` loop. Its `token_` undeclare and its `base_sub` undeclare each `return RMW_RET_ERROR` immediately on failure -- for `token_` that skips the loop and the `base_sub` undeclare entirely, and for `base_sub` it skips the trailing `sess_.reset()`. Same defect shape as 2, on the file I'd held up as the reference for it.

Patch for all three, applies cleanly on `536c1dd7`: https://gist.github.com/YuanYuYuan/15db5b3b12c9fca339c24ba133c7d938

PR #1033, opened 2026-08-20
Fix lost wakeup during wait set attachment

### LINKED ISSUE #1032: Lost wakeup: check_and_attach_condition() clears wait_set_data->triggered without holding condition_mutex (regression from #1005)
### Generated by Generative AI

Analysis assisted by Claude Fable 5. All source line references, the upstream history and the affected-branch list were verified by hand against the humble and rolling branches. The station symptoms and gdb captures are our own.

### Operating System:

Ubuntu 22.04 (jammy), x86_64, inside a Docker container. Kernel 7.0.0-28-generic.

### ROS version or commit hash:

Humble. ros-humble-rmw-zenoh-cpp 0.1.9-1jammy.20260723.022609 (binary from the ROS apt repo). Defect also present on jazzy, lyrical and rolling.

### RMW implementation (if applicable):

rmw_zenoh_cpp

### RMW Configuration (if applicable):

Control-path processes use client mode to a local router. Other processes on
the station use the default peer config, so the graph is mixed client/peer
bridged by the local router. The stall reproduces in both modes.

```json
{
  "mode": "client",
  "connect": { "endpoints": ["tcp/<router-host>:7447"] },
  "scouting": { "multicast": { "enabled": false, "interface": "lo" } }
}
```

**Shared memory is not enabled.** We use the Humble default
(`shared_memory.enabled: false` in `DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5`)
and set no `shared_memory` or `transport_optimization` keys anywhere. This is
therefore unrelated to the 0.1.9 SHM provider change described in #1030, no
traffic on the affected path uses SHM in either version.

### Client library (if applicable):

rclcpp — rclcpp::executors::MultiThreadedExecutor, via ros2_control controller_manager / ros2_control_node.

### 'ros2 doctor --report' output

None

### Steps to reproduce issue

 **This is a race, and we do not yet have a minimal reproducer.** The defect
 is however provable by inspection, see Actual behavior. We are filing on the
 code path plus symbolized evidence rather than waiting on a reproducer,
 because the window is workload-dependent and the analysis stands on its own.

 **What we observe (dual Robot Setup):**

 1. Two `ros2_control_node` processes, each running a `controller_manager`
   with a `MultiThreadedExecutor`.
 2. Each has one active controller with a single `sensor_msgs/JointState`
   subscription, **depth 1, default RELIABLE QoS**, plus one
    `TRANSIENT_LOCAL` `~/state` publisher. **No timers.**
 3. Targets are published to those subscriptions at ~30 Hz.
 4. The `controller_manager` update loop runs at 1 kHz with SCHED_FIFO
   priority 98, so the executor threads are heavily and irregularly
    preempted.
 5. Within tens of seconds to a few minutes, callback dispatch stops permanently in one process. The two processes fail independently
    (~16 s apart in our capture).
 6. **Running `ros2 topic echo` on any topic un-freezes the affected process
   for a while.**

 **Suggested minimal reproducer (designed from the analysis, NOT yet verified
 by us, offered as a starting point):**

 1. One node with ~50 subscriptions on distinct topics, to lengthen the
   `check_and_attach_condition()` walk.
 2. **No timers anywhere in the executor.** This is essential, see Additional
   information. A single timer masks the bug into millisecond-scale jitter.
 3. Spin it with a `MultiThreadedExecutor`.
 4. Publish to one of those topics at a few hundred Hz from another process.
 5. Add CPU contention (a few busy threads at high priority) to widen the
   window.
 6. Watch for callbacks stopping, using a plain `std::thread` watchdog, **not**
   a ROS timer.

### Expected behavior

A message delivered to a subscription that is attached to a wait set always
wakes `rmw_wait()`. Callback dispatch continues.

### Actual behavior

`rmw_wait()` sleeps indefinitely with `triggered == false` while messages sit
undispatched in `SubscriptionData::message_queue_`. The wakeup flag is erased by
a concurrent unsynchronized write.

**Root cause.** #1005 removed the `condition_mutex` that used to be held across
`check_and_attach_condition()`, but left this write at its tail
(`rmw_zenoh.cpp:2113-2116`, unchanged since #575, its comment is now false):

```cpp
// Note that wait_set_data->condition_mutex has been locked before calling
// check_and_attach_condition. So it's safe to modify the ... triggered flag.
wait_set_data->triggered = false;      // no lock is held here any more
```

**The race:**

1. `rmw_wait` clears `triggered` under the lock (2162-2166), then calls
   `check_and_attach_condition()` holding nothing.
2. Subscription `S` has an empty queue, so it is attached; the walk continues
   over the remaining entities.
3. An `rx` thread queues a message on `S`, takes `condition_mutex`, sets
   `triggered = true`, `notify_one()`. No waiter yet, so the flag is the only
   carrier of that wakeup.
4. The walk reaches line 2116 and sets `triggered = false` **unlocked**. The
   wakeup is destroyed.
5. `rmw_wait` takes the lock at 2182, sees `triggered == false`, and sleeps at
   2188 with `wait_timeout == NULL`. Nothing will wake it.

Steps 3 and 4 write the same plain `bool` with no common lock, so this is a data
race rather than merely an unlucky ordering. The same clobber can erase
`GuardCondition::trigger()` wakeups, so `rclcpp` cannot rescue the executor
either.

**Evidence.** Three detached, symbolized `gdb` samples per process one second
apart, same LWP at the same PC every time:

```text
rmw_wait()  at rmw_zenoh.cpp:2188        <- infinite wait
  wait_timeout = NULL   triggered = false   condition_mutex.__owner = 0
```

The Zenoh `rx-0`/`rx-1` threads were idle in `epoll_wait`, **not** in
`add_new_message`, `notify_event` or `check_and_attach_condition`. **This is not
the #921/#998 ABBA deadlock**; we looked for that cycle specifically and it is
absent.

Consistent with that, `ros2 topic echo` on any topic revives the frozen process:
the new node fires a graph guard condition, which sets `triggered = true` and
releases the sleeping wait. It re-freezes on a later iteration. No mutex cycle
could be broken by an unrelated external process. #1031 reports the same revival
via `ros2 topic hz`.

### Additional information

Delete `rmw_zenoh.cpp:2113-2116`.

```diff
-  // No conditions are available. Set the triggered flag of the wait_set to false.
-  // Note that wait_set_data->condition_mutex has been locked before calling
-  // check_and_attach_condition. So it's safe to modify the wait_set_data triggered flag.
-  wait_set_data->triggered = false;
-
   return false;
 }
```

This looks safe because #1005 already resets `triggered` at the top of
`rmw_wait()` under `condition_mutex` (lines 2162-2166) — which is exactly the
outcome #575 was after when it added this line. Post-#1005 the line is both
redundant and unsynchronized.

Note that wrapping the line in a lock guard would **not** fix it: that removes
the data race but keeps the clobber, so a legitimate concurrent trigger is
still erased and the wait still hangs.

### Why a timer normally hides this

`rcl_wait()` passes `timeout_argument = NULL` only when the wait set contains
no live, uncancelled timer (`rcl/wait.c`; `MultiThreadedExecutor` passes
`next_exec_timeout_`, default `-1`). Any timer produces a finite `wait_timeout`,
so a lost wakeup costs at most one timer period — a few ms of jitter nobody
ever reports.

Our stalling wait set had **no timer**, so there was no upper bound and the
lost wakeup became an unbounded hang. We suspect this is why the defect has
gone unnoticed since July, and it is why a reproducer must have **no timers**
in the executor.

### Why both known reports come from `ros2_control`

This looks structural rather than coincidental, and it follows from the
timer point above. In `ros2_control_node` the 1 kHz update loop is a
dedicated `SCHED_FIFO` thread, **not** a ROS timer, and controllers
typically add subscriptions rather than timers. So the executor's wait set
can legitimately contain **no timers at all** — which is exactly the
condition that turns a lost wakeup into an unbounded hang instead of a few
ms of jitter.

That would make `ros2_control` one of the few common workloads with no
safety net, which fits both this report and #1031 being `ros2_control`. It
is a hypothesis, but a cheap one to check: log `wait_timeout` on entry to
`rmw_wait` for a `ros2_control_node` and confirm it is `NULL`.

We do not think this is a `ros2_control` bug — we searched
`ros-controls/ros2_control` and found no matching report. It simply has the
entity shape that exposes the defect.

### Regression info

Introduced by #1005 (merged 16 Jul 2026), so present in every branch carrying
it: `rolling`, `lyrical` (#1010), `jazzy` (#1013/#1014), `humble` (#1015).
First affected Humble release is `rmw_zenoh_cpp` 0.1.9 (22 Jul 2026).

We verified line 2116 is still present on both `humble` and `rolling` as of
18 Aug 2026.

Downgrading is not a workaround, earlier versions carry the #921/#998
deadlock instead. There is currently no good version to sit on.

ktyang512:
## Description

Prevent `check_and_attach_condition()` from clearing
`wait_set_data->triggered` after its unlocked entity scan.

`rmw_wait()` already resets this flag under `condition_mutex` before the scan.
An entity can then set it while the remaining entities are being attached. The
second, unlocked reset erased that notification and could make a no-timer wait
sleep indefinitely.

This change makes attachment-scan completion read-only and adds a deterministic
regression test for the relevant ordering, without scheduler-dependent sleeps.

Fixes #1032

Yes. Executors without a timer no longer risk stalling indefinitely when an
entity becomes ready during wait-set attachment.

Yes. OpenAI Codex (GPT-5) assisted with repository analysis, the internal test
seam, regression test, implementation, validation, and PR text. The contributor
reviewed the rationale and selected the test design before submission.

### Additional Information

The regression failed before the fix because `triggered` became false. After
the fix, it passed 100 consecutive runs.

Local Rolling binary-path validation:

- built `zenoh_cpp_vendor`, `rmw_zenoh_cpp`, `zenoh_security_tools`, and
  `test_rmw_zenoh_cpp`;
- `colcon test-result`: 246 tests, 0 errors, 0 failures, 4 skipped;
- the skips are cppcheck records because the image's cppcheck 2.19 is
  intentionally not run by `ament_cppcheck` due to known performance issues;
- copyright, cpplint, lint_cmake, uncrustify, xmllint, the regression test, and
  isolated Zenoh session tests passed.

The Rolling-from-source / Ubuntu Resolute CI leg and release-branch backports
were not run locally.

### DIFF
--- a/rmw_zenoh_cpp/CMakeLists.txt
+++ b/rmw_zenoh_cpp/CMakeLists.txt
@@ -127,6 +127,7 @@ register_rmw_implementation(
 if(BUILD_TESTING)
   find_package(ament_cmake_copyright REQUIRED)
   find_package(ament_cmake_cpplint REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_cmake_lint_cmake REQUIRED)
   find_package(ament_cmake_uncrustify REQUIRED)
   find_package(ament_cmake_xmllint REQUIRED)
@@ -136,6 +137,13 @@ if(BUILD_TESTING)
   ament_lint_cmake()
   ament_uncrustify()
   ament_xmllint()
+
+  ament_add_gtest(test_rmw_wait test/test_rmw_wait.cpp)
+  target_include_directories(test_rmw_wait PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+  target_link_libraries(test_rmw_wait
+    ament_cmake_ros_core::ament_ros_defaults
+    rmw::rmw
+  )
 endif()
 
 install(
--- a/rmw_zenoh_cpp/package.xml
+++ b/rmw_zenoh_cpp/package.xml
@@ -31,6 +31,7 @@
   <depend>rmw_test_fixture</depend>
   <depend>tracetools</depend>
 
+  <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
 
--- a/rmw_zenoh_cpp/src/detail/rmw_wait.hpp
+++ b/rmw_zenoh_cpp/src/detail/rmw_wait.hpp
@@ -1,0 +1,34 @@
+// Copyright 2026 ktyang512
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DETAIL__RMW_WAIT_HPP_
+#define DETAIL__RMW_WAIT_HPP_
+
+#include "rmw/rmw.h"
+
+#include "rmw_wait_set_data.hpp"
+
+namespace rmw_zenoh_cpp::detail
+{
+
+inline bool condition_attachment_scan_complete(const rmw_wait_set_data_t *)
+{
+  // An entity may have set triggered while the attachment scan was in progress. Do not overwrite
+  // that notification here; rmw_wait() resets the flag under the mutex before starting the scan.
+  return false;
+}
+
+}  // namespace rmw_zenoh_cpp::detail
+
+#endif  // DETAIL__RMW_WAIT_HPP_
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -36,6 +36,7 @@
 #include "detail/message_type_support.hpp"
 #include "detail/qos.hpp"
 #include "detail/rmw_context_impl_s.hpp"
+#include "detail/rmw_wait.hpp"
 #include "detail/serialization_format.hpp"
 #include "detail/type_support_common.hpp"
 #include "detail/zenoh_utils.hpp"
@@ -2185,12 +2186,7 @@ check_and_attach_condition(
     }
   }
 
-  // No conditions are available. Set the triggered flag of the wait_set to false.
-  // Note that wait_set_data->condition_mutex has been locked before calling
-  // check_and_attach_condition. So it's safe to modify the wait_set_data triggered flag.
-  wait_set_data->triggered = false;
-
-  return false;
+  return rmw_zenoh_cpp::detail::condition_attachment_scan_complete(wait_set_data);
 }
 }  // namespace
 
--- a/rmw_zenoh_cpp/test/test_rmw_wait.cpp
+++ b/rmw_zenoh_cpp/test/test_rmw_wait.cpp
@@ -1,0 +1,31 @@
+// Copyright 2026 ktyang512
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "detail/rmw_wait.hpp"
+#include "detail/rmw_wait_set_data.hpp"
+
+TEST(TestRmwWait, preserves_trigger_set_during_condition_attachment)
+{
+  rmw_zenoh_cpp::rmw_wait_set_data_t wait_set_data{};
+
+  // Model the relevant interleaving without scheduler-dependent sleeps: an attached entity sets
+  // the flag while rmw_wait() is still scanning the remaining entities.
+  wait_set_data.triggered = true;
+
+  // Finishing the scan must not erase the notification before rmw_wait() evaluates its predicate.
+  EXPECT_FALSE(rmw_zenoh_cpp::detail::condition_attachment_scan_complete(&wait_set_data));
+  EXPECT_TRUE(wait_set_data.triggered);
+}


DoStraTech:
I believe I have a real-world Humble reproduction of this issue.

Environment:

* Ubuntu 22.04
* ROS 2 Humble
* `rmw_zenoh_cpp 0.1.9-1jammy.20260723.022609`
* MoveIt 2.5.9
* `rmw_zenohd` router

After running for a while, all services hosted by `move_group` stop responding, including:

* `/compute_ik`
* `/compute_fk`
* `/move_group/get_parameters`

An unrelated freshly started `AddTwoInts` service continues to work, so the Zenoh router and ROS service mechanism are not globally dead.

GDB of the stuck `move_group` shows the `MultiThreadedExecutor` thread blocked indefinitely in:

```text
rmw_wait()
rcl_wait()
rclcpp::Executor::wait_for_work()
rclcpp::Executor::get_next_executable()
rclcpp::executors::MultiThreadedExecutor::run()
```

The remaining `MultiThreadedExecutor` workers are blocked on the executor wait mutex.

As a first experiment, I rebuilt only `moveit_ros_move_group` 2.5.9 and changed:

```cpp
rclcpp::executors::MultiThreadedExecutor executor;
```

to use a 50 ms timeout:

```cpp
rclcpp::executors::MultiThreadedExecutor(
    rclcpp::ExecutorOptions(), 0, false,
    std::chrono::milliseconds(50));
```

This completely resolved the problem: `compute_fk`, `compute_ik`, and the parameter services responded normally. This was already consistent with a lost `rmw_wait()` wakeup, since the finite timeout forces the executor to periodically re-evaluate ready entities.

I then performed a more direct test of this PR.

I reverted MoveIt to the original executor with its default infinite wait and built `rmw_zenoh_cpp` 0.1.9 with a minimal Humble backport of the functional change from this PR: the `wait_set_data->triggered = false` reset at the end of `check_and_attach_condition()` was removed, while keeping the existing reset before the attachment scan.

With stock MoveIt behavior and only this `rmw_zenoh` change applied, the problem is also resolved.

So my test matrix is:

```text
stock MoveIt + stock rmw_zenoh 0.1.9
    -> services eventually freeze

MoveIt with 50 ms executor wait + stock rmw_zenoh 0.1.9
    -> works

stock MoveIt + rmw_zenoh 0.1.9 with this lost-wakeup fix
    -> works
```

This seems to be strong real-world confirmation that the lost wakeup fixed by this PR is the cause of the failure I was seeing on Humble.

The same application works with the other RMW implementations I tested; the freeze occurred specifically with `rmw_zenoh_cpp`.

It would be great to see this fix merged and, if possible, backported to Humble so it eventually reaches the vanilla apt-installed `rmw_zenoh_cpp`.

jmachowinski:
closing in favor for https://github.com/ros2/rmw_zenoh/pull/1036

PR #1005, opened 2026-07-02
Fix lock order inversion in rmw_wait

jmachowinski:
## Description

This is a second attempted to fix https://github.com/ros2/rmw_zenoh/issues/921

From my point of view the previous patches introduced in 
https://github.com/ros2/rmw_zenoh/pull/937 and #992 
introduced a use after free issue.

If the triggering thread is scheduled away after obtaining the wait_set_data_
pointer, another thread may delete the waitset directly afterwards.
Then the locally saved pointer points to invalid memory.

Fixes #921

### Is this user-facing behavior change?

No deadlocks any more ?

### Did you use Generative AI?

I asked AI up front and it came up with the broken patch from #937
This one was written by me...

### Additional Information

We hit this bug on shutdown of ros2 bag record.

PR #1036, opened 2026-08-26
Do not update wait_set_data triggered flag outside of mutex lock

Yadunund:
## Description

https://github.com/ros2/rmw_zenoh/pull/1005 did the right thing by resetting the `wait_set_data->triggered` flag to `false` under global lock as [seen here](https://github.com/ros2/rmw_zenoh/blob/e3159856e3816e726f2c5f1f9b4858ad6b7ee63e/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2237-L2241) but it kept the reset in [check_and_attach_condition()](https://github.com/ros2/rmw_zenoh/blob/e3159856e3816e726f2c5f1f9b4858ad6b7ee63e/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2188-L2191) that is no longer called under the same lock. This leads to the executor waiting indefinitely if one of the entities received a message after the initial check in `check_and_attach_condition()` given the reset happens outside of a lock.

Fixes # (issue)

https://github.com/ros2/rmw_zenoh/issues/1032

Yes. Fixes indefinite executor wait.

No

### DIFF
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -2185,11 +2185,6 @@ check_and_attach_condition(
     }
   }
 
-  // No conditions are available. Set the triggered flag of the wait_set to false.
-  // Note that wait_set_data->condition_mutex has been locked before calling
-  // check_and_attach_condition. So it's safe to modify the wait_set_data triggered flag.
-  wait_set_data->triggered = false;
-
   return false;
 }
 }  // namespace
@@ -2242,7 +2237,7 @@ rmw_wait(
 
   {
     // We explicitly do not lock the condition_mutex here
-    // This is fine, as the attachment returns atomically is a signal was ready
+    // This is fine, as the attachment returns atomically if a signal was ready
     // If anything triggers after that point, wait_set_data->triggered will be set
     // to true under mutex.
     // Note taking the mutex here leads to a deadlock.
--- a/rmw_zenoh_cpp/src/rmw_zenoh.cpp
+++ b/rmw_zenoh_cpp/src/rmw_zenoh.cpp
@@ -2111,6 +2111,9 @@ check_and_attach_condition(
   const rmw_events_t * const events,
   rmw_zenoh_cpp::rmw_wait_set_data_t * wait_set_data)
 {
+  // This function runs without wait_set_data->condition_mutex. Taking it here
+  // deadlocks against the producers, so this function must never write
+  // wait_set_data->triggered.
   if (guard_conditions) {
     for (size_t i = 0; i < guard_conditions->guard_condition_count; ++i) {
       rmw_zenoh_cpp::GuardCondition * gc =


jmachowinski (approved):

ahcorde (approved):

YuanYuYuan (commented):
@Yadunund Thank you for the fix. It looks good to me. Just a share of my understanding of this issue and a nit suggestion. 

P.S. This fix should also go through the other distros due to the previous backports.

## The fix closes the data race

I enumerated every access to `rmw_wait_set_data_t::triggered`. There are seven. After this patch, every one holds `condition_mutex`.

| site | access | locks held |
|---|---|---|
| [`rmw_zenoh.cpp#L2235`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2233-L2236) | write `false` | `condition_mutex` |
| [`rmw_zenoh.cpp#L2260`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2258-L2268) | read — condvar predicate | `condition_mutex` |
| [`rmw_zenoh.cpp#L2267`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2258-L2268) | read — condvar predicate | `condition_mutex` |
| [`guard_condition.cpp#L42`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/detail/guard_condition.cpp#L36-L45) | write `true` | `internal_mutex_` → `condition_mutex` |
| [`event.cpp#L259`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/detail/event.cpp#L255-L262) | write `true` | `event_condition_mutex_` → `condition_mutex` |
| [`rmw_subscription_data.cpp#L1176`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp#L1172-L1178) | write `true` | `mutex_` → `condition_mutex` |
| [`rmw_service_data.cpp#L268`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/detail/rmw_service_data.cpp#L264-L270) | write `true` | `mutex_` → `condition_mutex` |
| [`rmw_client_data.cpp#L255`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/detail/rmw_client_data.cpp#L251-L257) | write `true` | `mutex_` → `condition_mutex` |

The deleted line was the only unsynchronized access. All five writers take the locks in one order. The patch adds no new cycle.

## The fix also closes the lost wakeup

That is a second claim, and it needs a separate argument. #1032 states why: a lock guard around the deleted line removes the data race but keeps the clobber. The executor still hangs.

Attachment is what orders the two threads. The emptiness check and the attachment share one critical section of the entity lock — see [`SubscriptionData::queue_has_data_and_attach_condition_if_not`](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp#L916-L927). So exactly one of two orders holds for each entity.

```mermaid
sequenceDiagram
    autonumber
    participant E as Executor · rmw_wait
    participant S as SubscriptionData
    participant P as Zenoh rx thread
    E->>E: lock condition_mutex, triggered = false, unlock
    E->>S: queue_has_data_and_attach_condition_if_not()
    activate S
    Note over S: holds mutex_ across check and attach
    alt the queue already holds data
        S-->>E: true
        deactivate S
        Note over E: skip_wait — rmw_wait never sleeps
    else the queue is empty
        S->>S: wait_set_data_ = wait_set
        S-->>E: false
        P->>S: add_new_message()
        activate S
        Note over S: holds mutex_, then condition_mutex
        S->>E: triggered = true, notify_one()
        deactivate S
        E->>E: lock condition_mutex, predicate is true, return
    end
```

There is no third case. The deleted line was the only way to fall out of both. That is what makes this a complete fix, and not a narrowed window.

## Suggestion 

The deleted comment claimed a lock that #1005 had already removed. The comment read as authoritative and was false. Nothing reported that.

The rule this patch restores is negative: **`check_and_attach_condition()` runs unlocked, so it must never write `triggered`.** One line at [the head of that function](https://github.com/ros2/rmw_zenoh/blob/b58725f/rmw_zenoh_cpp/src/rmw_zenoh.cpp#L2104-L2112) states it where a future reader will look.

```cpp
// This function runs WITHOUT wait_set_data->condition_mutex. Taking it here
// deadlocks against the producers, so this function must never write
// wait_set_data->triggered. See #1032.
```

YuanYuYuan (approved):

PR #1043, opened 2026-09-02
Granular rclcpp/rclcpp.hpp and include what you use

ahcorde:
## Description

Granular rclcpp/rclcpp.hpp and include what you use

Claude Opus 4.8


wjwwood (approved):
tentative lgtm, but two things stand out:

- there are lots of not-rclcpp-related changes, which may need to be done, but are a bit off-topic
- why is rmw_zenoh depending on rclcpp in the first place? seems like it introduces a release dependency loop for this repository

PR #1053, opened 2026-09-07
Return topic info by const reference

mini-1235:
## Description

This PR return immutable topic metadata by const reference instead of copying it on each access. This updates Entity and endpoint topic_info() accessors, converts applicable local copies to borrowed references. TopicInfo is initialized with its owning entity and remains unchanged for the entity’s lifetime, so existing callers can safely inspect it without constructing and destroying an independent copy.

and

Also removes 2x unused variable in graph_cache.cpp

This came up while I was doing some profiling. Before this PR, the flamegraph shows the cost of copying `TopicInfo`:
Screenshot 2026-09-08 at 00-56-02

However, TopicInfo is not cheap to copy. It contains data such as:
- Topic/service name
- Type name and type hash
- Zenoh key-expression strings
- QoS data
- Optional backend metadata and containers

This copy was also happening in `take_one_message()`.

After this PR:
Screenshot 2026-09-08 at 00-57-24

The `TopicInfo` copy no longer appears in the flamegraph.

No I don't think so

Yes, Codex 5.6 Sol for reviewing and generating the PR description


ahcorde (approved):

Yadunund:
@mini-1235 glad to hear you were mindful of these effects during your development. I agree that relying on `shared_ptrs` would be safer and prevent any future regressions. Do you have the bandwidth to prototype those changes? Would be happy to review it and changes to optimize `hash_gid`.

mini-1235:
> @mini-1235 glad to hear you were mindful of these effects during your development. I agree that relying on `shared_ptrs` would be safer and prevent any future regressions. Do you have the bandwidth to prototype those changes? Would be happy to review it and changes to optimize `hash_gid`.

@Yadunund Yes, I have a few other things to prioritize at the moment. I am planning to continue the work in early to mid-October, and I will ping you once I have something ready for review!

Yadunund:
Sounds good, looking forward to it!

PR #1059, opened 2026-09-11
Remove unnecessary immutable getter locks

mini-1235:
## Description

Follow-up to https://github.com/ros2/rmw_zenoh/pull/1048: I took a closer look at the publisher, subscription, client, and service implementations today and found a few more unnecessary mutex locks.

This removes six unnecessary mutex acquisitions from the publisher/subscription GID, event-manager, and graph-cache getters. These getters either read immutable entity data or copy shared pointers that are never reassigned after initialization.

No

Yes, Codex 5.6 Sol for reviewing and helping with the PR description


ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport lyrical kilted jazzy humble

